PR #3, opened 2017-09-22
Add POSTagger benchmark

apaszke:


benchmarks/__init__.py
```diff
@@ -31,6 +31,7 @@ def make_params(**kwargs):
 class Benchmark(object):
     timer = time.time if PY2 else time.perf_counter
     default_params = []
+    params = make_params()
```
ezyang: What's this

apaszke: It just sets the default

PR #6, opened 2017-09-22
bnlstm

ezyang:
Signed-off-by: Edward Z. Yang <ezyang@fb.com>


ezyang:
Uh, not obviously? The batching and the input feeding seems to be different...

PR #8, opened 2017-10-05
add memnn benchmark

alexholdenmiller:
copied current memnn model from ParlAI, made benchmark from it


ezyang:
Great, thank you!

PR #19, opened 2018-05-24
[wip] Update rnn benchmarks

zou3519:
RNN benchmarks can be run with python rnns/runner.py which runs
everything and gives a nice summary view.
They are written in the style of the benchmarks in the scripts/ folder
(those are used by the CI).

Run individual benchmarks with (cd into ./rnns):

    python -m benchmarks.sru

Individuals scripts have their own command line options to test
different things (--jit, --backward, etc).

[Sample output](https://gist.github.com/zou3519/10f9b201b2a44e65acf93b73ca7facef)

WIP because the following things might be good to do:
- [ ] Cleanup code and bring (at least the test scripts) to Pytorch 0.4 style.
- [x] Try to apply jit tracing to some of the models (bnlstm, lstm variants). 

cc @zdevito


rnns/wsj.pkl
```diff
@@ -0,0 +1,2758 @@
+(lp0
```
apaszke: Is this checked in on purpose?

zou3519: Yes, but I can just read the file from `scripts/mini_sequence_labeler/wsj.pkl` instead

apaszke: No, that's fine. Just making sure it's not an accidental check in.

PR #20, opened 2018-05-24
Initial commit of cpu performance scripts

cpuhrsch:
This splits the repository into two categories and starts an effort to significantly extend the benchmark suite.


apaszke:
Well the model parts aren't really for checking accuracy. We're using them for benchmarking as well (mostly the JIT).

cpuhrsch:
I noticed a clear split however. We can still create a top-level folder to share code. But I think it's good to have this clear separation.

PR #24, opened 2018-05-31
Deduplicate more python scripts based on their intention

cpuhrsch:


zou3519:
WAIT

python/ubenchmarks/nn_comp.py
```diff
@@ -32,7 +32,7 @@
         # numbers = numbers.transpose(0, 1)
         # numbers = numbers.sigmoid()
         # numbers = numbers.transpose(0, 1)
-        # numbers = numbers.tanh()
+        numbers = numbers.tanh()
```
zou3519: Did you mean to commit this?

cpuhrsch: Yes

cpuhrsch:
@zou3519 - yes, I had briefly added this as a submodule to look at whether it's possible to pull in foreign benchmarks and interface them. it doesn't seem worth the trouble.

apaszke:
Do those files exist somewhere else as well?

cpuhrsch:
@apaszke - Yes, they're part of the toplevel 'rnn' folder whose files be integrated into our benchmark suite once it's setup.

PR #25, opened 2018-06-05
Initial python framework setup

cpuhrsch:
Basic framework setup. See examples under timing/python/benchmarks


timing/python/framework/benchmark.py
```diff
@@ -0,0 +1,290 @@
...
+# benchmark [--benchmark_list_tests={true|false}]
+#           [--benchmark_filter=<regex>]
+#           [--benchmark_format=<console|json|csv>]
+#           [--benchmark_out=<filename>]
+#           [--benchmark_out_format=<json|console|csv>]
+#           [--benchmark_color={auto|true|false}]
+#           [--benchmark_counters_tabular={true|false}]
+#           [--v=<verbosity>]
+# DONE
+#           [--benchmark_min_time=<min_time>]
```
zou3519: Not sure where to leave this comment, but what units are benchmark_min_time and benchmark_max_time in?

cpuhrsch: seconds. we still need user counters etc. I'm only leaving it for myself. I guess I can remove it. I expect another PR tonight that does 50% of this.

timing/python/framework/benchmark.py
```diff
@@ -0,0 +1,290 @@
...
+        self.results.append(result)
+
+
+class Benchmark(object):
+    def __init__(self):
+        if "args" not in dir(self):
+            logging.warn(
+                "Benchmark " + str(self.__class__) + " has no args set"
+            )
+            return
```
zou3519: Maybe you shouldn't return here?

PR #27, opened 2018-06-12
LSTM benchmark for the new python timing framework

zou3519:
This is (one of the) lstm benchmarks that is running on the CI. I've
ported it to the new framework.

cc @cpuhrsch

### DIFF
--- a/timing/python/benchmarks/__init__.py
+++ b/timing/python/benchmarks/__init__.py
@@ -1,3 +1,4 @@
 from benchmarks.compare_numpy import NumpyComparison
 from benchmarks.convnet_benchmark import Convnets
+from benchmarks.lstm_benchmark import LSTMBench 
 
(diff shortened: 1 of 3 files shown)


zou3519:
@cpuhrsch 
`python run.py --include LSTMBench`
I don't have a screen big enough to show this right now, but not all of the argument lists in the ListBenchmark have the same arguments (because they're being passed into a function with default args). This results in some weird printing, idk how much this matters though because I imagine when we're using this we'd want to specify all of the arguments so it is clear what is being benchmarked.

cpuhrsch:
Looks good to me. I'll merge and then debug if that's fine? This repo is still flagged as "in transit". So, no user can expect this code to work just yet.

zou3519:
Yes go ahead!

PR #40, opened 2018-11-28
Use Abs when computing tensor differences.

kit1980:


kit1980:
@soumith `abs(tensor - expected).max()` should work just as fine, no? https://pytorch.org/docs/stable/torch.html#torch.abs

I'll change if you think it's better style.

soumith:
`abs(x)` where `x` is a torch.Tensor treats the Tensor as an iterable, and calls the builtin `abs` function on it. This will be REALLY slow, compared to calling `torch.abs(x)` or `x.abs()`

PR #49, opened 2019-02-05
Add dropout lstm

wanchaol:
Add the lstm with dropouts to custom_lstm

### DIFF
--- a/rnns/fastrnns/custom_lstms.py
+++ b/rnns/fastrnns/custom_lstms.py
@@ -41,22 +41,19 @@ def script_lstm(input_size, hidden_size, num_layers, bias=True,
         stack_type = StackedLSTM2
         layer_type = BidirLSTMLayer
         dirs = 2
+    elif dropout:
+        stack_type = StackedLSTMWithDropout
+        layer_type = LSTMLayer
+        dirs = 1
     else:
         stack_type = StackedLSTM
         layer_type = LSTMLayer
         dirs = 1
 
-    if dropout:
-        return stack_type(num_layers, layer_type,
-                          first_layer_args=[LSTMCell, input_size, hidden_size],
-                          other_layer_args=[LSTMCell, hidden_size * dirs,
-                                            hidden_size],
-                          dropout=0.4)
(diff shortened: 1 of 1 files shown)


rnns/fastrnns/custom_lstms.py
```diff
@@ -246,12 +252,20 @@ def init_stacked_lstm(num_layers, layer, first_layer_args, other_layer_args):
 
 
 class StackedLSTM(jit.ScriptModule):
-    __constants__ = ['layers']  # Necessary for iterating through self.layers
+    # Necessary for iterating through self.layers and dropout support
+    __constants__ = ['layers', 'num_layers', 'dropout']
```
zou3519: I know this is bad from a software engineering perspective, but could you create a separate `StackedLSTMWithDropout?` I'd like to keep a vanilla "StackedLSTM" around (so that it does not get too complicated)

rnns/fastrnns/custom_lstms.py
```diff
@@ -246,12 +252,20 @@ def init_stacked_lstm(num_layers, layer, first_layer_args, other_layer_args):
...
+    def __init__(self, num_layers, layer, first_layer_args, other_layer_args,
+                 dropout=0.0):
         super(StackedLSTM, self).__init__()
         self.layers = init_stacked_lstm(num_layers, layer, first_layer_args,
                                         other_layer_args)
+        # If non-zero, introduces a Dropout layer on the outputs of each LSTM
+        # layer except the last layer, with dropout probability = dropout.
+        self.dropout = dropout
+        self.num_layers = num_layers
+        self.dropout_layer = jit.trace(nn.Dropout(dropout), torch.randn(2, 2),
```
zou3519: Is the tracing necessary -- can't we just do `self.dropout_layer = nn.Dropout(dropout)`?

wanchaol: Ohh you are right, nn.Dropout is a weak script module so it should be automatically converted to a ScriptModule here, I will update it.

PR #50, opened 2019-02-06
Add flake8 config from pytorch and format the code

wanchaol:
port .flake8 config from pytorch to the benchmarks repo and format the code, to make warnings disappear in the linter tool, only format changes so should not affect any existing code.


apaszke:
BTW If we want to commit to a style guide we should get a CI job running. Otherwise it's meaningless and will get out of sync quickly

PR #53, opened 2019-02-22
Add cudnn layernorm lowerbound benchmark

wanchaol:
Benchmark results:

```
Benchmarking LSTMs...
            name        avg_fwd          std_fwd          avg_bwd          std_bwd
cudnn_layernorm          32.71           0.7494            10.43          0.08965
jit_layernorm            41.25           0.7082            98.66             2.56
jit_layernorm_de         34.41           0.7501            113.3            1.037
```
only forward pass is valid

### DIFF
--- a/rnns/fastrnns/factory.py
+++ b/rnns/fastrnns/factory.py
@@ -267,6 +267,43 @@ def varlen_lstm_creator(script=False, **kwargs):
         backward=simple_backward)
 
 
+# cudnn_layernorm_lstm: since cudnn does not have Layernorm LSTM, we cannot benchmark
+# the lowerbound directly. Instead, we only benchmark the foward pass by mimicing the
+# computation of a cudnn lstm + seq_len * 3 layernorm computation. This should serve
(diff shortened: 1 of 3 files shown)


rnns/fastrnns/factory.py
```diff
@@ -267,6 +267,43 @@ def varlen_lstm_creator(script=False, **kwargs):
...
+            cy = ln_c(cy)
+
+        return out, (hy, cy)
+
+    return ModelDef(
+        inputs=[input, hidden],
+        params=flatten_list(module.all_weights),
+        forward=forward,
+        backward_setup=lstm_backward_setup,
+        backward=simple_backward)
```
zou3519: (just a note, no action items) The backward timing is a lower bound and probably not a model of what it should look like. Because of that, people looking at the benchmark should ignore the number. One way to do that is to set backward=None, this will make it so that the backward reports a 0.01 time (the launch overhead time) so it is obvious that the backward time is not valid

PR #54, opened 2019-03-07
separate bench groups to avoid unnecessary runs

wanchaol:
This PR separate the rnn/cnns benchmarks when specifying using --cnns or --rnns together with --group. If we don't specify the flags, run both groups. Mainly refactor this to avoid unnecessary resnet runs if I want to run rnn only

### DIFF
--- a/rnns/fastrnns/bench.py
+++ b/rnns/fastrnns/bench.py
@@ -137,6 +137,8 @@ def bench_group(model_list, bench_name, bench_group, bench_args):
 if __name__ == '__main__':
     parser = argparse.ArgumentParser(description='Profile RNNs')
 
+    default_groups = ['cnns', 'rnns']
+
     parser.add_argument('--seqLength', default='100', type=int)
     parser.add_argument('--numLayers', default='1', type=int)
     parser.add_argument('--inputSize', default='512', type=int)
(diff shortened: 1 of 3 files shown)


rnns/fastrnns/bench.py
```diff
@@ -145,41 +153,44 @@ def bench(rnn_runners, group_name, print_json=False, sep=' ', **params):
...
 
-    print_stderr('Benchmarking LSTMs...')
-    rnn_results = bench(get_rnn_runners(*rnns), 'lstm', **bench_args)
-    print_stderr('')
+    run_both = (args.cnns and args.rnns) or (not args.cnns and not args.rnns)
 
-    print_stderr('Benchmarking ResNets...')
-    cnn_results = bench(get_rnn_runners(*cnns), 'resnet', **bench_args)
-    print_stderr('')
+    if run_both:
```
zou3519: `run_both` seems redundant and won't scale well if we have more than two groups. Can't we do something like the following:
```
results = ... # Whatever the empty state is, I don't remember what this data structure is
if should_bench_varlen_lstms:
    results.update(bench_group(vlrnns, 'variable-length sequence LSTM', 'vl_lstm', bench_args))
if args.rnns:
    results.update(bench_group(rnns, 'LSTM', 'lstm', bench_args))
if args.cnns:
    results.update(bench_group(rnns, 'LSTM', 'lstm', bench_args))
```

wanchaol: this is surely something we can do, the only thing for this is, if we don't pass any args to specify which group, we should run all groups of the benchmarks, which means we need to have something to know if we have `not arg.cnns and not arg.rnns` to be True

zou3519: Hmm one way to do this is to set default arguments for args.cnns and args.rnns via ArgParser

according to https://docs.python.org/3/library/argparse.html something like the following should work:
```
parser.add_argument('bar', nargs='*', default=[1, 2, 3], help='BAR!')
```

wanchaol: Nice that should work 👍

PR #71, opened 2020-09-02
make MoCO model not distributed across multiple GPUs

nickgg:
The moco model is defined as a distrubuted NN, but we fix it to a single GPU in the benchmarks. Removing all the cross GPU communication for simplicitys sake.

### DIFF
--- a/models/moco/hubconf.py
+++ b/models/moco/hubconf.py
@@ -13,11 +13,9 @@ import torch
 import torch.nn as nn
 import torch.nn.parallel
 import torch.backends.cudnn as cudnn
-import torch.distributed as dist
 import torch.optim
 import torch.multiprocessing as mp
 import torch.utils.data
-import torch.utils.data.distributed
 import torchvision.transforms as transforms
 import torchvision.datasets as datasets
 import torchvision.models as models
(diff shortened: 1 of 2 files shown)


wconstab:
oh, can you also fix default device?  And check off the box here: https://github.com/pytorch/benchmark/projects/1

wconstab:
and make sure to raise NotImplemeted for jit?  It looks like jit is silently disabled for now.

nickgg:
I could comment out the code rather than delete it, but if/when we want to do distributed benchmarks probably easier to grab the copy from the original repo anyway.

The calls to nn.dist (eg the gathers) won't work if its not a DDP.

What do you mean by fix the default device and check off the project?

wconstab:
ok, deleting or commenting is fine just wanted to ask.

oops, i sent the wrong link.  check the moco specific 'nit' in this issue https://github.com/pytorch/benchmark/issues/65

models/moco/hubconf.py
```diff
@@ -127,9 +121,6 @@ def train(self, niterations=1):
 
         Leave warmup to the caller (e.g. don't do it inside)        
         """
-        if self.device != "cuda":
```
wconstab, line 130: i'm confused - above you raise on CPU, but here you allow it?

nickgg: sorry I'm trying to do this quickly in between some other work, this is a typo.

wconstab: hah you and me both. i know the feeling..

xuzhao9:
Addressed, closing this PR.

PR #75, opened 2020-09-03
Wconstab/maskrcnn

wconstab:
Add mask-rcnn model, but only support train() on cuda no-jit.  Checks in mini coco dataset.  Could be better to check in only one iteration worth as pkl.


asuhan:
@wconstab Which files should I look at, have you modified anything from upstream?

bertmaher (approved):
wow, nice job getting this into the framework.  I looked through hubconf just as a quick once-over but didn't review in depth.  Lgtm though :)

bertmaher:
Yeah, I'd probably get eval() working first and then decide what bits of data to pickle.

asuhan (approved):
Looks good. I don't think tracing would work well since each example can have a different number of boxes, so I believe you'd get a trace which is invalid for some / most of the examples.

PR #79, opened 2020-09-04
compare.py: add option to specify output format (csv, table, md, json)

ZolotukhinM:


compare.py
```diff
@@ -5,11 +5,12 @@
 Result = namedtuple("Result", ["name", "base_time", "diff_time"])
 
 def get_times(pytest_data):
-    return {b["name"]: b["stats"]["mean"] for b in pytest_data["benchmarks"]}
+    return {b["name"]: b["stats"]["min"] for b in pytest_data["benchmarks"]}
```
bertmaher, line 8: I always feel weird about using `min` for benchmarking, but it has its advantages, so I'm cool with it.

bertmaher (approved):

compare.py
ZolotukhinM: Oh, I didn't notice I changed that, will undo to avoid creeping in undesired changes.

PR #81, opened 2020-09-08
Update test.py to use unittest

zdevito:
* #85 make this work without changing directories
* #84 Get rid of path manipulation
* #83 Model names are valid python identifiers
* #82 Add yolov3 benchmark
* **#81 Update test.py to use unittest**

This allows for:
* test filtering using -k
* parallel testing using pytest
* verbose output of what is being run
* other timing functionalities that pytest provides

### DIFF
--- a/test.py
+++ b/test.py
@@ -23,14 +23,10 @@ def _load_test(model_class, model_path, device):
     dir, name = os.path.split(model_path)
     name = re.sub('[^A-Za-z0-9_]+', '_', name)
 
-    _obj = None
     def model_object(self):
         if device == 'cuda' and not torch.cuda.is_available():
             self.skipTest("torch.cuda not available")
-        nonlocal _obj
-        if _obj is None:
-            _obj = model_class(device=device)
-        return _obj
+        return model_class(device=device)
 
     def example(self):
         with workdir(model_path):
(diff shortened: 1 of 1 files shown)


wconstab (commented):
Happy to merge this despite the following comments once CI passes, since i think it still adds important filtering functionality we'd benefit from asap.

However, with these changes it seems like test.py is approaching being a near copy of test_bench.py but with subtle differences (the latter using benchmark framework instead of unittest, but both creating similar metaprogramming layers).  I'd like to see if we can combine into one script ultimately.  Any reason to use both unittest and pytest-benchmark frameworks?

zdevito:
I think it makes sense to keep `test.py` as the way to make sure the models are running correctly, and `test_benchmark.py` as a way to time them accurately. For instance, when adding a model, I don't want to wait a long time for it to accurately measure stuff.

wconstab:
> I think it makes sense to keep `test.py` as the way to make sure the models are running correctly, and `test_benchmark.py` as a way to time them accurately. For instance, when adding a model, I don't want to wait a long time for it to accurately measure stuff.
agreed about separate use cases, but i think one script would accomplish both.  for instance, `python test_bench.py` could have different behavior from `pytest test_bench.py`, and/or cmd line args could differentiate whether benchmarking vs sanity is performed

wconstab (approved):

zdevito:
fyi - these are ghstack commits so pressing merge on them doesn't merge them to master, but the author can land them with `ghstack land`

wconstab:
ahh i guess it would have been a good time to learn what ghstack does. i read the ezyang repo readme.

PR #86, opened 2020-09-09
Add yolov3 benchmark

zdevito:
* #87 Model names are valid python identifiers
* **#86 Add yolov3 benchmark**


wconstab (approved):
I didn't look inside training_loop() to see how wasteful it is to do that over and over in benchmark case, but i'm approving for now to unblock

PR #87, opened 2020-09-09
Model names are valid python identifiers

zdevito:
* **#87 Model names are valid python identifiers**
* #86 Add yolov3 benchmark

Change is rename only.


wconstab (commented):
hmm i guess this is a good change, but I think it's going to break our scuba datasets.

i'm ok with that, i think.  i'd actually like to take this opportunity if we're doing it, to make 'nice' names for each model (something the user types in their hubconf.py with a name() api) so that the printed tables and scuba data is easier to read, with consistent upper/lowercase and names limited in length.  thoughts?

zdevito:
Yeah, this change is necessary because I need each folder to become a valid python package in the next patch.

PR #88, opened 2020-09-09
Get rid of path manipulation

zdevito:
* **#88 Get rid of path manipulation**


zdevito:
This diff has more renames than the default github rename detection can handle, so to view a cleaner diff you might have to load it locally and up the rename limit.

test_bench.py
```diff
@@ -15,41 +15,37 @@
...
     """Constructs a model object for pytests to use.
     Any pytest function that consumes a 'modeldef' arg will invoke this
     automatically, and reuse it for each test that takes that combination
     of arguments within the module.
 
     If reusing the module between tests isn't safe, change 'scope' parameter.
     """
-    hubconf_file = 'hubconf.py'
+    Model, model_path = model
     with workdir(model_path):
```
wconstab, line 48: will this be needed once your changes to sanitize path manipulation are in?

i guess it's still good to make sure each model runs in a fresh CWD even if we don't gaurantee it runs in its preferred CWD

wconstab:
Can you also update ADDING_MODELS.md?  either in this PR or the next one, to reflect the latest

wconstab (approved):

test_bench.py
zdevito: It won't be necessary after the next PR, I will likely change the iterator to no longer return the model_path, and instead add a `benchmark_name` property to the Model object.

zdevito:
I will update the ADDING_MODELS.md in a new patch that also removes the directory iteration stuff.

PR #89, opened 2020-09-09
make this work without changing directories

zdevito:
* **#89 make this work without changing directories**


wconstab (commented):
should the test_bench.py 'with workdir...' thing be removed now?  (see corresponding comment on #88)

wconstab (approved):
ok LGTM saw your response on #88

PR #97, opened 2020-09-11
Move torch version check to test json and update repo fields

wconstab:
- change incorrect/confusing 'pytorch_commit_*' vars to git_* vars
  since the vars are used in repos such as benchmark
- record torch pkg versions explicitly regardless of repo used
  for benchmark runs


conftest.py
```diff
@@ -1,5 +1,8 @@
+import os
 import pytest
 import torch
+import torchtext
+import torchvision
```
malfet, line 5: Is this the intent to throw an exception if torchtext or torchvision is not found?

May be it's better to move those import inside pytest_benchmak_update_machine_info?
```suggestion
```

conftest.py
```diff
@@ -21,3 +24,10 @@ def set_fuser(fuser):
 
 def pytest_configure(config):
     set_fuser(config.getoption("fuser"))
+
+def pytest_benchmark_update_machine_info(config, machine_info):
+    machine_info['pytorch_version'] = torch.__version__
+    machine_info['torchtext_version'] = torchtext.__version__
+    machine_info['torchvision_version'] = torchvision.__version__
```
malfet, line 31: Why not trying to import those only inside the function and throw an just fill the version as not installed if import fails?
```suggestion
    try:
      import torchtext
      machine_info['torchtext_version'] = torchtext.__version__
    catch ImportError:
      machine_info['torchtext_version'] = '*not-installed*'
     try:
      import torchvision
      machine_info['torchvision_version'] = torchvision.__version__
    catch ImportError:
      machine_info['torchvision_version'] = '*not-installed*'    
```

wconstab:
@malfet Yea, i was going back and forth about how to do this but I like your way.

ZolotukhinM (approved):
Looks good to me!

PR #107, opened 2020-09-14
cleanup workdir, and document new model structure

zdevito:
* **#107 cleanup workdir, and document new model structure**

* remove workdir and module_path since they are no longer necessary
* document how to put models in the new structure, and how to
  fixup imports and path.


wconstab (approved):
LGTM thanks!

PR #109, opened 2020-09-15
Wconstab/maskrcnn2

wconstab:


.circleci/config.yml
```diff
@@ -13,6 +13,9 @@ jobs:
       - run:
           name: Install Conda
           command: ./scripts/install_basics.sh
+      - run:
```
zdevito, line 16: Why did this become necessary? Hitting the nvidia servers every time will make this more flaky, right?

zdevito (approved):
Thanks!

.circleci/config.yml
wconstab: i needed nvcc.

ultimately i disabled building/using apex/amp due to it introducing a memory leak (and it took a while to compile too annoyingly)

forgot if anything else in the repo needed nvcc but i kind of think so.

wconstab: not sure how much to worry about hitting the nv servers; but i'd be up for an alternative way if you have a suggestion.  this script/logic is copied from somewhere in pytorch repo, i thought it was already being used for those CI builds.

Issue #108, opened 2020-09-14
Add DLRM to benchmark


wconstab:
Original model source
https://github.com/facebookresearch/dlrm

looks like this one contains most (all?) the model code inside one file https://github.com/facebookresearch/dlrm/blob/master/dlrm_s_pytorch.py

the __main__ function should be a good indicator how to set up the model for training or eval, but it has a lot of extra baggage that can be removed when creating the benchmark version, such as profiling, onnx, save/load, etc. 

Also, note that pytorch benchmark repo was just reorganized.  @zdevito is updating ADDING_MODELS.md with some updated instructions asap.  Nothing changes for install.py but the hubconf.py is now named __init__.py, and there are more rules now about making no assumptions about current working directory for relative paths to load files.

PR #112, opened 2020-09-16
Adding Model DLRM

nikithamalgifb:
Summary:
========
This PR adds the model DLRM to torchbenchmark models. This model doesn't support JIT yet.
This PR also includes extra clean up in the tear_down api in test.py after the model has been run on the GPU. This makes sure that all the memory is freed on the gpu.

Fixes :  https://github.com/pytorch/benchmark/issues/108


test.py
```diff
@@ -20,6 +21,14 @@ def setUp(self):
     def tearDown(self):
         if 'cuda' in str(self):
             memory = torch.cuda.memory_allocated()
+            gc.collect()
+            # Deleting any tensors which were not picked up by GC
+            for obj in gc.get_objects():
+                try:
+                    if torch.is_tensor(obj) or (hasattr(obj, 'data') and torch.is_tensor(obj.data)):
+                        del obj
```
wconstab, line 29: this is actually a fallacy i think. 

gc.collect() should cause all 'deletable' objects to be deleted.  the ones that remain can't be deleted because someone holds references to them.  getting 'obj' from gc.get_objects() actually creates a new reference to the obj, so if it had 1 refcount before, now it has 2.  then del obj just sets it back to 1.  so, no benefit, i think.

i think adding gc.collect to master makes sense.  

the gc.get_objects part should be just a diagnostic though, for debugging.  to help you identify what holds the references to 'obj' that kept it from being collected.   not sure we should commit that part at all, or at least it should be separated out as a utility function or something...

wconstab (changes requested):
overall looks good, esp. __init__.py, just need to address the couple comments i made

test.py
nikithamalgifb: This makes sense. The del obj actually didn't do anything. I think this was my misunderstanding that it helped.  I removed this part.

torchbenchmark/models/dlrm/__init__.py
```diff
@@ -0,0 +1,199 @@
...
+            'arch_mlp_top': "4-2-1",
+            'mini_batch_size': 2,
+            'num_indices_per_lookup': 10,
+            'num_indices_per_lookup_fixed': True,
+            'numpy_rand_seed': 123,
+            'arch_sparse_feature_size': 2,
+        })
+
+        ### some basic setup ###
+        np.random.seed(self.opt.numpy_rand_seed)
```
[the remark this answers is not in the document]
nikithamalgifb: Ok, I moved this to the top of the file.

test.py
```diff
@@ -11,6 +11,7 @@
 import re, sys, unittest
 import os.path
 import torch
+import gc
 
 class TestBenchmark(TestCase):
     def setUp(self):
```
wconstab, line 17: can you add gc.collect() before this call to memory_allocated()?

this is because we are doing an assert equal rather than an assert less-equal.  It is possible that there is some memory allocated here which simply hasn't been collected yet but is collectible.  I observed on case in debugger where assertion failed with self.memory > memory.

test.py
```diff
@@ -20,6 +21,8 @@ def setUp(self):
     def tearDown(self):
         if 'cuda' in str(self):
             memory = torch.cuda.memory_allocated()
+            gc.collect()
```
wconstab, line 24: and try removing empty_cache, and moving gc.collect() above one line.  this is to make sure any collectible tensors are collected _before_ we take the measurement not after.

wconstab (approved):

Issue #93, opened 2020-09-09
Add StarGAN model to benchmark repo


wconstab:
We're building a suite of pytorch benchmarks by forking popular open-source models and modifying them to conform to a common API that facilitates a central tool installing, running, and collecting measurements.

Benchmark Repo: https://github.com/pytorch/benchmark
Instructions: https://github.com/pytorch/benchmark/blob/master/models/ADDING_MODELS.md
Source model: https://github.com/zou3519/stargan

The StarGAN model has already been forked and modified in a previous phase of this process, but the API (install.py, hubconf.py) had not yet been standardized.  Copy the repo below and follow the latest instructions in the playbook to add the new API.  Note, the 'install.sh' and 'run.sh' scripts added in the previous effort would serve as good hints on how to proceed.

note in this case it looks like there are no steps to do for installation, but the install.py file should still be present as a no-op for the API sake.

EdwardTyantov:
@wconstab pls review PR

PR #119, opened 2020-09-21
Add stargan model

EdwardTyantov:
https://github.com/pytorch/benchmark/issues/93


torchbenchmark/models/pytorch_stargan/requirements.txt
```diff
@@ -0,0 +1,2 @@
+# Nothing else aside from PyTorch!
+tensorflow>2.0
```
wconstab, line 2: what's tensorflow used for?

torchbenchmark/models/pytorch_stargan/main.py
```diff
@@ -0,0 +1,133 @@
...
+    parser.add_argument('--lr_update_step', type=int, default=1000)
+
+    # Extra arguments, not in the original stargan repo.
+    # --debug reference.out saves a file called "reference.out" with a result tensor
+    # from the last iteration of the model.
+    parser.add_argument('--debug', type=str, default='')
+    parser.add_argument('--deterministic', type=bool, default=False)
+    parser.add_argument('--should_script', type=bool, default=False)
+
+    config = parser.parse_args()
```
wconstab, line 126: this actually threw me a problem. we should figure out a way to test this in ci.  
when i ran locally `python test.py -k stargan` - the unittest parser correctly reads the -k argument and filters to only run your model.  but then the parser here trips over the -k that's still part of sys.argv and throws.

what i recommend doing is, 
L64: def parse_config(args=[])
L126: config = parser.parse_args(args)

that way by default all the defaults are used, but your utility can still be overridden if someone wanted to.

PR #131, opened 2020-09-25
Wconstab/score

wconstab:


score/generate_score_config.py
```diff
@@ -0,0 +1,121 @@
...
+its direct siblings.
+- if specific benchmark configurations are omitted under a benchmark name,
+all configurations present in the normalization data json are weighted equally
+
+Normalization Data
+------------------
+Used to 'fill in the gaps' in the human written specification. 
+
+Benchmark configurations (train/eval, device, compiler/runtime) present in 
+this data are frozen into the configuration and weighted equally if weights
```
bwasti, line 36: typo? "weights specific weights"

score/compute_score.py
```diff
@@ -0,0 +1,76 @@
...
+        benchmark_score = weight * math.log(norm / measured_mean)
+        # print(f"{name}: {benchmark_score}")
+        score += benchmark_score
+
+    score = target * math.exp(score)
+    assert abs(weight_sum - 1.0) < 1e-6, f"Bad configuration, weights don't sum to 1, but {weight_sum}"
+    return score
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description=__doc__)
```
bwasti, line 36: should we be able to specify device?

score/generate_score_config.py
wconstab: oops probably, ill fix

score/compute_score.py
wconstab: what do you mean? as in, cpu vs gpu? or as in, 'i ran this on a M40 and some CPU'?  what would be the point of it, to do some per-device normalization, or just to assert that you used the same device as the normalization data?

score/generate_score_config.py
```diff
@@ -0,0 +1,119 @@
...
+def generate_bench_cfg(spec, norm, target):
+    cfg = {
+        'target': target,
+        'benchmarks': {},
+    }
+    benchmark_names = [b['name'] for b in norm['benchmarks']]
+    benchmark_norms = {b['name']: b['stats']['mean'] for b in norm['benchmarks']}
+
+    assert len(spec['heirarchy']) > 0, "Must specify at least one category"
+    category_weight = 1.0 / len(spec['heirarchy'])
```
zdevito, line 60: hierarchy

score/torchbench_0.0.yaml
```diff
@@ -0,0 +1,227 @@
...
+  test_eval[attention_is_all_you_nee...-cpu-eager]:
+    norm: 0.8630710880000152
+    weight: 0.006944444444444444
+  test_eval[attention_is_all_you_nee...-cpu-jit]:
+    norm: 0.8489351521999652
+    weight: 0.006944444444444444
+  test_eval[attention_is_all_you_nee...-cuda-eager]:
+    norm: 0.18483205416672868
+    weight: 0.006944444444444444
+  test_eval[attention_is_all_you_nee...-cuda-jit]:
```
zdevito, line 41: Is this input data? Should it be using these abbreviated names?

score/score.yml
```diff
@@ -0,0 +1,38 @@
+---
```
zdevito, line 1: This kind of information should probably just be part of the `Model` object to allow for filtering when doing experiments. The hierarchy doesn't feel compelling to me because already there are a few shoehorns: 'recommendation, recommendation'. 'other, other task'. Perhaps just add a `keywords` field to the Model objects?

zdevito (approved):
Thanks for putting this together. Let's get it landed so we can start recording data. My overall impression is that the way you specify the way in which the score is computed can be simplified a lot. After reading this code along with its docs, I feel like I would not be able to meaningful generate a new weighting or understand how the initial weighting was created. In particular:
* There seems to be a lot of scripts to generate intermediate files whose only purpose is inputs to other scripts. I feel like a python function `compute_score_v1(input_timings) -> value` would be much less machinery, be easier to understand, and be easier to modify in the future.
* Hierarchically describing the models feels overly complicated. It is not clear what the levels of the hierarchy are and already some models don't fit the heirarchy.
* I don't understand our relative weights of CPU, CUDA, jit, eager. You might imagine that someone wants to see the improvement for JIT on CUDA, but it doesn't seem easy to extract the relative weighting of just the benchmarks.

score/torchbench_0.0.yaml
wconstab: well, these are actually the names we are storing in the benchmark .json.  we may want to change that.

score/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+## New score versions
+Periodically, as more workloads have been added to the torchbenchmark suite, or as changes to
+relative weights or categories have been proposed, a new score configuration should be generated
+rather than modifying an existing score definition. 
+
+See `python generate_score_config.py -h` 
+
+## Issues and Next Steps
+For accurate score comparisons, measurements should be computed on the same machine
```
zheng-xq, line 28: This sounds fine for now. But in our past experience, this requirement is hard to maintain in large scale across a long period of time. As we get more benchmarks, we will likely use more machines of the same type, and we might have to migrate to a different pool from time to time. 

So a good approach is to offset the scores to stay the same during transition. We don't have to do it now, but it might be worth mentioning in the future directions.

score/compute_score.py
```diff
@@ -0,0 +1,76 @@
...
+        measured_mean = measured_mean[0]
+        if fake_data is not None and name in fake_data:
+            # used for sanity checks on the sensitivity of the score metric
+            measured_mean = fake_data[name]
+        benchmark_score = weight * math.log(norm / measured_mean)
+        # print(f"{name}: {benchmark_score}")
+        score += benchmark_score
+
+    score = target * math.exp(score)
+    assert abs(weight_sum - 1.0) < 1e-6, f"Bad configuration, weights don't sum to 1, but {weight_sum}"
```
zheng-xq, line 32: Instead of asserting, will it be easier just to normalize?

score/compute_score.py
```diff
@@ -0,0 +1,76 @@
...
+
+from tabulate import tabulate
+
+def compute_score(config, data, fake_data=None):
+    target = config['target']
+    score = 0.0
+    weight_sum = 0.0
+    for name in config['benchmarks']:
+        cfg = config['benchmarks'][name]
+        weight, norm = cfg['weight'], cfg['norm']
```
zheng-xq, line 19: It might also be good to check that weights are non-negative to avoid numerical instability in the previous calculations.

wconstab:
> There seems to be a lot of scripts to generate intermediate files whose only purpose is inputs to other scripts. I feel like a python function compute_score_v1(input_timings) -> value would be much less machinery, be easier to understand, and be easier to modify in the future.
Yes- overly complex, probably. And I like the idea of hiding things behind a named command like 'compute_score_v1'.  We can make this change later without affecting the numerics, so i'm ok with disregarding for now.

> Hierarchically describing the models feels overly complicated. It is not clear what the levels of the hierarchy are and already some models don't fit the hierarchy.
I'm resigned to an imperfect hierarchy (e.g. the 'other' stuff) if we have one at all.  I don't want to try to have a comprehensive one, but I thought some structure was necessary to enable
  * weights for most leaf nodes don't have to be manually specified
  * manual weights can be specified at some level of the tree  (e.g. we can say recommenders gets an unfair share)
  * models within the same category have to share a fixed amount of the overall score
I'd be open to brainstorm other ways to accomplish that, especially taking the below into account

>I don't understand our relative weights of CPU, CUDA, jit, eager. You might imagine that someone wants to see the improvement for JIT on CUDA, but it doesn't seem easy to extract the relative weighting of just the benchmarks.
Yea, I struggled with how to aggregate top level score(s) and the choice of tree structure seems to hurt there.  Otoh, the tree is only needed for that top lvel score, and we can potentially also compute sub-scores for cuda vs cpu, jit vs eager, train vs eval, etc.  I'd also like to brainstorm this angle- what set of metrics would be useful to generate?  We can work backwards from that to an implementation that fits.

score/compute_score.py
zheng-xq: I am fine with either way that different devices have their own normalization data or not.  But it is important for us to track them on different devices consistently. One change could be neutral on Volta, but causes regressions on Ampere.

zheng-xq (approved):

score/README.md
wconstab: yea, i agree that we can't really limit to a single machine.  I have been thinking about defining a machine spec (cpu, gpu type, memory cap, potentially kernel version, etc.) and making sure that part matches even though the hostname is different.

score/compute_score.py
wconstab: well, the intent here was just error checking- in the current implementation the score generator should produce weights that sum to 1 so i wanted to make sure something wasn't dropped, rather than just recompute a norm.

i think i'll leave it like this for now as we may make more radical changes to the way things are computed following the rest of the feedback.

score/compute_score.py
wconstab: yes, i was thinking of adding numerical stability check on all of [weight, norm, measurement].  I can catch weight >0 at the same time.  What do you think is a sensible epsilon value?  I think 1e-6 is fine for weight.  For measure and norm, the expected range could be anywhere from 1e3 to 1e9 i think, but we probably want to at least warn if not assert on norm / weight > 100 or something.

score/torchbench_0.0.yaml
```diff
@@ -0,0 +1,227 @@
+benchmarks:
```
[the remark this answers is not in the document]
wconstab: yes, good point. that data is present in the input .json file, but not captured in the score .yml.  i planned to add it and also make an assert that the new measure data matches.

score/README.md
wconstab: oh, and my original intent was that a score configuration was bound to a machine type, but every quarter we might take the latest pytorch release, more benchmark models, and possibly a different machine type, and use that to compute a new normalization file with the target set to match what the latest target was on the old score.

it would also be possible to support multiple machine types at the same time, just by maintaining different sets of norms for each machine type.

Issue #142, opened 2020-10-21
Modify fastNLP to work with latest and older torchtext releases


wconstab:
torchtext 0.8.0 release moved AG_NEWS from datasets.experimental. to datasets.  fastNLP should be able to run under new and old torchtext for the sake of running historical builds, so we can check the torchtext version in the fastNLP __init__.py and conditionally import the right name.

xuzhao9:
Cherry-picked [the fix](https://github.com/pytorch/benchmark/commit/ac7a531235509dea0cf7579d60aa62f7ba980172) to 0.1 branch

PR #143, opened 2020-10-21
Modify AG_NEWS import to work under new and old torchtext #142

wconstab:
torchtext >=0.8.0 moved AG_NEWS from experimental to mainline

### DIFF
--- a/torchbenchmark/models/fastNLP/__init__.py
+++ b/torchbenchmark/models/fastNLP/__init__.py
@@ -10,9 +10,10 @@ import numpy as np
 import pickle
 import os
 
-if torchtext.__version__ >= "0.8.0":
+try:
     from torchtext.datasets import AG_NEWS
-else:
+except ImportError:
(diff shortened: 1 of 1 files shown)


malfet (approved):

torchbenchmark/models/fastNLP/__init__.py
```diff
@@ -11,6 +10,11 @@
 import pickle
 import os
 
+if torchtext.__version__ >= "0.8.0":
+    from torchtext.datasets import AG_NEWS
+else:
+    from torchtext.experimental.datasets import AG_NEWS
+
```
malfet, line 17: Rather than hardcoding string comparison, why not just try to import it from different location
```suggestion
try:
    from torchtext.datasets import AG_NEWS
except ImportError:
    # In older version of torchtext AG_NEWS is inside experimental module
    from torchtext.experimental.datasets import AG_NEWS

```

PR #144, opened 2020-10-23
Improve the compute score script to accept json directory or json file without requiring a config.

xuzhao9:
* **#144 Improve the compute score script so that it accepts json directory or json file without requiring a config.**


score/compute_score.py
```diff
@@ -34,16 +38,37 @@ def compute_score(config, data, fake_data=None):
...
     parser.add_argument("--benchmark_data_file",
         help="pytest-benchmark json file with current benchmark data")
     parser.add_argument("--benchmark_data_dir",
         help="directory containing multiple .json files for each of which to compute a score")
     parser.add_argument('--hack_data', nargs=2, action='append', help="keyword to match benchmark names, and multiplicative factor to adjust their measurement")
     args = parser.parse_args()
-    with open(args.configuration) as cfg_file:
-        config = yaml.full_load(cfg_file)
+
+    if args.configuration is not None:
```
wconstab, line 50: does it make sense to reduce lines of code by putting the config-file generation into the existing code blocks 
if args.benchmark_data_file is not None:
elif args.benchmark_data_dir is not None:
?

You can still have a check at line 53
else:
   if args.benchmark_data_file is None and args.benchmark_data_file is None, raise Exception "invalid arg.."

wconstab (approved):
LGTM! minor nit comment inline

score/compute_score.py
xuzhao9: That makes sense. I have updated the code.

PR #157, opened 2020-11-20
Added MAML model

Chillee:
Finally?


torchbenchmark/models/maml/__init__.py
```diff
@@ -0,0 +1,85 @@
...
+from pathlib import Path
+
+torch.manual_seed(1337)
+random.seed(1337)
+np.random.seed(1337)
+torch.backends.cudnn.deterministic = True
+torch.backends.cudnn.benchmark = False
+
+class Model:
+    def __init__(self, device=None, jit=False):
```
wconstab, line 23: i forget if we had a reason for making the default device 'cpu'.  i think it was at least important to not make it gpu for some reason. did you see other models use None?

torchbenchmark/models/maml/__init__.py
```diff
@@ -0,0 +1,85 @@
...
+        if self.jit:
+            raise NotImplementedError()
+        return self.module, self.example_inputs
+
+    def eval(self, niter=1):
+        if self.jit:
+            raise NotImplementedError()
+        self.module.eval()
+        for _ in range(niter):
+            self.module.finetunning(*[i[0] for i in self.example_inputs])
```
wconstab, line 70: what's the structure of the example inputs? might be better to do the list comprehension outside and save the result as a separate inputs member, since 'eval' will get called multiple times by the benchmark driver and you'll end up timing that part.

also, finetunning is a funny typo right? but not your code, their code?

wconstab (approved):

torchbenchmark/models/maml/__init__.py
Chillee: > what's the structure of the example inputs?

It's just 5 inputs, so I thought it was fine.

> also, finetunning is a funny typo right? but not your code, their code?

Yep :) I find that some researchers have a very very annoying habit of making a typo somewhere and just keeping it throughout their entire code.

PR #160, opened 2020-11-21
Generate config from benchmarks that run partial benchmarking suites.

xuzhao9:
Now generate_score_config.py can generate configs for benchmarks that doesn't run the entire testing suite.

### DIFF
--- a/score/generate_score_config.py
+++ b/score/generate_score_config.py
@@ -47,6 +47,7 @@ TODO
 import argparse
 import json
 import yaml
+from collections import defaultdict
 
 def generate_bench_cfg(spec, norm, target):
     cfg = {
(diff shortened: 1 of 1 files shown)


score/generate_score_config.py
```diff
@@ -105,6 +143,10 @@ def generate_bench_cfg(spec, norm, target):
         help="generated complete benchmark configuration")
     parser.add_argument("--target_score", default=1000,
         help="target score value given these normalizations and specifications")
+    parser.add_argument("--partial",
+                        action='store_true',
+                        help="generates partial config if the benchmark only runs part of the spec")
```
wconstab, line 148: i think we should take the opportunity to write a more clear explanation of what this is doing. even i forgot how this works at first.  we need to make the whole score thing more user friendly...

in this case what about updating the 'help' for partial to clarify a bit.

help="generates partial config if the benchmark only runs part of the spec.  normally, the spec is supposed to define the set of benchmarks that's expected to exist, and then the provided json data is expected to provide the norm values to match the spec.  To simplify debugging, and not for normal score runs, we allow a convenience for producing a score configuration that matches whatever json data is provided."

wconstab (approved):
take a look at my suggestion for the comment, make sure you agree with it or else make some changes so it's correct.  otherwise lgtm, until we can make this whole thing better.

PR #162, opened 2020-11-30
Add pytorch nightly benchmarking workflow.

xuzhao9:
This PR adds the nightly benchmark testing workflow.
It runs everyday on 8 am UTC (12 am PST), after the pytorch nightly build is pushed to conda.
For each workflow it runs 8 iterations and 20 rounds for each iteration.

To run this, the benchmark repo also needs to add secret SCRIBE_GRAPHQL_ACCESS_TOKEN and bm-ec2 self-hosted runner.


.github/scripts/run-benchmark.sh
```diff
@@ -0,0 +1,33 @@
...
+
+conda init bash; conda run /bin/bash
+
+if [[ -e ${NO_TURBO_FILE} ]]; then
+    sh -c "echo 1 > ${NO_TURBO_FILE}"
+fi
+
+pushd /workspace/benchmark
+
+for c in $(seq 1 $NUM_ITER); do
```
wconstab, line 25: just a couple of comments on the iterations/rounds.

- do you have a justification for running multiple (8) iterations? it will cost more, might not be needed?
- keep in mind that min-rounds is only minimum.  so if you set 20, some faster benchmarks will still run 100s of rounds.  This only matters later when uploading the data, need to be careful of not overwhelming the database. the uploader script has a limit currently.  just something to be aware of.

.github/scripts/run-docker.sh
```diff
@@ -0,0 +1,37 @@
...
+# Set the <memory, graphics> clock frequency
+# Need further study on how Nvidia card throttling affect overall performance variance
+sudo nvidia-smi -ac 5001,1500
+
+docker run \
+       --env GITHUB_RUN_ID \
+       --env SCRIBE_GRAPHQL_ACCESS_TOKEN \
+       --volume="${PWD}:/runner" \
+       --volume="${DATA_DIR}:/output" \
+       --gpus=all \
```
wconstab, line 32: wonder if we should actually limit to one or not?
in the runner we picked, there is only 1 gpu anyway.
most (all?) benchmarks have been configured to disable multigpu, but some might have been missed so behavior could change based on # visible gpu.

later, we might want to use 2 gpu for some multigpu benchmarking.  but maybe better to explicitly disable until we are ready to do that?

.github/scripts/upload-scribe.sh
```diff
@@ -0,0 +1,31 @@
...
+
+if [[ ! -v SCRIBE_GRAPHQL_ACCESS_TOKEN ]]; then
+    echo "SCRIBE_GRAPHQL_ACCESS_TOKEN is not set! Please check your environment variable. Stop."
+    exit 1
+fi
+
+pushd /workspace/benchmark/score
+
+# Generate the benchmark config
+python generate_score_config.py --specification score.yml \
```
wconstab, line 15: this doesn't look right to me. we shouldn't be generating a new score config on every night.  if we do, won't we just get a score of 'target_score' every night?

wconstab: fwiw, the score scripts are not user friendly.. but what i had in mind by designing this way was 
 - one time, admin would set up 'golden machine' and create score config on it
   - check this score config in, with a version number attached to its name
- then, anyone wanting to compute the score would use that file, use the same machine (or at least same type of machine), and run the same benchmarks.

.github/workflows/main.yml
```diff
@@ -0,0 +1,29 @@
+name: pytorch-nightly-benchmarking
+on:
+  schedule:
+    - cron: '0 8 * * *' # run at 8 AM UTC
+
+jobs:
+  build-docker:
```
wconstab, line 7: i'm curious about how you split build-docker and build-benchmark-docker.  What is the reason to split, and what happens in each?

also curious about what you plan to do with the docker images outside of this nightly job.  if we can reuse the same already-built docker image from multiple places, it is a win.  but if we build a new image every time we use one, i'm not sure what value docker provides?

.github/scripts/build-benchmark-docker.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/sh
+
+set -euo pipefail
+
+TODAY=$(date +"%Y%m%d")
+
+# Get the latest pytorch docker image
+TORCH_LATEST=$(docker images | grep "pytorch\s" | head -n 1)
```
wconstab, line 8: this is using the pytorch image you built in 'build-docker.sh'? 

i wonder why you build pytorch docker.  is it bc it already provides cuda or other deps?

if you just need cuda/deps but don't need pytorch in the base docker (bc you can install pytorch from conda or pip), then maybe it's easier to skip building pytorch and instead put 'from X' in your benchmark DOCKERFILE, where X is either an nvidia public docker image, or one of the pytorch docker images already in our docker hub?

.github/scripts/run-benchmark.sh
xuzhao9: Each iteration (20 rounds) costs about 50 mins to finish.
Therefore, 8 iterations is about 400 mins, or ~7 hours. This leaves us some time to bisect problematic PRs if we find some problem and we get enough data points to be confident that the benchmarking scores are stable.

.github/scripts/run-docker.sh
xuzhao9: Sure, we can limit current run to be just one gpu.

.github/scripts/build-benchmark-docker.sh
xuzhao9: I use pytorch docker because it packs nightly pytorch, torchtext, torchvision (from conda), as well as all their dependencies (including nvidia-cuda, cudnn, and so on). Therefore, I don't need to concern about any version or dependency problems, therefore avoids maintenance overhead (which is place on the pytorch dockerfile).
I believe it is still worthwhile to continue using it as a base image, if we need to customize pytorch version (not using nightly), we could first uninstall from conda, then install using pip.

.github/workflows/main.yml
xuzhao9: The reason to split is because I don't want to maintain the dependencies of pytorch/torchtext/torchvision myself (e.g. cudatoolkit, cudnn) as this part is already managed by pytorch Dockerfile.
build-benchmark-docker relies on build-docker, and installs all benchmark dependencies.

You are right that we need to rebuild docker every nightly test. But I still think it is worthwhile to use docker. First, it doesn't take much time (~10 min) comparing to the benchmark running time. Second, if we need to re-run or debug some benchmarks (say one of the models has regression), then we don't need to rebuild everything because we can just use the existing one. Also, we can control the number of history dockers (say past one month) to reproduce history performance data.

.github/scripts/upload-scribe.sh
xuzhao9: That is correct. We need to fix this problem by first determining the 'golden machine' and the config that it uses.

.github/scripts/build-benchmark-docker.sh
wconstab: i'm not sure what maintennence or dependency you mean.  if you just type 'conda install -c pytorch-nightly pytorch torchtext torchvision' doesn't that take care of deps?  

anyway, your approach seems ok just thought it might be more work than necessary

.github/scripts/run-benchmark.sh
wconstab: a related question is, what is the difference between doing 8 iterations of 20 rounds vs 1 iteration of 160 rounds?  

we should probably determine two things
 - how many rounds do we need for good signal?  (20? 40? 160?)
 - is there an important reason to break into multiple iterations? or just leave them together for simplicity?

.github/scripts/upload-scribe.sh
wconstab: ok. as long as you have a plan to deal with that.

.github/workflows/main.yml
wconstab: ok. i like having dockers.  i think we should make it easy for someone to find the history of dockers and clone one so they can benefit from your work.

i wonder if you need to build the pytorch docker yourself though?  can't you just use 'FROM' keyword and reference already-built pytorch nightly docker, built by existing pytorch CI?

.github/scripts/build-benchmark-docker.sh
xuzhao9: Just for record, currently pytorch doesn't build a nightly docker image so we have to build it from its Dockerfile. In the Dockerfile, pytorch team maintains all the dependencies such as cuda, cudnn, among other packages. We should rely on the correctness of the official Dockerfile instead of managing dependencies of our own.

.github/workflows/main.yml
xuzhao9: See above. There is no pre-existing built pytorch nightly docker, that's why we are building our own here.

.github/scripts/run-benchmark.sh
xuzhao9: Good question. For now we run for 5 iterations and 20 rounds to make sure the score is stable. But more testing is required to determine the minimum round/iterations required to get a reasonable score.

.github/scripts/run-docker.sh
```diff
@@ -0,0 +1,39 @@
...
+echo "Running pytorch-benchmark image ${TORCH_IMAGE_ID}"
+
+mkdir -p ${DATA_DIR}
+
+# Nvidia won't let this run inside docker
+# Make sure the Nvidia GPU is in persistence mode
+sudo nvidia-smi -pm ENABLED -i 0
+# Set the <memory, graphics> clock frequency
+# Need further study on how Nvidia card throttling affect overall performance variance
+sudo nvidia-smi -ac 5001,1500
```
wconstab, line 26: just thinking about keeping everything related to tuning in one place- 

is it possible to put all the nvidia, threads, cores, etc settings in the same script? or does it need to be split apart for technical reasons?

.github/scripts/upload-scribe.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/sh
+
+# [FB Internal] Upload every json file to Scribe
+
+# Use the config v0
+CONFIG_FILE=/config/config-v0-notacotron-noslomo.json
```
wconstab, line 6: i think it'd be good to delete the notacotron-noslomo part from the name.

the reason being, i think going forward any score version will include some subset of the models that are checked in, and we can keep the names simple.  

on the same token though, could you add a markdown document to describe the v0 score?  i think we should have a doc that shows what is supported for each score version, and explains the theory about it.  i can help with that if you want (can be a separate PR)

wconstab: oh and the doc should also describe the machine type and any other assumptions about how it's run and configured.

wconstab (approved):
looks great to me!

only thing i think we might want to tweak is NUM_ITER, but it doesn't seem bad to change that part later.

PR #165, opened 2020-12-02
Update run_bench_and_upload.sh

wconstab:
Pushing from PRs worked fine if the PR was from a branch on the pytorch/benchmark repo, but broke if the PR was from a forked repo.  This is becuase circle CI only provides the credentials that can push the data to github to the jobs that are from the trusted fork for security reasons.

Rather than fixing the logic for PRs I'm disabling it and only archiving for master.  The results from this CI job aren't critical anyway, and we're building new infra to run the real benchmarks so we'll revisit then.


xuzhao9 (approved):
LGTM

PR #166, opened 2020-12-03
Adjust iter to 1. Add job to upload data to pytorch_benchmark_data repo.

xuzhao9:


.github/scripts/upload-github.sh
```diff
@@ -0,0 +1,38 @@
...
+BENCHMARK_FILENAME=${GITHUB_SHA}_$(basename ${BENCHMARK_ABS_FILE})
+pushd ${BENCHMARK_DATA_DIR}
+mkdir -p ${SUBDIR}
+cp ${BENCHMARK_ABS_FILE} ${SUBDIR}/${BENCHMARK_FILENAME}
+ln -s -f ${SUBDIR}/${BENCHMARK_FILENAME} ${SUBDIR}/latest
+echo ${BENCHMARK_FILENAME} >> ${SUBDIR}/history
+
+git add ${SUBDIR}
+git commit -m"Add ${BENCHMARK_FILENAME} - to reproduce results, see pytorch and machine versions in .json."
+git push origin master
```
wconstab, line 36: i think it might be best to still put a check that the branch isn't a fork.

if you're sure that this script won't be run for any PR branches then it isn't necessary, but at least a comment would be good.  Explaining that git push will fail if the pytorchbot token isn't available, and it's only available to builds from the main pytorch fork.

.github/scripts/upload-github.sh
```diff
@@ -0,0 +1,38 @@
+#1/bin/sh
+
+set -eo pipefail
+
+# Upload the first json to the pytorch benchmark data repository
```
wconstab, line 5: what does 'first json' mean in this context? is this leftover from the time you ran 5 iter? or are there other nights' json files still in the dir, and you're grabbing the most recent one?

.github/workflows/main.yml
```diff
@@ -28,3 +28,8 @@ jobs:
       SCRIBE_GRAPHQL_ACCESS_TOKEN: ${{ secrets.SCRIBE_GRAPHQL_ACCESS_TOKEN }}
     steps:
       - run: bash ./.github/scripts/run-docker.sh /runner/.github/scripts/upload-scribe.sh
+  upload-github:
+    needs: run-benchmark
+    runs-on: [self-hosted, bm-ec2]
```
wconstab, line 33: i'm wondering if this approach of separate steps that run on the same label is robust?

if we add more jobs that use the runner, or if we add more runners for the job to share, will it be possible to hit either of these scenarios?
 - that 'run-benchmark' runs on worker0, then upload-scribe runs on worker1
 - a race condition where upload-scribe picks up a file generated by another job and stored on the worker?

.github/scripts/upload-github.sh
xuzhao9: No, one dir will only contain json files from the same nightly build. But if multiple iterations exist, I will only grab the first one. This remains true if the iteration number is 1, in which the first one is the only one.

.github/workflows/main.yml
xuzhao9: Good point. Unfortunately I don't find a way to assign runner on the workflow level. Therefore, if more runners are added to "bm-ec2" label it is possible for scenario 1) to happen.
However, for 2), since upload-scribe.sh uploads the file by GITHUB_RUN_ID, it is impossible to pick up json from another job.

wconstab: ok, it'd be good to confirm if the dependency gaurantees co-scheduling or not.  

if not, it could be that you just need to combine all the steps as part of one script so they get scheduled atomically for sure?

.github/scripts/upload-github.sh
```diff
@@ -0,0 +1,48 @@
...
+set -eo pipefail
+
+cat > ~/.netrc <<EOF
+machine github.com
+login pytorchbot
+password ${GITHUB_TORCHBENCH_TOKEN}
+EOF
+
+BENCHMARK_DATA_DIR=${HOME}/pytorch_benchmark_data
+BENCHMARK_DATA_REPO=https://github.com/wconstab/pytorch_benchmark_data
```
wconstab, line 12: this is just a placeholder, while you add the new repo right?

.github/scripts/run-docker.sh
```diff
@@ -7,17 +7,16 @@ CONFIG_VER=v0
 
 # Run Parameters
 RUN_SCRIPT=$1
-TODAY=$(date +"%Y%m%d")
 CONFIG_DIR=${PWD}/.github/configs/${CONFIG_VER}
 CONFIG_ENV=${CONFIG_DIR}/config-${CONFIG_VER}.env
-DATA_DIR=${HOME}/benchmark-results/${TODAY}_gh${GITHUB_RUN_ID}
```
wconstab, line 13: what is the contract around the 'gh' folder- it will be created fresh for each job, so TODAY is not needed?  or TODAY was more for archive, and not important now that you upload to github?

.github/workflows/main.yml
xuzhao9: I asked a question here: https://github.community/t/does-gh-actions-co-schedule-dependent-jobs-on-the-same-runner/146913. Will also talk with github people when I have a chance.

.github/scripts/run-docker.sh
xuzhao9: Yes, it will be created for each job so TODAY is not needed.

.github/scripts/upload-github.sh
xuzhao9: Yes, I will defer this PR until we figure out the new github token

PR #170, opened 2020-12-07
Fix v0 config. Clear docker cache before building pytorch docker.

xuzhao9:


wconstab (commented):
i understand that you added the prune to fix the docker cache issue. 

what is the score change? i thought you already put the new config in.  also the diff says you deleted 41k lines and replaced with 200 lines.  manually opening the file looks like the 200 line one is about right..

xuzhao9:
> i understand that you added the prune to fix the docker cache issue.
> 
> what is the score change? i thought you already put the new config in. also the diff says you deleted 41k lines and replaced with 200 lines. manually opening the file looks like the 200 line one is about right..

In my previous PR(#169), I actually checked-in the benchmarking result json, not the config json. This PR is to fix this problem.

PR #178, opened 2020-12-21
Fix a connection error by requesting torchtext Multi30k dataset.

xuzhao9:


wconstab (approved):
so basically this puts the file already inside the docker container where the test will find it and not try to download? sgtm. thanks!

xuzhao9:
> so basically this puts the file already inside the docker container where the test will find it and not try to download? sgtm. thanks!

The test will only download the data when the local version is not available. So this patch should be good for now.

PR #184, opened 2021-01-08
Fix a bug in pytorch_struct that raises NotImplemented error in the latest pytorch-nightly 20210107

xuzhao9:


wconstab (approved):
Did you also make this change on the master branch?  might be best practice is to merge first to master then cherry pick to 0.1, but either way as long as it gets on both branches its ok

xuzhao9:
> Did you also make this change on the master branch? might be best practice is to merge first to master then cherry pick to 0.1, but either way as long as it gets on both branches its ok

This time I will submit separate PRs. Next time I will submit to master first then cherry-pick to 0.1.

PR #185, opened 2021-01-10
Misc requirements and debuggability fixes

jansel:
This is a collection of minor fixes and quality of life improvements I made while making the install script work for python 3.8 (some models still don't run in 3.8, but install now completes).

Tested with:
```
PYTHON_VERSION=3.8 ./scripts/recreate_conda_environment.sh
PYTHON_VERSION=3.7 ./scripts/recreate_conda_environment.sh
conda activate torchbenchmark
pytest test.py
./compare.sh
```

### DIFF
--- a/compare.sh
+++ b/compare.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-
+rm -f old.json te.json
 pytest test_bench.py -k cuda-jit --fuser old --benchmark-json old.json
 pytest test_bench.py -k cuda-jit --fuser te --benchmark-json te.json
 python compare.py old.json te.json
--- a/torchbenchmark/__init__.py
+++ b/torchbenchmark/__init__.py
@@ -25,10 +25,15 @@ def _test_https(test_url='https://github.com', timeout=0.5):
 
 def _install_deps(model_path):
     if os.path.exists(os.path.join(model_path, install_file)):
-        subprocess.check_call([sys.executable, install_file], cwd=model_path)
+        try:
+            subprocess.check_call([sys.executable, install_file], cwd=model_path)
+        except subprocess.CalledProcessError:
+            print(f"Error while running {model_path}/{install_file}")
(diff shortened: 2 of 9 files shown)


compare.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
-
+rm -f old.json te.json
```
wconstab, line 2: huh, was it appending not overwriting?

torchbenchmark/__init__.py
```diff
@@ -25,11 +25,16 @@ def _test_https(test_url='https://github.com', timeout=0.5):
 
 def _install_deps(model_path):
     if os.path.exists(os.path.join(model_path, install_file)):
-        subprocess.check_call([sys.executable, install_file], cwd=model_path)
+        try:
+            subprocess.check_call([sys.executable, install_file], cwd=model_path)
+        except subprocess.CalledProcessError:
+            print(f"Error while running {model_path}/{install_file}")
+            sys.exit(-1)
```
wconstab, line 32: in your comments i thought you said your changes allowed the install to complete even if some models didn't work in py3.8.  this looks like it just bails and later models wouldn't be attempted?

compare.sh
jansel: This script is running without `-e` so failures are ignored (sometimes tests fail, so this is the right behavior).

I wanted to be paranoid and doubly make sure there wasn't some failure that causes us to pick up old data in a later step.

(This is more of an "avoid nasty gotchas" change than a bugfix.)

wconstab (approved):
LGTM thanks for the cleanups!

torchbenchmark/__init__.py
jansel: The old version would also bail (by raising an exception), but it was impossible to tell which model failed to install.  This just prints out the name of the model causing the failure.

PR #187, opened 2021-01-11
Switch BERT, pytorch_struct, and fastNLP to develop mode

jansel:
I got hit by the "gotcha" where I was editing benchmark files and nothing was changing.  Turns out you need to re-run `install.py`to make changes visibile for these benchmarks because we install them into the conda environment.  Hopefully this fix saves someone else a headache.


wconstab (approved):
Good catch. I guess they should all be in develop mode, but I never tried to check/enforce that.

PR #188, opened 2021-01-12
Make benchmarks run quietly for inference

jansel:
I was running some tests on benchmarks and was bombarded by pages and pages of printouts which made it hard to see the data I wanted.   This PR removes all those printouts (at least for inference). 

The following now prints nothing:
```
for benchmark_cls in list_models():
    for device in ("cpu", "cuda"):
        try:
            benchmark = benchmark_cls(device=device)
            model, example_inputs = benchmark.get_module()
            model(*example_inputs)
        except NotImplementedError:
            pass
```

### DIFF
--- a/torchbenchmark/models/maskrcnn_benchmark/maskrcnn_benchmark/data/build.py
+++ b/torchbenchmark/models/maskrcnn_benchmark/maskrcnn_benchmark/data/build.py
@@ -127,7 +127,7 @@ def make_data_loader(cfg, is_train=True, is_distributed=False, start_iter=0, is_
         num_iters = None
         start_iter = 0
 
-    if images_per_gpu > 1:
+    if images_per_gpu > 2:
         logger = logging.getLogger(__name__)
         logger.warning(
             "When using more than one image per GPU you may encounter "
--- a/torchbenchmark/models/maskrcnn_benchmark/maskrcnn_benchmark/data/build.py
+++ b/torchbenchmark/models/maskrcnn_benchmark/maskrcnn_benchmark/data/build.py
@@ -127,19 +127,6 @@ def make_data_loader(cfg, is_train=True, is_distributed=False, start_iter=0, is_
         num_iters = None
         start_iter = 0
 
-    if images_per_gpu > 2:
-        logger = logging.getLogger(__name__)
-        logger.warning(
-            "When using more than one image per GPU you may encounter "
-            "an out-of-memory (OOM) error if your GPU does not have "
-            "sufficient memory. If this happens, you can reduce "
-            "SOLVER.IMS_PER_BATCH (for training) or "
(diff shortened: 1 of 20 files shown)


torchbenchmark/models/maskrcnn_benchmark/maskrcnn_benchmark/data/build.py
```diff
@@ -127,7 +127,7 @@ def make_data_loader(cfg, is_train=True, is_distributed=False, start_iter=0, is_
         num_iters = None
         start_iter = 0
 
-    if images_per_gpu > 1:
+    if images_per_gpu > 2:
```
wconstab, line 130: is this just a hacky way of disabling the print below? maybe just comment it all out?

torchbenchmark/models/yolov3/yolo_models.py
```diff
@@ -103,13 +103,13 @@ def create_modules(module_defs, img_size, cfg):
...
                                 stride=stride[yolo_index])
 
-            # Initialize preceding Conv2d() bias (https://arxiv.org/pdf/1708.02002.pdf section 3.3)
             try:
                 j = layers[yolo_index] if 'from' in mdef else -1
                 bias_ = module_list[j][0].bias  # shape(255,)
                 bias = bias_[:modules.no * modules.na].view(modules.na, -1)  # shape(3,85)
-                bias[:, 4] += -4.5  # obj
-                bias[:, 5:] += math.log(0.6 / (modules.nc - 0.99))  # cls (sigmoid(p) = 1/nc)
+                with torch.no_grad():  # avoids "requires grad is being used in an in-place operation"
```
wconstab, line 110: i think this is fine, but since it could affect perf i'd not want to merge this part to the 0.1 branch

wconstab (approved):
overall lgtm. made a couple of comments. 

I don't typically run the benchmarks the way you were running them; pytest handles swallowing the output for you.

In general, I wonder if it's better to remove the prints or to switch the runner infra to run in a subprocess that doesn't spew the output, etc.  If something like that would be helpful, let me know.  Otherwise, more prints will just creep back in..

torchbenchmark/models/yolov3/yolo_models.py
jansel: This code is part of the setup so shouldn't effect perf

torchbenchmark/models/maskrcnn_benchmark/maskrcnn_benchmark/data/build.py
jansel: yeah, I can just delete it

jansel:
print() is often super slow because it involves syscalls and pipes, so IMO benchmarks shouldn't print -- especially inside the timed part of the code.

wconstab:
@jansel i'll let you press the button in case you were making any more changes but it looks good to me.

Issue #125, opened 2020-09-23
Add DeOldify Model


wconstab:
Follow instructions at torchbenchmark/models/ADDING_MODELS.md.

Model source:
https://github.com/bwasti/DeOldify

This model has already been forked and modified as part of an earlier hackathon, so while it needs work to get into the current benchmark format,  the install/data/train steps should already be clear.  Script/tracing has already been attempted without success, so that can be skipped.

PR #194, opened 2021-01-26
Add model DeOldify without jit version (#125)

huiguoo:
Issue #125 :

Added `install.py`, `__init__.py` and test data.

Note: the model is not scripted for `jit` version.


huiguoo:
@wconstab I am able to test the added model using `pytorch/benchmark/test.py` but got an error when running `pytest pytorch/benchmark/test_bench.py --collect-only`. Did I skip any configurations for `pytest`?  

Error msg:
```
ERROR test_bench.py::TestBenchNetwork - OSError: libtorch_cpu.so: cannot open shared object file: No such file or directory
```

huiguoo:
Guess I have found the reason for the error -  I downgraded my pytorch from 1.7.0 to 1.4.0 for `DeOldify` and the benchmarks were built against pytorch 1.7.0. Will try to debug the pytorch version for `DeOldify`.

wconstab:
ahh - yea, the best way to do multiple pytorch versions is 
 - create separate conda env for each pytorch version
 - install the benchmark suite into the conda env after installing pytorch

ping me if you're getting stuck.

PR #196, opened 2021-01-26
fix master

wconstab:
This benchmark has been causing headaches for some users, and now appears broken with latest torch.  Additionally, we intend to replace it with the maskrcnn implementation in torchvision.


xuzhao9 (approved):
LGTM.

PR #198, opened 2021-01-26
Add pstate CPU frequency check and setup.

xuzhao9:
Amazon g4dn.metal machine allow users to scale CPU frequency via Intel pstate driver.
To get stable TorchBench score, we need to pin CPU frequency to a fixed number.


wconstab (approved):
i think we can add 'set_' function to the 'configure' command, but that can be another PR.

PR #201, opened 2021-01-27
Remove environment checking in branch 0.1.

xuzhao9:
The environment checking should be done in the machine_util.py. Therefore, we don't need to do it in the CI script.

### DIFF
--- a/.github/scripts/run-docker.sh
+++ b/.github/scripts/run-docker.sh
@@ -11,7 +11,6 @@ CONFIG_DIR=${PWD}/score/configs/${CONFIG_VER}
 CONFIG_ENV=${CONFIG_DIR}/config-${CONFIG_VER}.env
 # Use the latest pytorch-benchmark image
 TORCH_IMAGE_ID=torchbench/pytorch-benchmark:latest
-NO_TURBO_FILE="/sys/devices/system/cpu/intel_pstate/no_turbo"
 if [ -z "$2" ]; then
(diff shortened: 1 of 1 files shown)


.github/scripts/run-docker.sh
```diff
@@ -11,7 +11,6 @@ CONFIG_DIR=${PWD}/score/configs/${CONFIG_VER}
 CONFIG_ENV=${CONFIG_DIR}/config-${CONFIG_VER}.env
 # Use the latest pytorch-benchmark image
 TORCH_IMAGE_ID=torchbench/pytorch-benchmark:latest
-NO_TURBO_FILE="/sys/devices/system/cpu/intel_pstate/no_turbo"
```
wconstab, line 14: i'm confused- no-turbo setting here seemed to work before; why did it start having a permission problem now that we moved to metal?

if anything i'd expect the machine_util would have problems with permissions since it runs _inside_ docker?  can you explain better in the git commit message?

wconstab (approved):

.github/scripts/run-docker.sh
xuzhao9: It didn't have permission problem before because on x5large this file doesn't exist.
Inside docker, I have tested that machine_util.py can *read* all the system values, but it will not be able to write into system files (Read-only file system). Therefore, the check_machine_configured() function still works.

There is another problem with the Docker environment: Docker provides a ('Ubuntu', '18.04', 'bionic') distribution (which PyTorch official docker uses), not Amazon Linux.

PR #203, opened 2021-01-27
Remove shortened names

wconstab:
- shortened names were intened for pretty-print in console output of
pytest, but they impact the names stored in json and that isn't ideal


xuzhao9 (approved):
LGTM

PR #204, opened 2021-01-27
Clean up UI for pytest machine config

wconstab:
- when exiting due to assertion, make the message more focused by
  avoiding pytests INTERNALERROR wrapper and whole stacktrace
- also add suggestion of available --ignore_machine_config override


xuzhao9 (approved):
LGTM!

PR #205, opened 2021-01-27
Update config v0 on master branch to reflect the GH runner change

xuzhao9:


torchbenchmark/score/configs/v0/config-v0.yaml
```diff
@@ -1,209 +1,209 @@
...
   test_train[pytorch_struct-cuda-jit]:
-    norm: 0.21210659145144745
+    norm: 0.24725276905005558
     weight: 0.008333333333333333
   test_train[yolov3-cuda-eager]:
-    norm: 9.4928129260079
+    norm: 9.692653778850104
     weight: 0.02222222222222222
-target: 1000.0
+target: 1000
```
wconstab, line 209: is the change to int type here going to matter? or it's fine bc of how we use it later?

.github/scripts/run-docker.sh
```diff
@@ -1,17 +1,16 @@
...
-set -euo pipefail
+set -eo pipefail
 
 # Version of the config
 CONFIG_VER=v0
 
 # Run Parameters
 RUN_SCRIPT=$1
-CONFIG_DIR=/workspace/benchmark/score/configs/${CONFIG_VER}
+CONFIG_DIR=${PWD}/score/configs/${CONFIG_VER}
```
wconstab, line 10: is this path wrong now? i moved score into torchbenchmark/score/...

wconstab (changes requested):
fix score path

torchbenchmark/score/configs/v0/config-v0.yaml
xuzhao9: Not sure why generate_score_config.py gives different types of result. Fixed

.github/scripts/run-docker.sh
xuzhao9: Right. Thanks for pointing out. Updated.

wconstab (approved):
in general not sure why you wanted to merge this into my branch instead of directly into master? but i'm ok with it, don't want you to redo it..

PR #206, opened 2021-01-27
Fix shortened model names on 0.1 branch to full names

xuzhao9:


score/torchbench_0.0.yaml
```diff
@@ -29,16 +29,16 @@ benchmarks:
   test_eval[Super_SloMo-cuda-jit]:
     norm: 0.5758905388000131
     weight: 0.006944444444444444
-  test_eval[attention_is_all_you_nee...-cpu-eager]:
```
wconstab, line 32: just noticing.. why do we have both this file and score/config_v0?  is this file useless? lets delete it then

wconstab: maybe scripts/run_bench_and_upload.sh is using this still? if so i guess it's worth pointing it to the real file instead?

wconstab (approved):
lgtm

score/torchbench_0.0.yaml
xuzhao9: Yes, the main purpose to keep this file is because run_bench_and_upload.sh is using it. Do you mean by removing this file and updating run_bench_and_upload.sh so that it uses score/configs/v0/config-v0.yaml?

wconstab: yea, that's what i meant.

we can make this change later. we'll have to revisit this anyway when we start thinking about v1 and whether there is a separate score that include everything on master and how to organize all those score files.  Also Nikitha is working on changing the score files.

PR #208, opened 2021-01-28
Updated __init__.py with domain and task class attributes

nikithamalgifb:
Fixes issue: #209
All the models have been updated with domain and class attributes. This change is in correlation with the task of eliminating score.yml from the torchbenchmark framework.


wconstab (commented):
What do you think about making an enum class in one place where all the domains and tasks are listed, and then using the enum values here rather than just using strings?  

with the string approach, i'm just worried about new models being added and people not knowing what the existing categories are, maybe making new ones up when using existing ones would have been better...  

I think it's fine if people add new categories later, (it would impact the score but it would be for a new score version anyway) but it'd be good to have some friction to adding new ones without good reason.  WDYT?

nikithamalgifb:
> What do you think about making an enum class in one place where all the domains and tasks are listed, and then using the enum values here rather than just using strings?
> 
> with the string approach, i'm just worried about new models being added and people not knowing what the existing categories are, maybe making new ones up when using existing ones would have been better...
> 
> I think it's fine if people add new categories later, (it would impact the score but it would be for a new score version anyway) but it'd be good to have some friction to adding new ones without good reason. WDYT?

This makes total sense, let me change.

torchbenchmark/__init__.py
```diff
@@ -14,6 +15,29 @@
...
+    RECOMMENDATION = "recommendation"
+    REINFORCEMENT_LEARNING = "reinforcement learning"
+    OTHER = "other"
+
+# Enum class to hold all the available task names
+class Task(Enum):
+    SEGMENTATION = "segmentation"
+    CLASSIFICATION = "classification"
+    DETECTION = "detection"
+    BACKGROUND_MATTING = "Background_Matting"
```
wconstab, line 33: better to be consistent (in domain you used spaces and lowercase) - i think spaces and lowercase is better, it's also easy for someone to add underscores later programmatically if they want them)

torchbenchmark/models/BERT_pytorch/__init__.py
```diff
@@ -18,6 +19,8 @@
 from pathlib import Path
 
 class Model:
+    domain = Domain.NLP
+    task = Task.TRANSLATION
```
wconstab, line 23: BERT should be a language modeling task, right? transformers (and RNN/LSTM) could be used for different tasks but i'm pretty sure this BERT is configured for LM.

torchbenchmark/__init__.py
```diff
@@ -14,6 +15,31 @@
...
+    COMPUTER_VISION = "computer vision"
+    OTHER_COMPUTER_VISION = "other computer vision"
+    NLP = "natural language processing"
+    SPEECH = "speech"
+    RECOMMENDATION = "recommendation"
+    REINFORCEMENT_LEARNING = "reinforcement learning"
+    OTHER = "other"
+
+# Enum class to hold all the available task names
+class Task(Enum):
```
wconstab, line 29: i'm wondering if there is a way to organize these hierarchically so that the relationship of Domain->Task is implied by the structure.

It works OK as is, but it's a little annoying that someone _could_ mix NLP domain with OTHER_RL task, for example.

What if you just flattened it - e.g.

class COMPUTER_VISION(Enum):
   SEGMENTATION = ...
   
class NLP(Enum):
   OTHER_NLP = ...
   
this way, you'd only have to set one value per model and it'd capture both parts.

you could also move these enums into a new file torchbenchmark/tasks.py 
so that in each model file you'd be able to 
from torchbenchmark.tasks import COMPUTER_VISION
and then 
self.task = COMPUTER_VISION.SEGMENTATION

torchbenchmark/__init__.py
```diff
@@ -14,6 +15,31 @@
 model_dir = 'models'
 install_file = 'install.py'
 
+# Enum class to hold all the available domain names
+class Domain(Enum):
+    COMPUTER_VISION = "computer vision"
+    OTHER_COMPUTER_VISION = "other computer vision"
```
wconstab, line 21: OTHER_COMPUTER_VISION should be a task not a domain

wconstab:
also, i only counted 20 models with labels added, but we have 25 models.  I'm not sure if all 5 of the missing models are ones we are disabling for now, but even if they are disabled shouldn't we label them and use another (more explicit) way to exclude them from the score than just leaving them unlabled?

nikithamalgifb (commented):
> also, i only counted 20 models with labels added, but we have 25 models. I'm not sure if all 5 of the missing models are ones we are disabling for now, but even if they are disabled shouldn't we label them and use another (more explicit) way to exclude them from the score than just leaving them unlabled?

Yes those are the missing models and I agree we should label them. I was not sure about the correct Domain and Task to be labelled with for these models and hence didn't not include it here. I see two options to handle this:
1.) Define Domain & Task to be NOT_DEFINED. `domain = NOT_DEFINED & task = NOT_DEFINED` for now and update it later when we have the data ready.
2.) When I have the data ready, open a new PR to enable these models directly. 

use another (more explicit) way to exclude them from the score than just leaving them unlabled?
 >> In my other PR #212, the models are filtered based on the required list when generating the default spec. 

Wdyt?

torchbenchmark/__init__.py
nikithamalgifb: Ok, the score.yml defines this as Domain:

```other computer vision:
        Background_Matting:
 ```

File: torchbenchmark/models/BERT_pytorch/__init__.py (hunk 2 above)
Comment (nikithamalgifb) [reply torchbenchmark/models/BERT_pytorch/__init__.py:23]: OK, yeah agreed I'll fix this.

File: torchbenchmark/__init__.py (hunk 3 above)
Comment (nikithamalgifb) [reply torchbenchmark/__init__.py:29]: Yea, I think this makes sense. I just did such a mistake as well. So this is a better design. I'll fix this

File: torchbenchmark/__init__.py (hunk 4 above)
Comment (wconstab) [reply torchbenchmark/__init__.py:21]: https://github.com/pytorch/benchmark/blob/master/torchbenchmark/score/score.yml#L14
i think it is defined as a task there, but since it comes after a commented out block, I wonder if yaml parser did not read it correctly? that is odd..

File: torchbenchmark/__init__.py [hunk 5]
```diff
@@ -14,6 +15,29 @@
 model_dir = 'models'
 install_file = 'install.py'
 
+# Enum class to hold all the available domain names
```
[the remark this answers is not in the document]
wconstab: when i suggested to move to a new file, it's not because this is 'wrong' but just to make the import cleaner.  Moving it only helps if you also make the other changes I suggested, since if you made my changes and left them here you'd have 
from torchbenchmark import ComputerVision
which is a little less clean than
from torchbenchmark.domains import ComputerVision

and the .domains comes from putting them in a file domains.py inside torchbenchmark.

wconstab:
I would rather not introduce an enum value for 'NOT_DEFINED' since it's essentially just a short-term hack.  

If you prefer to land this PR first and then implement the part that separates an explicit list of models and adds the missing categories later, thats OK (but I'd want you to commit to following up on that soon); but I would also be happy to help with labeling the tasks on the missing 5 now.

SuperSloMo: CV--OtherCV
Tacotron: Speech--Synthesis
which other ones are disabled?

torchbenchmark/models/alexnet/__init__.py
```diff
@@ -3,8 +3,11 @@
 import torch
 import torch.optim as optim
 import torchvision.models as models
+from torchbenchmark import Domain, Task
 
 class Model:
+    domain = Domain.COMPUTER_VISION
+    task = Task.CLASSIFICATION
```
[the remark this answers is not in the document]
wconstab: not sure i understand, is there a question here?

nikithamalgifb:
> I would rather not introduce an enum value for 'NOT_DEFINED' since it's essentially just a short-term hack.
> 
> If you prefer to land this PR first and then implement the part that separates an explicit list of models and adds the missing categories later, thats OK (but I'd want you to commit to following up on that soon); but I would also be happy to help with labeling the tasks on the missing 5 now.
> 
> SuperSloMo: CV--OtherCV
> Tacotron: Speech--Synthesis
> which other ones are disabled?

Models currently not labelled:
- maml - Domain: Reinforcement Learning? Task : Classification?
- vgg16 - Domain: Computer Vision Task: classification/Detection?
- squeezenet - Domain: Computer Vision Task: Classification?

Can you confirm if these are right? I'll just add them in this PR

wconstab (approved):
this looks great to me. thanks for iterating on it and getting here.

Issue #209, opened 2021-01-28
Remove  score.yml


nikithamalgifb:
The current torchbenchmark framework uses score.yml as the spec. Instead of maintaining a static file, we can generate the spec and thus eliminate the usage of score.yml. Below are the set of sub - tasks needed to accomplish this:

- [ ] Update the Benchmark Model class , to have domain and task class attributes
- [ ] Remove score.yml and generate the spec by reading the data from the Model class

PR #212, opened 2021-01-28
Remove score.yml and generate default spec

nikithamalgifb:
Fixes issue: #209

This PR depends on #208

Summary:
========
- Creates an explicit list of enabled (required) models which are used for the score and default spec
- Removes score.yml and uses a helper function to generate the default spec

### DIFF
--- a/torchbenchmark/score/compute_score.py
+++ b/torchbenchmark/score/compute_score.py
@@ -9,10 +9,42 @@ import sys
 import os
 import yaml
 
+from collections import defaultdict
 from tabulate import tabulate
+from torchbenchmark import list_models
 
-SPEC_FILE_DEFAULT = "score.yml"
-TARGET_SCORE_DEFAULT = 1000
+def generate_spec():
+    """
+    Helper function which constructs the spec dictionary by iterating
+    over the existing models. This API is used for generating the
+    default spec hierarchy configuration.
+    return type:
+        Returns a dictionary of type collections.defaultdict. Defaultdict
+        handles `None` or missing values and hence, no explicit `None` check
+        is required.
+    Arguments: `None`
+    Note: Only those models available inside the required_models list are
+          used to construct the default spec hierarchy.
+    """
+    spec = {'hierarchy':{'model':defaultdict(dict)}}
+    # These are the models required to generate the default spec.
+    # Please update this list if any new model needs to be a part of
+    # the default spec configuration. Before you update, please consult
+    # with someone in the benchmark team.
+    required_models = ['pytorch_mobilenet_v3', 'yolov3', 'attention_is_all_you_need_pytorch', \
+                       'BERT_pytorch', 'fastNLP', 'dlrm', 'LearningToPaint', 'moco', 'demucs', \
+                       'pytorch_struct']
+
+    for model in list_models():
+        if model.name in required_models and hasattr(model, 'domain'):
+            if not spec['hierarchy']['model'][model.domain]:
+                spec['hierarchy']['model'][model.domain] = defaultdict(dict)
+            if not spec['hierarchy']['model'][model.domain][model.task]:
+                spec['hierarchy']['model'][model.domain][model.task] = defaultdict(dict)
+            spec['hierarchy']['model'][model.domain][model.task][model.name] = None
+
+    assert len(spec) == 0, f"Spec is empty. Make sure to use models only in the required list."
+    return spec
 
 def compute_score(config, data, fake_data=None):
     target = config['target']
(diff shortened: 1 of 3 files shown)


torchbenchmark/score/compute_score.py
```diff
@@ -9,10 +9,42 @@
...
+    Arguments: `None`
+    Note: Only those models available inside the required_models list are
+          used to construct the default spec hierarchy.
+    """
+    spec = {'hierarchy':{'model':defaultdict(dict)}}
+    # These are the models required to generate the default spec.
+    # Please update this list if any new model needs to be a part of
+    # the default spec configuration. Before you update, please consult
+    # with someone in the benchmark team.
+    required_models = ['pytorch_mobilenet_v3', 'yolov3', 'attention_is_all_you_need_pytorch', \
```
wconstab, line 34: i realize now, that this way of listing just the models is less 'secure' than the previous way of listing each fully-qualified test name.  

for instance, if on monday 'yolov3' runs in cpu and gpu mode, but on tuesday someone disables it on cpu for some reason, we'll compute a subtly different weighting of the score without realizing it.

What if we take the 'required_tests' directly from the normalization file?

In the case of a 'plot sweep' the normalization file could be anything, but I think it's OK/expected that it is flexible since developers on master can be using the plot on different sets of models.

For CI, where we want to be using a locked down version, it's critical anyway that the 'golden file' be used for reference, and since that file has to be frozen in the repo anyway, we can just trust it as the source of truth for which models to expect.  What do you think?

torchbenchmark/score/compute_score.py
```diff
@@ -9,10 +9,42 @@
 import os
 import yaml
 
+from collections import defaultdict
 from tabulate import tabulate
+from torchbenchmark import list_models
 
-SPEC_FILE_DEFAULT = "score.yml"
-TARGET_SCORE_DEFAULT = 1000
+def generate_spec():
```
wconstab, line 16: i'm kind of wondering if adding this helper makes things any cleaner.  I guess at least the spec file doesn't have to exist... but i think we can still do better.

Given that you are essentially creating an empty data structure here based on the models that exist and are expected, could you do that just as easily inside generate_bench_config?

torchbenchmark/score/compute_score.py
nikithamalgifb: Please find my comments here:

> i realize now, that this way of listing just the models is less 'secure' than the previous way of listing each fully-qualified test name.
> 
> for instance, if on monday 'yolov3' runs in cpu and gpu mode, but on tuesday someone disables it on cpu for some reason, we'll compute a subtly different weighting of the score without realizing it.

> This PR will only replace what score.yml did. So `generate_spec` will only generate the default spec for only those models that were enabled as part of `score.yml`. So incase the user doesn't specify any `spec.yml` , this function will generate the default spec used for calculating the score. May be the name `generate_spec` is misleading. Renaming it to `generate_default_spec` is more appropriate. This will not alter any score calculation yet. 

> What if we take the 'required_tests' directly from the normalization file?
>  Does normalization file here refer to the user provided score.yml? If that's the case, then the required tests from the normalization file. 

> In the case of a 'plot sweep' the normalization file could be anything, but I think it's OK/expected that it is flexible since developers on master can be using the plot on different sets of models.
> 
> For CI, where we want to be using a locked down version, it's critical anyway that the 'golden file' be used for reference, and since that file has to be frozen in the repo anyway, we can just trust it as the source of truth for which models to expect. What do you think?
> This default spec will be the 'golden file' - the default locked down version. 

Does this make sense?

torchbenchmark/score/compute_score.py
nikithamalgifb: In short we can. But I am kind of confused. Even if we move , that would mean we either bake this code inside `generate_bench_config` or add a call to this helper inside `generate_bench_config`.
```
if spec is None:
   spec = generate_spec()
```

wconstab: i'm looking for a bigger cleanup than just moving the helper.  

At at a first order, I was thinking that maybe there is no point in creating the dictionary, only to then use it as a way to drive the nested for loops starting at L61 here:
https://github.com/pytorch/benchmark/blob/master/torchbenchmark/score/generate_score_config.py#L61

So, you could just try to clean up how generate_score_config works.

But.. thinking even bigger, do we need the 'generate_score_config' function at all?  or is there a cleaner way to do the whole thing?  

All the user really wants to do is run 'compute_score', but we need it to be accurate and have the safety checks somehow.

PR #226, opened 2021-01-31
set eval and train modes before benchmarking a model

Krovatkin:


test_bench.py
```diff
@@ -73,12 +73,16 @@ class TestBenchNetwork:
     """
     def test_train(self, hub_model, benchmark):
         try:
+            (model, inputs) = hub_model.get_module()
```
wconstab, line 76: my only concern with this approach is if we know that _every_ model uses the thing that is returned from .get_module() as the pytorch model object that's used for training. 

iirc there was at least one case where the train function and eval function actually construct different model objects, and bc of this i think the thing that get_module returns could be not used for training in some instances.  have you checked this?

wconstab: as for alternatives, 

1) putting model .train() or .eval() call inside of the benchmark_model.train() and .eval() functions would annoyingly include it in the time measurement, which you probably wanted to avoid...

2) we could construct a .train_model and .eval_model in __init__ for each benchmark,  and ensure they are pre-configured correctly.  but this would use more memory,  might break some models on gpu..

3) we could add an API to the benchmark_model wrapper, set_mode(train/eval) which lets each model choose whether to simply call .eval()/.train() on a shared model, whether to maintain 2 copies of the model, etc.

Krovatkin: > get_module returns could be not used for training in some instances. have you checked this?

I checked all vision models and attention, but I will check the rest.

> 1. 

Exactly, that's why I wanted to keep it outside `benchmark` calls.

> 2.

Yes, there could memory issues.

> 3. 

I like this one, I think, if I understand it correctly :-) I will have to go through all the models anyways, might as well add `set_mode` to all of them. I didn't quite get the `benchmark_model` wrapper point (since I couldn't find the base extension point) I'll ping soon to clarify

torchbenchmark/models/BERT_pytorch/__init__.py
```diff
@@ -58,6 +58,16 @@ def __init__(self, device=None, jit=False):
...
 
+    def set_eval(self):
+        self.set_mode(False)
+
+    def set_train(self):
+        self.set_mode(True)
+
+    def set_mode(self, train):
+        (model, _) = self.get_module()
+        model.train(train)
```
wconstab, line 69: wait, i'm confused why you added these API overrides for (every?) model?  i thought you would only implement this logic once in the base class, and also special logic in any particular model that doesn't work.  You seem to violate your own suggestion that it is not needed to override in ADDING_MODELS.md?

wconstab: oh...i just realized that we don't actually have a base class for the 'Model' each model implements.  i'm not sure if there is a good reason for this.. i thought we did at one point?

torchbenchmark/models/maml/__init__.py
```diff
@@ -53,20 +53,37 @@ def __init__(self, device='cpu', jit=False):
...
+        self.set_mode(False)
+
+    def set_train(self):
+        self.set_mode(True)
+
+    def set_mode(self, train):
+        (model, _) = self.get_module()
+        model.train(train)
+
+    def set_eval(self):
```
wconstab, line 68: duplicated?

torchbenchmark/models/yolov3/__init__.py
```diff
@@ -52,9 +52,24 @@ def get_module(self):
...
+
+    def set_eval(self):
+        self.set_mode(False)
+
+    def set_train(self):
+        self.set_mode(True)
+
+    def set_mode(self, train):
+        if not train:
+            (model, _) = self.get_module()
```
wconstab, line 67: in this case it would be fine to just no-op set_mode and rely on the existing line55 model.to(opt.device).eval() right?

torchbenchmark/models/BERT_pytorch/__init__.py
Krovatkin: I initially intended to add a base class and add this logic to it. Then when you suggested to `Model`s in `__init__.py` I assumed you didn't want the base class. I'd rather add a base class. Let me update the PR

torchbenchmark/models/yolov3/__init__.py
Krovatkin: I find it less consistent and more confusing, but I don't have a strong opinion here )

torchbenchmark/models/BERT_pytorch/__init__.py
wconstab: Oh, I'm really sorry! Sounds like we both confused each other.  I think it's clear now :)

torchbenchmark/models/yolov3/__init__.py
wconstab: i also don't have a strong opinion.

torchbenchmark/models/LearningToPaint/__init__.py
```diff
@@ -54,9 +54,17 @@ def __init__(self, device=None, jit=False):
     def get_module(self):
         return self.module,[self.example_inputs]
 
-    def train(self, niter=1):
-        self.module.train()
+    def set_eval(self):
```
wconstab, line 57: missed moving this one to BenchmarkModel?

torchbenchmark/models/pytorch_CycleGAN_and_pix2pix/__init__.py
```diff
@@ -34,6 +36,11 @@ def __init__(self, device='cpu', jit=False):
     def get_module(self):
         return self.model, self.input
         
+    def set_train(self):
```
wconstab, line 39: isn't it better to override set_mode?  if someone calls set_mode or set_eval it'd still call through to self.model.eval() i guess?  although i think it's also true that the overriding is not technically important here, since if people use the APIs correctly they will call set_eval before using the eval model, and the train model will always work right...

torchbenchmark/util/model.py
```diff
@@ -0,0 +1,22 @@
+import json
+import os
+import pandas as pd
+import typing
+from  collections.abc import Iterable
+import torch
+
+class BenchmarkModel():
+    def __init__(self, *args, **kwargs):
```
wconstab, line 9: It'd be better to finish this implementation - e.g.
- fill in the train and eval methods with raise NotImplemented
- set a docstring at least pointing to the place where we define this interface in .md file (maybe later we move the comments to inline and generate doc but i'd rather not duplicate the same doc in 2 places for now)

wconstab (approved):
made a couple small comments but lgtm. thanks a lot for fixing this up @Krovatkin!!

wconstab (approved):
lgtm!

PR #227, opened 2021-02-01
Lock spacy version

Krovatkin:
Spacy 3 deprecated short names (e.g. `en`, `de`) for trained embeddings, lock spacy version for now until attention-all-you-need is updated


wconstab (approved):
thanks for fixing this!

iiuc this is a change that preserves current behavior against a future deletion of those apis.  in that case can you also PR this change to the 0.1 branch?

xuzhao9:
Created PR for this on 0.1: https://github.com/pytorch/benchmark/pull/228

PR #228, opened 2021-02-01
Fix spacy version and fix a sweep test script bug of work directory

xuzhao9:


xuzhao9:
Closed because this is fixed by #232

PR #230, opened 2021-02-02
[WIP] set_mode

Krovatkin:


torchbenchmark/models/alexnet/__init__.py
```diff
@@ -3,16 +3,27 @@
...
         self.example_inputs = (torch.randn((32, 3, 224, 224)),)
 
+    def eval(self):
+        self.set_mode(self, False)
+    
+    def train(self):
+        self.set_mode(self, True)
+
+    def set_mode(self, train = True):
+        self.model.train(train)
```
wconstab, line 25: yes, i think this is great.  I guess it will work out of the box for most models, without changing the model.  The key part is to carefully check if any model does not work this way, and make sure to fix it;  

Also, can you update the 'Adding_models.md' so that anyone following the instructions knows to make sure they implement this if they need to.

conftest.py
```diff
@@ -9,6 +9,9 @@ def pytest_addoption(parser):
     parser.addoption("--ignore_machine_config",
                      action='store_true',
                      help="Disable checks/assertions for machine configuration for stable benchmarks")
+    parser.addoption("--check_results",
```
wconstab, line 12: the result-checker is a good idea.  I'm wondering if we should add it in pytest or in the `python test.py` runner that is based on some other test framework.  I guess it also depends if you want the results checker to happen during a run that also measures timings, and whether you want it to happen during pytest/benchmark CI runs.  

Currently on the pytest/benchmark repo we run these configs to catch bugs in the benchmark suite:
 - python test.py (mostly for checking for error, runs quickly)
 - pytest test_bench.py (to exercise the benchmark timing and json parsing and score computation, but configured to be faster and noisier)

On the pytorch/pytorch CI, we do the actual invocation of `pytest test_bench.py` with the tuned CI machine and the number of rounds needed to gather good data and log to scuba etc.  But we don't run a correctness test here currently.

PR #232, opened 2021-02-03
Add script that runs benchmark without Docker.

xuzhao9:
Disable the usage of Docker because we find unexpected noises in Docker makes score to be unstable.

### DIFF
--- a/.github/scripts/run-nightly-nodocker.sh
+++ b/.github/scripts/run-nightly-nodocker.sh
@@ -1,0 +1,78 @@
+#!/bin/sh
+
+set -euo pipefail
+
+# Sanity checks
+if [[ ! -v GITHUB_RUN_ID ]]; then
+    echo "GITHUB_RUN_ID is not set! Please check your environment variable. Stop."
+    exit 1
+fi
+
+if [[ ! -v SCRIBE_GRAPHQL_ACCESS_TOKEN ]]; then
+    echo "SCRIBE_GRAPHQL_ACCESS_TOKEN is not set! Please check your environment variable. Stop."
+    exit 1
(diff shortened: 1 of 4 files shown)


.github/scripts/run-nightly-nodocker.sh
```diff
@@ -0,0 +1,77 @@
...
+export GOMP_CPU_AFFINITY="${CORE_LIST}"
+
+mkdir -p ${DATA_DIR}
+conda create -y -q --name ${CONDA_ENV_NAME} python=${PYTHON_VERSION}
+. activate ${CONDA_ENV_NAME}
+conda init bash; conda run /bin/bash
+
+cp docker/multi30k.tar.gz torchbenchmark/models/attention_is_all_you_need_pytorch
+pushd torchbenchmark/models/attention_is_all_you_need_pytorch
+mkdir -p .data/multi30k
```
wconstab, line 39: these are some kind of caching to speed up the subsequent install.py?

xuzhao9: This is because multi30k.tar.gz is not consistently available (the download fails sometimes). To make sure CI is not interrupted by the broken URL link, I store it locally in the repository.

xuzhao9: Previously we didn't have this problem because the files was packaged together in the Docker, see: https://github.com/pytorch/benchmark/pull/178

wconstab: ok, yea this is a good catch. those download mirrors are annoying.

wconstab (approved):

PR #238, opened 2021-02-03
Make sure diagnostic variables aren't set when benchmarking

Krovatkin:


conftest.py
```diff
@@ -25,8 +25,21 @@ def set_fuser(fuser):
         torch._C._jit_override_can_fuse_on_gpu(True)
         torch._C._jit_set_texpr_fuser_enabled(True)
 
+def check_environment():
+    checks = [
+        # VAR_NAME, forbidden value
```
wconstab, line 30: shouldn't this read VAR_NAME, required value?

sounds backward to me.  IIUC you want to make sure logging is off, for instance?

conftest.py
```diff
@@ -25,8 +25,21 @@ def set_fuser(fuser):
...
+        ("PYTORCH_JIT_LOG_LEVEL", None)
+    ]
+
+    for check in checks:
+        if check[0] in os.environ and (check[1] == None or os.environ[check[0]] == check[1]):
+            raise RuntimeError(f"{check[0]} is set")
+
 def pytest_sessionstart(session):
     try:
+        check_environment()
```
wconstab, line 42: What do you think about moving this check into the util file, and calling check_environment from inside check_machine_configured?  Any reason to separate them?

wconstab (approved):

PR #242, opened 2021-02-05
Add initial workflow and scripts for the nightly instruction count benchmarking

xuzhao9:
This PR creates a starting point for the nightly CI of instruction count benchmarking.
Other than code checkout, the workflow has three steps:
1. create-nightly-env: create the conda environment "instr_cnt_benchmark" for testing. It will also install PyTorch nightly wheels via pip.
2. run-benchmark: calls the script `.github/scripts/instr-count-benchmark/run-benchmark.sh` to run the benchmark. Currently, this file is empty.
3. remove-nightly-env: delete the conda environment "instr_cnt_benchmark" as cleanup.

The `config.env` file stores the environment variables shared by this workflow, such as Python version, CUDA version, and the Conda environment name.


wconstab (approved):
lgtm, but let's see if it makes sense to @robieta

robieta (approved):
This all makes sense. Thanks for putting this together; it will definitely jump start my ability to get things wired up!

PR #244, opened 2021-02-05
Make test and install scripts handling the failure caused by maskrcnn

penguinwu:
Make test and install scripts handling the failure caused by maskrcnn benchmark more gracefully (i.e., not abort the entire test or installation process

### DIFF
--- a/torchbenchmark/__init__.py
+++ b/torchbenchmark/__init__.py
@@ -87,6 +87,9 @@ def list_models():
     for model_path in _list_model_paths():
         model_name = os.path.basename(model_path)
         module = importlib.import_module(f'.models.{model_name}', package=__name__)
+        if not hasattr(module, 'Model'):
+            print(f"Warning: {module} does not define attribute Model, skip it")
+            continue
         Model = getattr(module, 'Model')
         if not hasattr(Model, 'name'):
             Model.name = model_name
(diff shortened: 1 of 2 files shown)


torchbenchmark/__init__.py
```diff
@@ -87,6 +87,9 @@ def list_models():
     for model_path in _list_model_paths():
         model_name = os.path.basename(model_path)
         module = importlib.import_module(f'.models.{model_name}', package=__name__)
+        if not hasattr(module, 'Model'):
+            print(f"Warning: {module} does not define attribute Model, skip it")
```
wconstab, line 91: I didn't test this, but does it correctly continue the for loop if one model is not importable due to installation failure?   I wasn't sure what the behavior of importlib is  in this case-  does it return a module object that is missing the MaskRCNN model?  or it just crashes when you try import_module?

wconstab (approved):
lgtm overall. thanks for cleaning up!

torchbenchmark/__init__.py
penguinwu: `importlib` works fine, but the `module` created by it (for maskrnn) does not have attribute `Model` defined, causing a Python runtime error. That's why I put in this fix to check the `Model` attribute first.

`python test.py` runs fine in my environment.

PR #249, opened 2021-02-09
add `check_opt_vs_noopt_jit` option to check that results between baseline and optimized jitted versions match

Krovatkin:
add `check_opt_vs_noopt_jit` option to check that results between baseline and optimized jitted versions match

### DIFF
--- a/torchbenchmark/util/model.py
+++ b/torchbenchmark/util/model.py
@@ -48,7 +48,7 @@ class BenchmarkModel():
         (model, _) = self.get_module()
         model.train(train)
 
-    def check_results(self):
+    def check_opt_vs_noopt_jit(self):
         if not self.jit:
             return
 
--- a/torchbenchmark/util/model.py
+++ b/torchbenchmark/util/model.py
@@ -5,7 +5,9 @@ import typing
 from  collections.abc import Iterable
 import torch
 from contextlib import contextmanager
-
+import warnings
+import inspect
+import os
 
 @contextmanager
 def no_grad(val):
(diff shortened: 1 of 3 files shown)


torchbenchmark/util/model.py
```diff
@@ -29,4 +29,34 @@ def _set_mode(self, train):
         (model, _) = self.get_module()
         model.train(train)
 
-
+    def check_results(self):
```
wconstab, line 32: i'm wondering how this is intended to be used.   If it's used mainly locally by jit developers, it won't affect all users of the benchmark...  Should we make it slightly less intrusive by not needing to implement a 'check_results' override for models that don't jit?  is it sufficient to just try/except NotImplementedError inside this func to avoid the ones that aren't supported?

torchbenchmark/util/model.py
```diff
@@ -46,4 +47,52 @@ def _set_mode(self, train):
...
+            return
+
+        model_name = inspect.getfile(self.__class__)[-2]
+        model_blacklist = [
+            'demucs',
+            'yolov3',
+        ]
+
+        if model_name in model_blacklist:
+            warnings.warn(UserWarning(f"{model_name}.get_module() doesn't support `check_results` yet!"))
```
wconstab, line 61: is this just WIP (that you have both blacklist and try/except)? or you need both for some reason?

Krovatkin: @wconstab unfortunately, for some benchmarks results don't match, while the others are set up in a way I can get at their jitted module to be able to reset the optimization version and recompile it at noopt.
I added some comments. 
Eventually, I'm hoping to enable results checking for as many models as I can

wconstab (approved):
lgtm- other than one nit, which is that you could add a comment and/or change check_results function name, explaining the idea of verifying that the JIT optimization behaves same as baseline JIT. 

(Disambiguate from also important but non-addressed case of making sure the model is producing correct results..)

asuhan (approved):

PR #250, opened 2021-02-09
Use `parse_lscpu_cpu_core_list` to get active cpus

xuzhao9:
Also, check hyper-threading status is disabled before getting/setting pstate frequencies, and enable pstate frequency setup when running with ``--configure`` option.

### DIFF
--- a/torchbenchmark/util/machine_config.py
+++ b/torchbenchmark/util/machine_config.py
@@ -213,8 +213,7 @@ def get_pstate_frequency():
 
     if hyper_threading_enabled():
         return None
-    active_cpus = parse_lscpu_cpu_core_list()
-    cpu_dirs = ["cpu" + f for f in active_cpus]
+    cpus_dirs = ["cpu" + str(cpu[0]) for cpu in parse_lscpu_cpu_core_list() if cpu[2])
     output = dict()
     for cpu_dir in cpu_dirs:
         full_path = os.path.join(CPU_FREQ_BASE_DIR, cpu_dir, "cpufreq")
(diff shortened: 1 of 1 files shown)


torchbenchmark/util/machine_config.py
```diff
@@ -210,7 +210,10 @@ def parse_block(block):
 def get_pstate_frequency():
     CPU_FREQ_BASE_DIR = '/sys/devices/system/cpu'
     CPU_FREQ_FILES = ["scaling_min_freq", "scaling_max_freq", "scaling_cur_freq"]
-    cpu_dirs = [f for f in os.listdir(CPU_FREQ_BASE_DIR) if re.match(r'cpu[0-9]+', f)]
+
+    if hyper_threading_enabled():
+        return None
```
wconstab, line 215: is this really needed?

i think i'm confused about what interacts badly with hyperthreading and min/max freq.

it seems valid to get/set min/max freq if hyperthreading is enabled _or_ disabled.

the bug i reported was in a mixed-case where hyperthreading was disabled, but the script tried to get/set freq for disabled cpuid.

i suspect things would work ok if we 
- allow get/set min/max when ht is enabled or disabled
- make sure we disable ht before we set min/max if that is for some reason important
- inside get/set only interact with the 'online' cores

does that make sense, or you think there is a problem if you do it that way?

xuzhao9: I think the question is: does setting the frequency of offline core have any effect? From my experiment, it seems to have no effect.

Therefore, the only effect of removing this check is to set min/max freq when HT is enabled, which could be useful in some case, maybe?

wconstab: right. although i'm not sure if it is even valid to set the freq of a second virtual core.  presumably, the cpu does not change freq when it changes threads, but maybe it can, i have no idea.  i guess, this is not a major issue i just thought it's cleaner library code to have it function for both cases unless one case is truly invalid.  (i agree that i don't plan to use hyperthreading so it's not particularly useful)

xuzhao9: Based on the [kernel documentation](https://www.kernel.org/doc/html/v5.0/admin-guide/pm/intel_pstate.html), if HT is enabled, and two virtual cores on the same CPU don't have the same frequency policy, they will be inter-affected. It doesn't mention when HT is disabled. I will remove the HT disable check for sanity purpose.

wconstab (approved):
thanks! lgtm except just noticed one more thing-
can you also add get_pstate_freq into config dictionary for get_machine_config, so it's recorded in bench json?

PR #253, opened 2021-02-11
[WIP] Set up torchbenchmark for running against custom PyTorch

Krovatkin:
One of the most common workflow for me is running torchbenchmark against Pytorch with some custom changes. This script is supposed to build Pytorch from a given branch and compile vision/audio/text against it.

### DIFF
--- a/scripts/install_torchbenchmark.sh
+++ b/scripts/install_torchbenchmark.sh
@@ -1,4 +1,12 @@
 #!/bin/bash
+################################################################################
+# The purpose of this script is to automate a TorchBenchmark set-up that uses
+# a custom PyTorch built from a given branch and repository.
+# This is useful for quickly testing an impact of a particular optimization/change
+# Arguments:
+#   BRANCH_NAME
+#   REPO
+#######################################
 
 if [ "$1" = "-h" -o "$1" = "--help" ]; then
   echo "Usage: prog BRANCH_NAME REPO"
@@ -21,13 +29,13 @@ conda activate $BRANCH_NAME
 echo "Installing PyTorch dependencies"
 conda install -y numpy ninja pyyaml mkl mkl-include setuptools cmake cffi typing_extensions future six requests dataclasses 2>&1 > /dev/null
 conda install -y -c pytorch magma-cuda100 2>&1 > /dev/null
-echo "Checking out $"
+echo "Checking out $ORIG_BRANCH_NAME"
 git clone $REPO $BRANCH_NAME
-cd $BRANCH_NAME
+pushd $BRANCH_NAME
 git checkout $ORIG_BRANCH_NAME
 git submodule sync 2>&1 > /dev/null
 git submodule update --init --recursive 2>&1 > /dev/null
-echo "Builing PyTorch"
+echo "Building PyTorch"
 export USE_CUDA=1
 export BUILD_CAFFE2_OPS=0
 export USE_XNNPACK=0
@@ -44,25 +52,25 @@ test $? -eq 0 || { echo "PyTorch build failed!"; exit; }
 unset CMAKE_PREFIX_PATH
 
 git clone --recursive  https://github.com/pytorch/vision.git torchvision
-cd torchvision
+pushd torchvision
 python setup.py install 2>&1 > /dev/null
 test $? -eq 0 || { echo "torchvision build failed!"; exit; }
-cd .. 
+popd
 
 git clone --recursive https://github.com/pytorch/audio.git torchaudio
-cd torchaudio
+pushd torchaudio
 BUILD_SOX=1 python setup.py install 2>&1 > /dev/null
 test $? -eq 0 || { echo "torchaudio build failed!"; exit; }
-cd ..
+popd
 
 git clone https://github.com/pytorch/text torchtext
-cd torchtext
+pushd torchtext
 git submodule update --init --recursive
 python setup.py clean install 2>&1 > /dev/null
 test $? -eq 0 || { echo "torchtext build failed!"; exit; }
-cd ..
+popd
 
 git clone --recursive https://github.com/pytorch/benchmark.git
-cd benchmark
+pushd benchmark
 python install.py
 
--- a/scripts/install_torchbenchmark.sh
+++ b/scripts/install_torchbenchmark.sh
@@ -28,8 +28,20 @@ git checkout $ORIG_BRANCH_NAME
 git submodule sync 2>&1 > /dev/null
 git submodule update --init --recursive 2>&1 > /dev/null
 echo "Builing PyTorch"
+export USE_CUDA=1
+export BUILD_CAFFE2_OPS=0
+export USE_XNNPACK=0
+export USE_MKLDNN=1
+export USE_MKL=1
+export USE_CUDNN=1
+# if USE_LLVM isn't set check in common locations
+if ! [ -z ${USE_LLVM} ]; then
+    [ -e /usr/lib/llvm-9/ ] && export USE_LLVM=/usr/lib/llvm-9/
+fi
+export CMAKE_PREFIX_PATH=$CONDA_PREFIX
 python setup.py install 2>&1 > /dev/null
 test $? -eq 0 || { echo "PyTorch build failed!"; exit; }
+unset CMAKE_PREFIX_PATH
 
 git clone --recursive  https://github.com/pytorch/vision.git torchvision
 cd torchvision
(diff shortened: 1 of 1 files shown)


scripts/install_torchbenchmark.sh
```diff
@@ -0,0 +1,60 @@
...
+git clone $REPO $BRANCH_NAME
+cd $BRANCH_NAME
+git checkout $ORIG_BRANCH_NAME
+git submodule sync 2>&1 > /dev/null
+git submodule update --init --recursive 2>&1 > /dev/null
+echo "Builing PyTorch"
+export USE_CUDA=1
+export BUILD_CAFFE2_OPS=0
+export USE_XNNPACK=0
+export USE_MKLDNN=1
```
bertmaher, line 34: I recommend a few more build options:
Set cmake prefix to properly pick up MKL from conda:
```
CMAKE_PREFIX_PATH=$CONDA_PREFIX
```

LLVM if you want CPU fusion (this could be optional?)
```
USE_LLVM=/usr/lib/llvm-9/
```
I don't know if these are needed or if they're automatic:
```
USE_MKL=1
USE_CUDNN=1
```

bertmaher:
This looks very useful, I've definitely done these steps by hand several times now.

bertmaher (approved):

scripts/install_torchbenchmark.sh
Krovatkin: @bertmaher I wonder if I should just remove any opinionated defaults? The chances are these are already set in one's environment?

bertmaher: I kinda prefer to have them in a script like this.  I'd rather have a script do the right thing than depend on random people knowing the right environmental magic.

wconstab:
I'm ok with this from the perspective of it's useful and you guys want it.  

@xuzhao9 is maintaining somewhat overlapping scripts that do some of this in CI, so we should converge to share implementation where possible.  Xu, can you help take a look at how easy it would be to reuse some of your scripts for this, or comment on which parts are similar/different to what you do?  If possible maybe Nick's script could call one of your scripts for at least part of the process?

Also, i'm still undecided on the 'right' way to handle the torchvision/text libs for the v1 score, where we not only depend on them for slow-changing stuff like datasets, but started to directly import whole models from them.  The issue being whether to freeze torchvision at a particular commit for the lifetime of the v1 score vs let it roll along with pytorch which exposes us more to model structure changes.

scripts/install_torchbenchmark.sh
```diff
@@ -0,0 +1,56 @@
...
+echo "Installing PyTorch dependencies"
+conda install -y numpy ninja pyyaml mkl mkl-include setuptools cmake cffi typing_extensions future six requests dataclasses 2>&1 > /dev/null
+conda install -y -c pytorch magma-cuda100 2>&1 > /dev/null
+echo "Checking out $"
+git clone $REPO $BRANCH_NAME
+cd $BRANCH_NAME
+git checkout $ORIG_BRANCH_NAME
+git submodule sync 2>&1 > /dev/null
+git submodule update --init --recursive 2>&1 > /dev/null
+echo "Builing PyTorch"
```
xuzhao9, line 30: Nit: Building PyTorch

scripts/install_torchbenchmark.sh
```diff
@@ -0,0 +1,56 @@
+#!/bin/bash
+
```
xuzhao9, line 2: Can you add a doc string to explain the purpose of this script?

scripts/install_torchbenchmark.sh
```diff
@@ -0,0 +1,56 @@
...
+cd $BRANCH_NAME
+git checkout $ORIG_BRANCH_NAME
+git submodule sync 2>&1 > /dev/null
+git submodule update --init --recursive 2>&1 > /dev/null
+echo "Builing PyTorch"
+python setup.py install 2>&1 > /dev/null
+test $? -eq 0 || { echo "PyTorch build failed!"; exit; }
+
+git clone --recursive  https://github.com/pytorch/vision.git torchvision
+cd torchvision
```
xuzhao9, line 47: `pushd`/`popd` is more script-friendly than `cd`/`cd ..`

scripts/install_torchbenchmark.sh
Krovatkin: added them back and added optional `USE_LLVM`

scripts/install_torchbenchmark.sh
```diff
@@ -0,0 +1,68 @@
...
+REPO=$2
+BRANCH_NAME=$(python -c 'import sys; print (sys.argv[1].replace("/","_"))' $ORIG_BRANCH_NAME)
+echo "Your branch will be built into $BRANCH_NAME and torchbenchmark will be installed in ${BRANCH_NAME}/benchmark"
+conda create -y -n $BRANCH_NAME python=3.7 2>&1 > /dev/null
+test $? -eq 0 || { echo "failed to create env ${BRANCH_NAME}!"; exit; }
+conda activate $BRANCH_NAME
+echo "Installing PyTorch dependencies"
+conda install -y numpy ninja pyyaml mkl mkl-include setuptools cmake cffi typing_extensions future six requests dataclasses 2>&1 > /dev/null
+conda install -y -c pytorch magma-cuda100 2>&1 > /dev/null
+echo "Checking out $"
```
xuzhao9, line 24: Should it be `echo "Checking out $BRANCH_NAME"`?

Krovatkin:
@wconstab 

> we should converge to share implementation where possible

We should definitely try sharing as much as possible. Unfortunately, after quickly scanning through the existing scripts, I got an impression that most of them download pre-built packages.

> Also, i'm still undecided on the 'right' way to handle the torchvision/text libs for the v1 score

 I was thinking that, for the purpose of testing the impact of a particular change, it hopefully shouldn't matter much as long as the same version of torchvision/audio/text is used to test w/ and w/o the change?

scripts/install_torchbenchmark.sh
```diff
@@ -0,0 +1,68 @@
...
+. ${CONDA_EXE%bin/conda}/etc/profile.d/conda.sh
+ORIG_BRANCH_NAME=$1
+REPO=$2
+BRANCH_NAME=$(python -c 'import sys; print (sys.argv[1].replace("/","_"))' $ORIG_BRANCH_NAME)
+echo "Your branch will be built into $BRANCH_NAME and torchbenchmark will be installed in ${BRANCH_NAME}/benchmark"
+conda create -y -n $BRANCH_NAME python=3.7 2>&1 > /dev/null
+test $? -eq 0 || { echo "failed to create env ${BRANCH_NAME}!"; exit; }
+conda activate $BRANCH_NAME
+echo "Installing PyTorch dependencies"
+conda install -y numpy ninja pyyaml mkl mkl-include setuptools cmake cffi typing_extensions future six requests dataclasses 2>&1 > /dev/null
```
xuzhao9, line 22: Maybe use `conda install --only-deps pytorch` to install only dependencies?

xuzhao9 (approved):
Overall looks good, added a few inline comments.

This script focuses more on building and installing customized PyTorch, while the CI scripts are more about running the benchmark, so they are not really overlapping.

scripts/install_torchbenchmark.sh
Krovatkin: @xuzhao9  let me try seeing if that works. I think for building there are more dependencies than listed in a conda package

Krovatkin: @xuzhao9 yup, it breaks :-( 
```
Building PyTorch
  File "setup.py", line 368, in check_pydep
    importlib.import_module(importname)
  File "/home/villedepommes/miniconda3/envs/krovatkin_shapes_grad/lib/python3.7/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1006, in _gcd_import
  File "<frozen importlib._bootstrap>", line 983, in _find_and_load
  File "<frozen importlib._bootstrap>", line 965, in _find_and_load_unlocked
ModuleNotFoundError: No module named 'yaml'
```

xuzhao9: Ahh right, this is because `yaml` is a build-time dependency, not runtime dependency. In this case, we have to install the build time deps manually.

wconstab:
ok sounds like you're clear to land, and i'd just ask if @xuzhao9 is building torch as part of bisection or other reasons, we should try to keep from having duplicated scripts.

PR #254, opened 2021-02-12
Refactor compute_score to pre-calculate static weights and calculates score per config

nikithamalgifb:
Fixes issue: #209

Summary:
=======

- This PR calculates score per config for every benchmark
- Adds support to pre-compute static weights based on the spec
- Eliminates the use of generate_spec_cfg.


torchbenchmark/score/compute_score.py
```diff
@@ -8,31 +8,147 @@
...
+    return benchmark_norms
+
+def get_weights_from_spec(spec):
+    """
+    Calculates the static benchmark weights by iterating the spec
+    file and constructs a b_weight dictionary with (key, value) pair
+    is (task, weight_for_benchmark_per_task)
+
+    The spec here can be user defined .yaml file or if no spec is defined,
+    the default score.yaml is used
```
wconstab, line 38: It sounds like you don't have to pass in spec, and then score.yml is used, but it looks like you have to pass in spec.

I think spec should be an optional input.

This function should probably not know about the yaml format.  Instead, it should accept the simplest python data structure it needs to do its job, which may be a dictionary with a certain structure.  The stuff about yaml files or json files or whatever can be left outside in the function that calls this.

I'm not sure what the default should be if no spec is input.  It could either come from reading a file or it could come from looking at the enums.  I lean toward the enums since we could then delete a file and not have to duplicate content.  What do you think?

torchbenchmark/score/compute_score.py
```diff
@@ -8,31 +8,147 @@
...
+        test = "eval"
+    if "cpu" in b:
+        device = "cpu"
+    if "jit" in b:
+        mode = "jit"
+
+    # Extract the model name from the benchmark b
+    pos1 = b.index("[")
+    pos2 = b.index("-"+device)
+    model_name = b[pos1+1:pos2]
```
wconstab, line 76: this section seems like it'd be cleaner to just use a regex.  you could do it all in one line (i think i wrote one before for benchmarks but i forgot where it is)

torchbenchmark/score/compute_score.py
```diff
@@ -8,31 +8,147 @@
...
+
+    # Get the Model task value by reading the Model attributes.
+    p = Path(__file__).parent.joinpath('../../models/'+ model_name)
```
wconstab, line 79: do we record the task into the benchmark json? if not we probably should, so we can just read it form the data instead of having to reconstruct it this way.  

also, if we are going to do this, it'd probably be better to make this lookup a helper function so we could possibly improve how it's implemented later.  something like get_model_task(model_name)

torchbenchmark/score/compute_score.py
```diff
@@ -8,31 +8,147 @@
...
+def construct_benchmark_run_db(data):
+    """
+    Construct a benchmark database by going over through the data file
+    for the run and update the dictionary by task and model_name
+
+    For eg., the (key, value) for this dictionary is of the form
+    [generation][pytorch_stargan] = [(1.2, (eval, cuda, jit),
+                                    test_eval[pytorch_stargan-cuda-jit])]
+    """
+    found_benchmarks = defaultdict(dict)
```
wconstab, line 94: i forgot, can you do defaultdict(defaultdict(list))?  then you can delete the if not found_benchmarks below maybe?

torchbenchmark/score/compute_score.py
```diff
@@ -8,31 +8,147 @@
...
+        # Append the tuple(mean, config, model_name) for all the configs the
+        # benchmark was run with.
+        found_benchmarks[task][model_name].append((mean, config, name))
+    return found_benchmarks
+
+def get_score_per_config(data, b_weight, norm):
+    """
+    This function iterates over found benchmark dictionary
+    and calculates the weight_sum and benchmark_score.
+    A score_db is then constructed to calculate the cummulative
```
wconstab, line 110: cumulative

torchbenchmark/score/compute_score.py
```diff
@@ -8,31 +8,147 @@
...
+    # score_db: score_per_config for now is just calculated.
+    # This needs to passed to plot_sweep.py to plot all the
+    # scores per config.
+    """
+    score = 0.0
+    score_db = get_score_per_config(data, weight_db, norm)
+
+    for config, scores in score_db.items():
+        score += scores
+        score_db[config] = target * math.exp(scores * 0.125)
```
wconstab, line 151: this is confusing me, you're modifying score_db as you iterate through it?  but then you're not using score_db again, and you're returning score?

compute_score.py
```diff
@@ -25,22 +24,27 @@
...
     if args.benchmark_data_file is not None:
+        norm = None
         with open(args.benchmark_data_file) as data_file:
             data = json.load(data_file)
-            if config is None:
-                config = generate_bench_cfg(spec, data, TARGET_SCORE_DEFAULT)
-        score = compute_score(config, data)
+            if norm is None:
+                norm = get_benchmark_norms(data)
+        score = compute_score(weight_db, data, TARGET_SCORE_DEFAULT, norm)
```
wconstab, line 47: i'm thinking this function has too complicated an API.  Maybe the way to help it is to make a new class on which compute_score is a method.

I'm thinking instead of create these dictionaries and give them confusing names like 'weight_db' or 'spec' or 'config' (not your fault, it's just not possible to come up with a name that explains what the thing does), what if we do this

score_config = TorchBenchScore(args to configure it)
score_config.compute_score(data)

and in the default case, when no spec file is provided, 
from somewhere import TorchBenchmarkScore_v0 //it'd be pre-constructed in 'somewhere'
TorchBenchmarkScore_v0.compute_score(data)


TARGET_SCORE_DEFAULT should be just a default arg to construct the score 

it should be possible to construct a TorchbenchScore() directly from a reference data, without a spec, i think.  
from somewhere import TorchBenchScore
score_config = TorchBenchScore.from_reference_data(ref_data)
score_config.compute_score()

I wonder if it's really important to have the weight_db or spec as an exposed thing?  If the goal is to enable filtering out models (assume the reference data has too many models, but you don't care about all of them), maybe just pass a list of excludes?
score_config = TorchBenchScore.from_reference_data(ref_data, exclude_models=None, exclude_tests=None, exclude_device=None)...
score_config.compute_score()

torchbenchmark/score/compute_score.py
```diff
@@ -7,148 +7,151 @@
...
+                benchmarks = tasks[task]
+                benchmark_weight = 1.0 / len(benchmarks)
+                self.weights[task] = domain_weight * task_weight * benchmark_weight
+
+    def setup_benchmark_norms(self, data):
+        """
+        Helper function which gets the normalization values per benchmark
+        by going through the reference data file.
+
+        This reference data can be user provided or by default the first file
```
wconstab, line 57: try to decouple the comments here from the context of the plot code where it's currently being called. e.g. no point mentioning 'the first file' here - all this comment should care about is that you provide the reference data.  (and you could rename data to ref_data to make that clearer)

torchbenchmark/score/compute_score.py
```diff
@@ -7,148 +7,151 @@
...
+        in the directory of data files is considered as the reference data file
+        """
+        self.norm = {b['name']: b['stats']['mean'] for b in data['benchmarks']}
+
+    def _get_model_task(self, model_name):
```
wconstab, line 62: this probably shouldn't be a function of the TorchBenchScore class, seems like a standalone func

torchbenchmark/score/compute_score.py
```diff
@@ -7,148 +7,151 @@
...
+        This helper also extracts the task and model_name of the benchmark
+        The function returns a tuple of the configs, model task, model_name
+        """
+        test, device, mode = "train", "cuda", "eager"
+        if "eval" in b: test = "eval"
+        if "cpu" in b: device = "cpu"
+        if "jit" in b: mode = "jit"
+
+        # Extract the model name from the benchmark b
+        model_name = (re.findall(r'\[(.*)\-'+device, b))[0]
```
wconstab, line 91: for this i was thinking you could do it all in one line and not even need this helper function.  e.g.

test, device, mode = re.parse("the regexp", b).groups()[0]

well, it would still need error handling before doing [0] but that's the only thing.  
https://regex101.com/ is good if you want to quickly figure out the regexp.  i would use - and [ characters and begin-line character as anchors and () to grab the different parts.

torchbenchmark/score/compute_score.py
```diff
@@ -7,148 +7,151 @@
...
+        Construct a benchmark database by going over through the data file
+        for the run and update the dictionary by task and model_name
+
+        For eg., the (key, value) for this dictionary is of the form
+        [generation][pytorch_stargan] = [(1.2, (eval, cuda, jit),
+                                        test_eval[pytorch_stargan-cuda-jit])]
+        """
+        found_benchmarks = defaultdict(lambda: defaultdict(list))
+
+        for b in data['benchmarks']:
```
wconstab, line 108: wondering if it makes sense to split this out as a helper function.  

it seems like it might be cleaner to just combine this loop with the loops in get_score_per_config, since there doesn't seem to be any value in having the 'db' object exist or keeping it around, and adding the helper just forces you to maintain 2 functions and comments for them, making it more burden to read the code.

torchbenchmark/score/compute_score.py
wconstab: also wondering if it makes sense for the user to call setup_benchmark_norms, or should it be called by __init__?  Is there a valid case to construct TorchBenchScore without providing ref_data? if so, this is good; if not, maybe just force ref_data be provided for __init__ and call this automatically

torchbenchmark/score/compute_score.py
wconstab: if you feel it's easier to construct a data structure and then use it to guide the iterator for the score computation, wouldn't it be better to construct the data structure using the ref_data rather than the computation data?

torchbenchmark/score/compute_score.py
nikithamalgifb: Fixed in the new revision

torchbenchmark/score/compute_score.py
nikithamalgifb: Fixed

torchbenchmark/score/compute_score.py
nikithamalgifb: Ok, agreed. Moved from the class

torchbenchmark/score/compute_score.py
nikithamalgifb: Ok, done.

torchbenchmark/score/compute_score.py
nikithamalgifb: I have moved this dictionary construction in `get_score_per_config`.

torchbenchmark/score/compute_score.py
nikithamalgifb: I have added ref_data in `init`. I have created  `TORCHBENCH_V0_SCORE` which points to `torchbench_0.0.yaml`.I know we discussed creating it as object rather than a variable like this, but somebody who wants to use this as a reference data can easily import and pass this as the ref_data in the init of the Torchbench class. I have also created a placeholder for `TORCHBENCH_V1_SCORE`, so once available we just have to update the name of the file. If we create a TorchBenchScore Object for v0_score, the one thing I was concerned about was the spec and target gets set to the default spec and target values that was used while constructing the v0 score object. If someone wants to even try out a different spec or target values that wouldn't be possible. Wdyt?

nikithamalgifb: Well, I can provide a configure method inside TorchBenchScore class to manipulate the spec and target values if we want to go by this approach.

wconstab: I think it's more important to make it easy to grab the already configured v0 or v1 score, not to customize it.  The intent behind the v0 or v1 score is to make sure you're doing it _exactly_ the same way as CI.  I think it's a bad idea to let people think they are computing a 'customized v0 score' since any change makes it really not a v0 score anymore.

for the case of plotting relative score sweeps, it's more important to be flexible.  You can either use the ref_data as the source of truth (assume all the models present in it should be part of the score), or possibly do that but also add an exclude set so that you can make a relative score plot that skips some 'bad' models in your data.  This can be useful when coming up with a new score formula for example.

torchbenchmark/score/compute_score.py
```diff
@@ -7,32 +7,130 @@
...
 
+class TorchBenchScore:
+    def __init__(self, spec=SPEC_FILE_DEFAULT, target=TARGET_SCORE_DEFAULT, ref_data=None):
+        self.spec = spec
+        self.target = target
+        self.ref_data = ref_data
+        self.weights = None
+        self.norm = None
+
+    def setup_weights(self):
```
wconstab, line 43: i think these 2 functions are good as-is, although i would suggest 2 things
- put _ in the name if it's an internal helper function not intended to be called
- call both from __init__ unless there's a good reason to not have it be ready to use once init'ed?

torchbenchmark/score/compute_score.py
```diff
@@ -7,32 +7,130 @@
...
+            if Path(self.ref_data).exists():
+                with open(self.ref_data) as ref_file:
+                    ref = yaml.full_load(ref_file)
+                self.norm = {b: ref['benchmarks'][b]['norm'] for b in ref['benchmarks']}
+            else:
+                raise ValueError ("File {file} doesn't exist ".format(file=self.ref_data))
+        else:
+            self.norm = {b['name']: b['stats']['mean'] for b in self.ref_data['benchmarks']}
+
+    def get_score_per_config(self, data):
```
wconstab, line 82: does this function serve much of a purpose on its own? if not, maybe just fold it into compute_score?

nikithamalgifb: This will be helpful when we are plotting the scores for all the config in `plot_sweep.py`. We can directly call this function to get all the scores per config.

torchbenchmark/score/compute_score.py
nikithamalgifb: Agreed and fixed now.

torchbenchmark/score/compute_score.py
nikithamalgifb: Agreed and fixed except exclude_list.  For the exclude_list, I'll open up a follow up PR for this.

torchbenchmark/score/compute_score.py
wconstab: oh, i see. good point. forgot about that.

torchbenchmark/score/compute_score.py
```diff
@@ -7,32 +7,137 @@
...
+                benchmarks = tasks[task]
+                benchmark_weight = 1.0 / len(benchmarks)
+                self.weights[task] = domain_weight * task_weight * benchmark_weight
+
+    def _setup_benchmark_norms(self):
+        """
+        Helper function which gets the normalization values per benchmark
+        by going through the reference data file.
+        """
+        if not self.ref_data: return
```
wconstab, line 76: why is it valid to not have ref_data?  shouldn't this be an error and then you don't need the elif?

torchbenchmark/score/compute_score.py
```diff
@@ -7,32 +7,137 @@
...
+import re
 import yaml
+import importlib
 
 from tabulate import tabulate
+from pathlib import Path
+from collections import defaultdict
 
-SPEC_FILE_DEFAULT = "score.yml"
+SPEC_FILE_DEFAULT = "torchbenchmark/score/score.yml"
```
wconstab, line 18: these kind of hardcode paths should be changed to something based on __file__ probably.  otherwise it can break.

Path(__file__).parent() / "score.yml"
does that work? not sure if .parent() exists

torchbenchmark/score/compute_score.py
nikithamalgifb: This is to handle the case, when no ref_data is provided in the init. In this case we delay the construction of norm till compute_score is called and ref_data is then initialized  to data.

torchbenchmark/score/compute_score.py
```diff
@@ -7,32 +7,137 @@
...
-    score = target * math.exp(score)
-    assert abs(weight_sum - 1.0) < 1e-6, f"Bad configuration, weights don't sum to 1, but {weight_sum}"
-    return score
+def _get_model_task(model_name):
+    """
+    Helper function which extracts the task the model belongs to
+    by iterating over the Model attributes.
+    """
+    MODEL_PATH="../../models"
+    p = Path(__file__).parent.joinpath(MODEL_PATH + model_name)
```
wconstab, line 29: oh. i guess parent does exist!  

wait, i'm confused - isn't a better way here
p = Path(__file__).parent.parent / "models" / model_name
?
I don't like the "../../" explicit string if avoidable, and i'm not sure why you go 2 levels higher than parent?  shouldn't it be one level higher than parent?

torchbenchmark/score/compute_score.py
```diff
@@ -7,32 +7,137 @@
...
+    if(p):
+        module = importlib.import_module(f'torchbenchmark.models.{model_name}', package=__name__)
+        Model = getattr(module, 'Model')
+    return Model.task.value
 
+class TorchBenchScore:
+    def __init__(self, spec=SPEC_FILE_DEFAULT, target=TARGET_SCORE_DEFAULT, ref_data=None):
```
wconstab, line 36: I think ref_data should be a required argument, at least in the current design.
a) not clear what the point of constructing a score without a ref data is
b) you don't provide a way to add ref_data later anyway

given these, i would say it's simpler/cleaner to just require it done in _init_ and then delete the if/else checks in other parts of code.  Once you construct the score, it's ready to use- easy to reason about that way.

torchbenchmark/score/compute_score.py
```diff
@@ -7,32 +7,137 @@
...
+
+        return score_db
+
+    def compute_score(self, data):
+        """
+        This API calculates the total score for all the benchmarks
+        that was run  by reading the data (.json) file.
+        The weights are then calibrated to the target score.
+        """
+        if not self.ref_data:
```
wconstab, line 131: pretty unclear why you would want to do this: wouldn't you always get a score of TARGET_SCORE if you use data as ref_data? I would argue that if someone ends up on this codepath, they were using it wrong.  lmk if you had a particular use case in mind.

torchbenchmark/score/compute_score.py
wconstab: oh, backing up a step, it doesn't seem like you need to even check the path exists.

you aren't really being safe by checking the path exists and then doing import_module.  import_module could still fail, if for some reason the model code is broken.  So I'd suggest just skip the path exists check, and wrap import_module in a try: /except, raising a RuntimeError(f"Unable to get task for model {model_name}")

torchbenchmark/score/compute_score.py
nikithamalgifb: So you would only initialize the ref_data once. So the first file in the data dir will be used as the ref_data and the score for that data will be `TARGET_SCORE` and rest of the scores will be compared against that.  The idea was to initialize the ref_data to first file in the data directory incase no ref_data was provided.

torchbenchmark/score/compute_score.py
nikithamalgifb: Fixed now.

torchbenchmark/score/compute_score.py
nikithamalgifb: Agreed and fixed.

torchbenchmark/score/compute_score.py
nikithamalgifb: Fixed

compute_score.py
```diff
@@ -7,63 +7,51 @@
...
-                    if keyword.lower() in b['name'].lower():
-                        fake_data[b['name']] =  b['stats']['mean'] * float(factor)
-
-            hacked_score = compute_score(config, data, fake_data)
-            print(f"Using hacks {args.hack_data}, hacked_score {hacked_score}")
-
+            benchmark_data.append(data)
     elif args.benchmark_data_dir is not None:
-        scores = [('File', 'Score', 'PyTorch Version')]
         for f in os.listdir(args.benchmark_data_dir):
```
xuzhao9, line 36: Python lists directory in random order (https://docs.python.org/3/library/os.html#os.listdir). It is better to sort the file by name before using the first file as the reference.

compute_score.py
```diff
@@ -7,63 +7,51 @@
...
+                    files.append(f)
+                    benchmark_data.append(data)
+
+    if args.configuration == "relative":
+        ref_data = benchmark_data[0]
+        score_config = TorchBenchScore(ref_data)
+    elif args.configuration == "v0":
+        score_config = TORCHBENCH_V0_SCORE
+    else:
+        raise ValueError("Invalid score configuration")
```
xuzhao9, line 50: Is it a problem if the reference json and the SPEC_FILE_DEFAULT doesn't match? E.g. only part of the benchmarks in SPEC_FILE_DEFAULT is in reference json.

plot_sweep.py
```diff
@@ -20,36 +20,33 @@
...
-from torchbenchmark.score.compute_score import compute_score
-from torchbenchmark.score.generate_score_config import generate_bench_cfg
-
+from torchbenchmark.score.compute_score import TorchBenchScore
 
 if __name__ == "__main__":
     parser = argparse.ArgumentParser(description=__doc__)
-    parser.add_argument("data_dir", nargs='+', 
-                        help="One or more directories containing benchmark json files. " 
+    parser.add_argument("data_dir", nargs='+',
```
xuzhao9, line 27: Also sort the file before determining the first json in the first directory. Otherwise the result may not be stable.

xuzhao9 (changes requested):

compute_score.py
wconstab: wow i didn't realize that.

compute_score.py
nikithamalgifb: So only those benchmarks which are available in the reference json, will be accounted for scoring. So it should not be a problem if the ref json contains only part of the SPEC_FILE_DEFAULT.

compute_score.py
nikithamalgifb: Fixed

plot_sweep.py
nikithamalgifb: Here, we enforce the user to provide the reference json. So this check is not required

compute_score.py
```diff
@@ -7,63 +7,51 @@
...
-            config = yaml.full_load(cfg_file)
-    else:
-        if args.benchmark_data_file is None and args.benchmark_data_dir is None:
-            parser.print_help(sys.stderr)
-            raise ValueError("Invalid command-line arguments. You must specify a config, a data file, or a data dir.")
-        with open(SPEC_FILE_DEFAULT) as spec_file:
-            spec = yaml.full_load(spec_file)
+    if args.benchmark_data_file is None and args.benchmark_data_dir is None:
+        parser.print_help(sys.stderr)
+        raise ValueError("Invalid command-line arguments. You must specify a config, a data file, or a data dir.")
```
xuzhao9, line 26: Since the `--configuration` option defaults to relative, is it better to change the error message to "Invalid command-line arguments. You must specify a data file or a data dir."?

compute_score.py
xuzhao9: Makes sense!

plot_sweep.py
```diff
@@ -20,36 +20,33 @@
...
+                        help="One or more directories containing benchmark json files. "
                              "Each directory will be plotted as a separate series. "
                              "By default, the first file in the first directory will be used"
                              " to generate a score configuration with a target of 1000,"
                              " and everything else will be relative to that.")
     parser.add_argument("--output_html", default='plot.html', help="html file to write")
-    parser.add_argument("--score_heirarchy", default='torchbenchmark/score/score.yml', 
+    parser.add_argument("--score_heirarchy", default='torchbenchmark/score/score.yml',
```
xuzhao9, line 34: Maybe use "compute_score.SPEC_FILE_DEFAULT" to replace the default value here?

torchbenchmark/score/generate_score_config.py
```diff
@@ -49,43 +49,6 @@
...
-                    for b in found_benchmarks:
-                        weight = domain_weight * task_weight * benchmark_weight * config_weight
-                        cfg['benchmarks'][b] = {
-                                'weight': weight,
-                                'norm': benchmark_norms[b],
-                        }
-    return cfg
-
 # Support generate a config from benchmark data that runs partial of the spec
 def generate_bench_cfg_partial(spec, norm, target):
```
xuzhao9, line 53: Sorry I don't quite understand why this function is being removed. It seems still useful to generate config yaml from json. For example, to generate future V1 config.

xuzhao9 (approved):
Overall LGTM. Leave a few comments to be addressed.

compute_score.py
nikithamalgifb: Yeah, makes sense. Let me update this.

plot_sweep.py
nikithamalgifb: Yeah, I can change that.

torchbenchmark/score/generate_score_config.py
nikithamalgifb: So instead of this function, the functionality of this has been replicated in the class `TorchBenchScore`. This class basically calculates the weights by ready the config.yml and also calculates score per configuration(like., CPU/GPU, Eager/JIT, eval/train), which will later be used  to generate 8 plots for the configurations. To generate future V1 config, all we have to do is create an object of this class. This is similar to generating TORCHBENCH_V0_SCORE. I already have  a place holder to create the future V1 config.

PR #256, opened 2021-02-14
Add FX-based profiler

jansel:
Perhaps others would find this useful. [Results as of last week.](https://docs.google.com/spreadsheets/d/1K3DBia64Q0xrXLDJ2ZWGWkUIBVAsU0CYh-8soWBRqKg/edit?usp=sharing)

### DIFF
--- a/test.py
+++ b/test.py
@@ -34,9 +34,11 @@ class TestBenchmark(TestCase):
         except ImportError:  # older versions of PyTorch
             raise unittest.SkipTest("Requires torch>=1.8")
         from fx_profile import main, ProfileAggregate
-        with patch.object(ProfileAggregate, "save"):
+        with patch.object(ProfileAggregate, "save") as mock_save:
             # just run one model to make sure things aren't completely broken
             main(["--repeat=1", "--filter=pytorch_struct", "--device=cpu"])
+            self.assertGreaterEqual(mock_save.call_count, 1)
+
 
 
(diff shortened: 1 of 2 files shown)


jansel:
@wconstab you think it makes sense to merge this?  Or just leave it as a standalone thing.

wconstab:
@jansel yea I'd definitely merge it in.  I will review now.  I'm wondering if this is something we'd want to run every nightly or at some interval, or just for manual use.

test.py
```diff
@@ -27,8 +27,16 @@ def tearDown(self):
             self.assertEqual(self.memory, memory)
             torch.cuda.empty_cache()
 
-def run_model(model_class, model_path, device):
-    m = model_class(device=device)
+    def test_fx_profile(self):
```
wconstab, line 30: I guess this is a correctness test rather than something you're using to gather data.  Did you confirm if this gets invoked in CI as you wrote it? I think we run all tests from test.py.

jansel: Correct, this is just a correctness test.  I verified it locally, I did not check the CI though.

wconstab (approved):
lgtm, and I confirmed there is one new test (157 on your branch vs 156 master) being run in CI although the test names aren't logged.

Depending on what people want to do with this we can talk about automating it in CI to collect stats or whatever.  But seems good/useful as-is so please merge.

PR #257, opened 2021-02-16
clean up `get_module` for yolov3

Krovatkin:
We should also clean up `training` but it looks a little bit more involved, so I'll do it in a separate PR.


xuzhao9:
Fixed in https://github.com/pytorch/benchmark/pull/630

PR #258, opened 2021-02-18
Add PR bisection feature to 0.1 branch

xuzhao9:
Runs bisection to determine PRs that cause performance regression.
Performance regression is defined by TorchBench score dropping greater than the threshold.

Bisection Algorithm: for the bisection range [start, end]
Step 1: Fetch commit list from pytorch git repo: [start, ..., mid, ..., end]
Step 2: Enqueue pair (start, end) into queue `bisectq`
Step 3: Dequeue the first pair (start, end) from `bisectq`.
Step 4: Run TorchBench for start and end and get their scores. If `abs(start.score-end.score)>threshold` ...
             ... and start/end are adjacent, add pair(start, end) to result json.
Step 5: If abs(start.score-end.score)>threshold, but start/end are not adjacent, test benchmark on commit `mid`, and:
              if abs(end.score - mid.score) > threshold: enqueue (mid, end) into `bisectq`
              if abs(start.score - mid.score) > threshold: enqueue (start, mid) into `bisectq`
Step 6: goto step 2 unless `bisectq` is empty

Known limitations:
- Dependencies such as torchtext and torchvision need to be recompiled after building pytorch. Currently, the version of dependencies are fixed to the latest git version.


torchbenchmark/__init__.py
```diff
@@ -2,7 +2,6 @@
 from pathlib import Path
 import subprocess
 import sys
-import torch
```
xuzhao9, line 5: Remove the torch import here because it is not needed and will create problem if we import torchbenchmark without torch installed. In this case, torchbenchmark is imported at initialization, where torch is not yet installed.

.github/scripts/run-bisection.sh
```diff
@@ -0,0 +1,26 @@
...
+
+BISECT_CONFIG=${BISECT_BASE}/config.env
+set -a
+source $BISECT_CONFIG
+set +a
+
+# get torch_nightly.html
+curl -O https://download.pytorch.org/whl/nightly/cu102/torch_nightly.html 
+
+. activate ${BISECT_CONDA_ENV} &> /dev/null
```
wconstab, line 17: can you better-document the intended API of this script at the top? e.g. which env vars are supposed to be set for it to work, and what args to provide on CLI?

.github/scripts/run-nodocker.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/sh
```
wconstab, line 1: what is this script for? 
please answer not in github review but in comment on the header of the file :)

bisection.py
```diff
@@ -0,0 +1,373 @@
+"""bisection.py
+Runs bisection to determine PRs that cause performance regression.
+Performance regression is defined by TorchBench score drop greater than the threshold.
+By default, the torchvision, torchaudio, and torchtext package version will be fixed to the latest git version.
+
+Usage:
+  python bisection.py --pytorch-src <PYTORCH_SRC_DIR> \
```
wconstab, line 7: does this script assume that PYTORCH_SRC_DIR is a git repo, and in a 'clean' state, and it's OK to change branches on it?

bisection.py
```diff
@@ -0,0 +1,373 @@
...
+        self.commit_date = datetime.strptime(last_commit_date, "%Y-%m-%d")
+        return True
+    
+    def get_mid_commit(self, left: Commit, right: Commit) -> Optional[Commit]:
+        left_index = self.commit_dict[left.sha]
+        right_index = self.commit_dict[right.sha]
+        if right_index == left_index + 1:
+            return None
+        else:
+            return self.commits[(left_index + right_index) / 2]
```
wconstab, line 92: does this work? i would think that sometimes there are an odd number divided by 2 and you get a float 1.5 index into a dict.

bisection.py
```diff
@@ -0,0 +1,373 @@
...
+        env["USE_CUDNN"] = "1"
+        env["CMAKE_PREFIX_PATH"] = env["CONDA_PREFIX"]
+        return env
+
+    # Install dependencies such as torchaudio, torchtext and torchvision
+    def install_deps(self, build_env):
+        # Build torchvision
+        print(f"Building torchvision ...", end="", flush=True)
+        command = "python setup.py install &> /dev/null"
+        subprocess.check_call(command, cwd=TORCHBENCH_DEPS["torchvision"], env=build_env, shell=True)
```
wconstab, line 109: a little nit pick, i think it's better to give lists to subprocess for cmd.  e.g.
['python', 'setup.py', ..] 
as well as there are ways to let subprocess capture the output of the process for you using PIPE and not just discard to /dev/null.  I think that can be changed later, but happy to help with pointers to the right methods if you want to do it.

bisection.py
```diff
@@ -0,0 +1,373 @@
...
+ 
+    def build(self, commit: Commit):
+        # checkout pytorch commit
+        gitutils.checkout_git_commit(self.srcpath, commit.sha)
+        # setup environment variables
+        build_env = self.setup_build_env(os.environ.copy())
+        # build pytorch
+        print(f"Building pytorch commit {commit.sha} ...", end="", flush=True)
+        # pytorch doesn't update version.py in incremental compile, so generate manually
+        command = "python tools/generate_torch_version.py --is_debug on"
```
wconstab, line 130: so this part just forces regen of version, since you changed the git checkout and recompile the same build tree i think?

bisection.py
```diff
@@ -0,0 +1,373 @@
...
+        print("done")
+        # Build torchtext
+        print(f"Building torchtext ...", end="", flush=True)
+        command = "python setup.py clean install &> /dev/null"
+        subprocess.check_call(command, cwd=TORCHBENCH_DEPS["torchtext"], env=build_env, shell=True)
+        print("done")
+ 
+    def build(self, commit: Commit):
+        # checkout pytorch commit
+        gitutils.checkout_git_commit(self.srcpath, commit.sha)
```
wconstab, line 124: i'll have to look at checkout_git_commit impl to see, but did you make sure to update submodules recursively when checking out?

bisection.py
```diff
@@ -0,0 +1,373 @@
...
+    bench: TorchBench
+    output_json: str
+
+    def __init__(self,
+                 workdir: str,
+                 torch_src: str,
+                 bench_src: str,
+                 start: str,
+                 end: str,
+                 threshold: int,
```
wconstab, line 255: maybe should be a float, im sure int is good enough for now

bisection.py
```diff
@@ -0,0 +1,373 @@
...
+    parser.add_argument("--work-dir",
+                        help="bisection working directory",
+                        type=exist_dir_path,
+                        required=True)
+    parser.add_argument("--pytorch-src",
+                        help="the directory of pytorch source code git repository",
+                        type=exist_dir_path,
+                        required=True)
+    parser.add_argument("--torchbench-src",
+                        help="the directory of torchbench source code git repository",
```
wconstab, line 338: good to mention for each repo whether it will be modified by the script;
and whether it's required to be in a clean state before running.

relatedly, is it very useful to provide your own pytorch repo, or should bisector just clone its own repo into either a tmpdir or just a ~/.torchbench folder it controls?  (at least by default) 

i'm fine with leaving as-is, maybe mark this as a change for future if anything.

bisection.py
```diff
@@ -0,0 +1,373 @@
...
+                        help="7-digit SHA hash of the start commit to bisect",
+                        required=True)
+    parser.add_argument("--end",
+                        help="7-digit SHA hash of the end commit to bisect",
+                        required=True)
+    parser.add_argument("--threshold",
+                        help="the torchbench score threshold to report a regression",
+                        type=float,
+                        required=True)
+    parser.add_argument("--bmfilter",
```
wconstab, line 351: what happens if unspecified?

i think it's good to have a mode that focuses on one test like we discussed.

but when running a 'score' driven bisection, specifying filter is a little awkward.  instead might be easier to just specify score version and the filter is derived from the score config?

torchbenchmark/util/gitutils.py
```diff
@@ -0,0 +1,116 @@
+"""gitutils.py
+
+Utils for getting git-related information.
```
wconstab, line 3: might be worth using this rather than maintaining our own? 
https://gitpython.readthedocs.io/en/stable/tutorial.html

wconstab (approved):
lgtm, some comments/changes requested but OK to merge first and address later as you prefer.

bisection.py
xuzhao9: Yes. This is to workaround the problem that pytorch doesn't update version after incremental build.

bisection.py
xuzhao9: Yes, I did "git submodule sync && git submodule update --init --recursive" after each checkout.

.github/scripts/run-bisection.sh
xuzhao9: Now it requires only one external file to work: the "config.yaml". I have added documentations to the file header.

bisection.py
xuzhao9: Fixed by using bisection config yaml.

bisection.py
xuzhao9: I think the bisector can clone its own repo, but it can use an option to specify build artifact caches instead (can be ~/.torchbench by default).

bisection.py
xuzhao9: Fixed by converting to float

bisection.py
xuzhao9: Sometimes there will be untracked files after checkout and it can't be removed, otherwise repo won't compile. Seems no actions needed in this case.

bisection.py
xuzhao9: Fixed. Thanks for pointing out.

bisection.py
xuzhao9: Yes it does. Maybe I should check the repo state before modifying it.

PR #260, opened 2021-02-18
add NVFuser

Krovatkin:
Add an option to run NVFuser. The flags enabled are the ones Jie told me to set to enable NVFuser.


xuzhao9 (approved):
LGTM. `fuser` is not used in the CI so we just set its default to None. Could you please clarify in the argument doc(https://github.com/pytorch/benchmark/blob/b1aae816b80926654414c76b206472f80cce8efa/conftest.py#L8) about what is this argument and its usage?

Issue #259, opened 2021-02-18
v0 torchbench (and master) broken by BC change to torchtext


wconstab:
fix both v0 and master branches so that both old and new torchtext works.  
 - try to import the old way, if it fails, try to import the new way

https://github.com/pytorch/text/commits/master

see this error 
ImportError: cannot import name 'TranslationDataset' from 'torchtext.datasets' (/home/ec2-user/anaconda3/envs/gh578224670/lib/python3.7/site-packages/torchtext/datasets/__init__.py)

from nightly CI
https://github.com/pytorch/benchmark/runs/1927202422?check_suite_focus=true

but look for any related imports of the other renamed packages, since the CI stopped at first failure.

nikithamalgifb:
Now fixed in both master and v0. PR's for the fixes - #269, #266.  Closing this bug.

PR #261, opened 2021-02-18
Fix torchtext imports

nikithamalgifb:
Fixes issue #259

### DIFF
--- a/torchbenchmark/models/attention_is_all_you_need_pytorch/__init__.py
+++ b/torchbenchmark/models/attention_is_all_you_need_pytorch/__init__.py
@@ -7,8 +7,14 @@ from tqdm import tqdm
 import torch
 import torch.nn.functional as F
 import torch.optim as optim
-from torchtext.data import Field, Dataset, BucketIterator
-from torchtext.datasets import TranslationDataset
+import torchtext.data
+from torchtext.data import Dataset, BucketIterator
+if torchtext.__version__ >= "0.9":
+    from torchtext.legacy.data import Field
+    from torchtext.legacy.datasets.translation import TranslationDataset
+else:
+    from torchtext.data import Field
+    from torchtext.datasets import TranslationDataset
 
 from .transformer import Constants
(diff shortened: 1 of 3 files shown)


nikithamalgifb:
The torchtext imports haven't been updated yet for torchtext installed via: https://download.pytorch.org/whl/nightly/cu110/torch_nightly.html

So these changes are irrelevant for now in Master. Once, this nightly whl is updated, we can then merge this PR.

torchbenchmark/models/attention_is_all_you_need_pytorch/__init__.py
```diff
@@ -7,8 +7,14 @@
 import torch
 import torch.nn.functional as F
 import torch.optim as optim
-from torchtext.data import Field, Dataset, BucketIterator
-from torchtext.datasets import TranslationDataset
+import torchtext.data
+from torchtext.data import Dataset, BucketIterator
+if torchtext.__version__ >= "0.9":
```
wconstab, line 12: does it work to say 'if torchtext.__version__' 
without first importing torchtext?

wconstab:
@xuzhao9 can you help with the questions around cuda version and torchtext wheel?  seems there is a difference in torchtext version available for 10.2 and 11 cuda.  and maybe our CI config is not all using the consistent cuda version yet.

Krovatkin:
@wconstab the script to build torchbench against custom pytorch also breaks because of this.

torchbenchmark/models/attention_is_all_you_need_pytorch/__init__.py
nikithamalgifb: Tried it, it doesn't work.

xuzhao9:
@wconstab @nikithamalgifb 
 Our CI is running only on 0.1 branch and CUDA 10.2. To me the current commit looks fine since it tries to import the latest packages first, and fallback to the old import if unavailable.

wconstab (approved):

PR #263, opened 2021-02-19
GHA workflow for per-pr testing

xuzhao9:
This PR adds GHA tests for the correctness of PRs. It runs on self-hosted runner.
Since multiple jobs on self-hosted runner will be queued, it won't affect the correctness of the nightly CI.


xuzhao9:
Currently GHA self-hosted runner only works single-threaded, which is far from ideal. So we stick to circle ci at the moment.

PR #270, opened 2021-02-22
Enable HT before set pstate freq, and disable after set pstate freq.

xuzhao9:
On a raw Amazon g4dn.metal instance, I found pinning CPU frequency only works when HT is enabled.
Therefore, when fixing the CPU freq with disabled HT, we first enable HT, set the freq, then disable it.


wconstab (commented):
are you sure about this? it's a little funny.. anything in the driver documentation to back this up?

PR #274, opened 2021-02-23
fix vision models to use the correct device

Krovatkin:
fix vision models to use the correct device

### DIFF
--- a/torchbenchmark/models/alexnet/__init__.py
+++ b/torchbenchmark/models/alexnet/__init__.py
@@ -12,10 +12,10 @@ class Model(BenchmarkModel):
         super().__init__()
         self.device = device
         self.jit = jit
-        self.model = models.alexnet()
+        self.model = models.alexnet().to(self.device)
         if self.jit:
             self.model = torch.jit.script(self.model)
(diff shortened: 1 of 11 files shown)


asuhan (approved):

wconstab (changes requested):
Can you also add an assert as @asuhan had described?  I'd like to enforce this so it doesn't happen again.  

I guess it's not possible to do this outside of the benchmark code as things are structured now.  What's the penalty of adding one assert into every measurement?  should be a wash, i guess?

Krovatkin:
> I guess it's not possible to do this outside of the benchmark code as things are structured now. What's the penalty of adding one assert into every measurement? should be a wash, i guess?

we should follow up in another PR since

1. for non-vision models, it's a bit tricky to get at models' inputs, let alone models' outputs.
2. I'm thinking we could do a little bit restructuring to make freezing a little bit easier and hopefully this will let us keep asserts outside.
3. If we decide to do asserts, I'd still like to do measurements to make sure we aren't affecting a score too much.

wconstab (approved):
ok I will not block this then.

PR #280, opened 2021-02-25
Create the bisection github runner

xuzhao9:
People need to manually do the following before running a bisection job:

1. Come up with a cool name of your bisect case, call it ${BISECT_ISSUE}.
2. On the github self-hosted runner, create a bisection work directory at ${HOME}/.torchbench/bisection/${BISECT_ISSUE}
    The directory should include only one file: config.yaml. Example of config.yaml can be found [here](https://github.com/pytorch/benchmark/blob/0.1/.github/scripts/bisection-config.sample.yaml).
3. Specify the BISECT_ISSUE name as the "Bisection Issue Name" field in the workflow trigger UI.
4. Click the "Run workflow" button on master branch.

### DIFF
--- a/.github/workflows/bisection.md
+++ b/.github/workflows/bisection.md
@@ -1,0 +1,13 @@
+# How to manually run a bisection job
+
+To run a bisection job, please follow the instructions below:
+
+1. Create a cool, unique name of your bisect job, call it `${BISECT_ISSUE}`. To avoid name conflict, we suggest using the github issue number.
+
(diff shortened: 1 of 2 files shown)


xuzhao9:
> LGTM but can you take the instructions from the PR and move to a .MD file? Do you anticipate the instructions will be invalid soon or you will keep this working even when you add auto-bisect?

I will create a MD file to explain how to use it. The plan is even later we add auto-bisect, we should still be able to manually run a bisection.

PR #293, opened 2021-03-04
Add dumping IR facility to the benchmarks

penguinwu:
Add the script to dump graph IRs for all the benchmarks

### DIFF
--- a/collect_graph_ir.py
+++ b/collect_graph_ir.py
@@ -1,0 +1,58 @@
+#!/usr/bin/env python
+import argparse
+import gc
+import logging
+import os
+import re
+import warnings
+
+from torchbenchmark import list_models
+import torch
+
+NO_JIT = {"demucs", "dlrm", "maml", "yolov3", "moco", "pytorch_CycleGAN_and_pix2pix", "tacotron2"}
+NO_GET_MODULE = {"Background_Matting"}
+
+def get_dump_filename(name, device):
+    return f"{name}.{device}.last_executed_graph_dump.log"
+
+def iter_models(args):
+    device = "cpu"
+    for benchmark_cls in list_models():
+        #if (not re.search("|".join(args.filter), benchmark_cls.name, re.I) or
+        #        re.search("|".join(args.exclude), benchmark_cls.name, re.I) or
+        if benchmark_cls.name in NO_GET_MODULE:
+            print(f"{benchmark_cls.name} has no get_module, skipped")
+            continue
+        if benchmark_cls.name in NO_JIT:
+            print(f"{benchmark_cls.name} has no scripted module, skipped")
+            continue
+        try:
+            # disable profiling mode so that the collected graph does not contain
+            # profiling node
+            torch._C._jit_set_profiling_mode(False)
+
+            benchmark = benchmark_cls(device=device, jit=True)
+            model, example_inputs = benchmark.get_module()
+
+            if benchmark_cls.name == "BERT_pytorch":
+                # extract ScriptedModule object for BERT model
+                model = model.bert
+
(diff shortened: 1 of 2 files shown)


collect_graph_ir.py
```diff
@@ -0,0 +1,58 @@
...
+        if benchmark_cls.name in NO_GET_MODULE:
+            print(f"{benchmark_cls.name} has no get_module, skipped")
+            continue
+        if benchmark_cls.name in NO_JIT:
+            print(f"{benchmark_cls.name} has no scripted module, skipped")
+            continue
+        try:
+            # disable profiling mode so that the collected graph does not contain
+            # profiling node
+            torch._C._jit_set_profiling_mode(False)
```
bertmaher, line 32: I'm a little wary of this because the non-profiling executor isn't a default mode anymore.  I don't know how different the graphs would look but I don't know that it's a good idea to collect stats from a different mode than we run with.

To get rid of the profile nodes, I'd suggest adding a "warmup" before dumping stats:
```
for _ in range(torch._C._jit_num_profiled_runs()):
  model(*example_inputs)
```

collect_graph_ir.py
```diff
@@ -0,0 +1,58 @@
...
+NO_JIT = {"demucs", "dlrm", "maml", "yolov3", "moco", "pytorch_CycleGAN_and_pix2pix", "tacotron2"}
+NO_GET_MODULE = {"Background_Matting"}
+
+def get_dump_filename(name, device):
+    return f"{name}.{device}.last_executed_graph_dump.log"
+
+def iter_models(args):
+    device = "cpu"
+    for benchmark_cls in list_models():
+        #if (not re.search("|".join(args.filter), benchmark_cls.name, re.I) or
```
bertmaher, line 21: Minor: remove the commented code.

collect_graph_ir.py
```diff
@@ -0,0 +1,58 @@
...
+            fname = get_dump_filename(benchmark.name, device)
+            print(f"Dump Graph IR for {benchmark.name} to {fname}")
+            with open(fname, 'w') as dump_file:
+                print(model.graph_for(*example_inputs), file=dump_file)
+        except NotImplementedError:
+            print(f"Cannot collect graph IR dump for {benchmark.name}")
+            pass
+
+def main(args=None):
+    #parser = argparse.ArgumentParser()
```
bertmaher, line 50: Commented code here too.

bertmaher (approved):
Looks good, but do consider my suggestion about warming up rather than disabling the profiling executor.

wconstab (approved):
lgtm other than comments from Bert. Thanks for adding this!

collect_graph_ir.py
penguinwu: Good suggestion. I made the default collection to use the default setting (i.e., profile executor enabled and warmed up). Bu I added additional options to dump graphs as `inlined_graph` and also dump graph w/o profiling executor. Sometimes we need to see the graph before profiling executor to check if TS IR level optimizations can handle them (e.g., for symbolic shape analysis or alias analysis).

```
usage: collect_graph_ir.py [-h] [--benchmark BENCHMARK] [--no_profiling] [--inlined_graph]

dump last_executed graph for all benchmarks with JIT implementation

optional arguments:
  -h, --help            show this help message and exit
  --benchmark BENCHMARK, -b BENCHMARK
                        dump graph for <BENCHMARK>
  --no_profiling        dump last_executed graphs w/o profiling executor
  --inlined_graph       dump graphs dumped by module.inlined_graph
```

collect_graph_ir.py
penguinwu: Cleaned up

PR #298, opened 2021-03-05
write_sys_file expects str or write throws

Krovatkin:
It's better to just cast a value to string in `write_sys_file` but this at least fixes the current issue for now while I remember about it.


wconstab:
thats my bad, i think I assumed that by marking the type annotation on the func it would actually do something like a cast.

wconstab (approved):

PR #301, opened 2021-03-08
Set device/jit properties before we return from c-tor

Krovatkin:
Otherwise, we can't access these properties consistently in `test_bench.py` for diagnostics.


wconstab (approved):
lgtm thanks for fixing!

PR #315, opened 2021-03-11
Record PyTorch git version and cuda version in the result.

xuzhao9:
CUDA version information is useful if we want to abtest CUDA regression in the future.
Adding PyTorch git version to easier nail down the exact commit used by TorchBench.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -42,6 +42,9 @@ def pytest_configure(config):
 
 def pytest_benchmark_update_machine_info(config, machine_info):
     machine_info['pytorch_version'] = torch.__version__
+    machine_info['pytorch_git_version'] = torch.version.git_version
+    machine_info['cuda_version'] = torch.version.cuda
+
     try:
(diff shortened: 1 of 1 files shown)


conftest.py
```diff
@@ -42,6 +42,9 @@ def pytest_configure(config):
 
 def pytest_benchmark_update_machine_info(config, machine_info):
     machine_info['pytorch_version'] = torch.__version__
+    machine_info['pytorch_git_version'] = torch.version.git_version
+    machine_info['cuda_version'] = torch.version.cuda
```
wconstab, line 46: is this safe (in cases where not torch.has_cuda)?

wconstab (approved):
lgtm except the little question

conftest.py
xuzhao9: If cuda is not available, torch.version.cuda will be None (tested on my macbook). Do you think it is fine?

PR #318, opened 2021-03-12
Rewrite demucs to be TorchScript-able

### LINKED ISSUE #311: [Scripting needed] demucs
`demucs` supports inference and training for both cpu and cuda, none of which were scripted.

gmagogsfm:
* Used MonkeyType to annotate types, saved me a lot of time.

fixes #311

### DIFF
--- a/torchbenchmark/models/demucs/__init__.py
+++ b/torchbenchmark/models/demucs/__init__.py
@@ -10,6 +10,10 @@ from .demucs.augment import FlipChannels, FlipSign, Remix, Shift
 from .demucs.utils import capture_init, center_trim
 from ...util.model import BenchmarkModel
 from torchbenchmark.tasks import OTHER
+from torch import Tensor
+from torch.nn.modules.container import Sequential
+from torchbenchmark.models.demucs.demucs.model import Demucs
+from typing import Optional, Tuple
 
 
 torch.manual_seed(1337)
@@ -20,12 +24,12 @@ torch.backends.cudnn.benchmark = False
 
 
 class DemucsWrapper(torch.nn.Module):
-    def __init__(self, model, augment):
+    def __init__(self, model: Demucs, augment: Sequential) -> None:
         super(DemucsWrapper, self).__init__()
         self.model = model
         self.augment = augment
 
-    def forward(self, streams):
+    def forward(self, streams: Tensor) -> Tuple[Tensor, Tensor]:
         sources = streams[:, 1:]
         sources = self.augment(sources)
         mix = sources.sum(dim=1)
@@ -34,7 +38,7 @@ class DemucsWrapper(torch.nn.Module):
 
 class Model(BenchmarkModel):
     task = OTHER.OTHER_TASKS
-    def __init__(self, device=None, jit=False):
+    def __init__(self, device: Optional[str]=None, jit: bool=False) -> None:
         super().__init__()
         self.device = device
         self.jit = jit
@@ -67,10 +71,13 @@ class Model(BenchmarkModel):
 
         self.model = DemucsWrapper(self.model, self.augment)
 
+        if self.jit:
+            self.model = torch.jit.script(self.model)
+
     def _set_mode(self, train):
         self.model.train(train)
 
-    def get_module(self):
+    def get_module(self) -> Tuple[DemucsWrapper, Tuple[Tensor]]:
         self.model.eval()
         return self.model, self.example_inputs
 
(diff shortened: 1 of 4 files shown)


torchbenchmark/models/demucs/__init__.py
```diff
@@ -34,7 +38,7 @@ def forward(self, streams):
 
 class Model(BenchmarkModel):
     task = OTHER.OTHER_TASKS
-    def __init__(self, device=None, jit=False):
+    def __init__(self, device: Optional[str]=None, jit: bool=True) -> None:
```
penguinwu, line 41: Let's set `jit` default value to be `False` in the Model constructor as all the other benchmarks do. We will create Model instances for `jit=True` for jit-runs.

torchbenchmark/models/demucs/demucs/augment.py
```diff
@@ -23,7 +23,7 @@ def forward(self, wav):
             if not self.training:
                 wav = wav[..., :length]
             else:
-                offsets = th.randint(self.shift, [batch, sources, 1, 1], device=wav.device)
+                offsets = th.randint(self.shift, [batch, sources, 1, 1], device=wav.device, dtype=th.int64)
```
penguinwu, line 26: why do we need an additional parameter here?

torchbenchmark/models/demucs/demucs/augment.py
```diff
@@ -37,7 +37,7 @@ class FlipChannels(nn.Module):
     def forward(self, wav):
         batch, sources, channels, time = wav.size()
         if self.training and wav.size(2) == 2:
-            left = th.randint(2, (batch, sources, 1, 1), device=wav.device)
+            left = th.randint(2, (batch, sources, 1, 1), device=wav.device, dtype=th.int64)
```
penguinwu, line 40: why need an additional parameter here?

torchbenchmark/models/demucs/demucs/augment.py
```diff
@@ -77,7 +77,10 @@ def forward(self, wav):
         device = wav.device
 
         if self.training:
-            group_size = self.group_size or batch
+            if self.group_size is not None:
```
penguinwu, line 80: Is this rewrite because we do not support Python's `or` semantics for `None`? If so, this seems a common pattern, and could be a usability improvement.

torchbenchmark/models/demucs/demucs/augment.py
gmagogsfm: Ditto

torchbenchmark/models/demucs/demucs/model.py
```diff
@@ -64,19 +66,19 @@ class Demucs(th.jit.ScriptModule):
...
+                 depth: int=6,
+                 rewrite: bool=True,
+                 glu: bool=True,
+                 upsample: bool=False,
+                 rescale: float=0.1,
+                 kernel_size: int=8,
+                 stride: int=4,
+                 growth: float=2.,
+                 lstm_layers: int=2,
+                 context: int=3) -> None:
```
penguinwu, line 81: One usability improvement thought: if a parameter is unannotated but has a non-`None` default value, we can deduce parameter types based on the default values. I think this would be strictly better than assuming the default `Tensor` type. For default value of `None`, it is better to assume its type to be `Optional[Tensor]` than the default `Tensor` type.

torchbenchmark/models/demucs/demucs/augment.py
gmagogsfm: `offsets` is used later as indices for `wav.gather` and operator schema requires it to be int64.  I don't think this is ideal since int can always be upcast to `int64`, but currently we have to make this change.

torchbenchmark/models/demucs/demucs/utils.py
```diff
@@ -177,7 +177,7 @@ def save_model(model, path):
     th.save((klass, args, kwargs, state), save_to)
 
 
-def capture_init(init):
+def capture_init(init: Callable) -> Callable:
```
penguinwu, line 180: Are we supporting `Callable` now? If so, need to add to the language spec (of course, not to be addressed in this issue)

penguinwu (changes requested):
Look good to me. Just one true change request. The others are thoughts and questions.

BTW, are you able to run the benchmark w/ JIT enabled?

torchbenchmark/models/demucs/demucs/augment.py
gmagogsfm: `or`, TorchScript thinks this `or` makes `group_size` a boolean, even explicit annotation doesn't fix that. Thus resulting in a longer non-pythonic code.

gmagogsfm (commented):
Revised according to comments. PTAL.

penguinwu (approved):

PR #319, opened 2021-03-12
Adding qat model to torchbenchmark

HDCharles:
see https://github.com/pytorch/benchmark/pull/323 instead

* **#319 Adding qat model to torchbenchmark**

Summary: This is primarily just the mobilenet_v2 model with qat prep and conversion added, however some care was necessary
to avoid issues with eval without training.

Test Plan: with-proxy pytest test_bench.py -k "mobilenet_v2" --ignore_machine_config --warmup
Screen Shot 2021-03-12 at 3 09 36 AM

the eval benchmarks for cpu between qat and non qat demonstrate the expected performance increase.

Reviewers:

Subscribers:

Tasks:

Tags:


vkuzo:
lgtm, will wait for @wconstab to accept since I'm not super familiar with this code.  Is there some config that needs to be updated to run this benchmark continuously, or do we need to run it manually?

vkuzo:
also, would it be possible to run the non-quantized benchmark in the same way and add the results here, so we can compare side to side and check if the numbers look reasonable?

torchbenchmark/models/mobilenet_v2_qat/__init__.py
```diff
@@ -0,0 +1,61 @@
...
+        optimizer = optim.Adam(self.model.parameters())
+        loss = torch.nn.CrossEntropyLoss()
+        for _ in range(niter):
+            optimizer.zero_grad()
+            pred = self.model(*self.example_inputs)
+            y = torch.empty(pred.shape[0], dtype=torch.long, device=self.device).random_(pred.shape[1])
+            loss(pred, y).backward()
+            optimizer.step()
+
+    def set_eval(self):
```
wconstab, line 35: do you need to also override set_train?  i think as-written it could work in CI today but if the order of the tests were changed (eval before train) would it still work?  maybe it would be OK since you construct the model in train mode each time, i guess i forgot if the test harness constructs a new model always.

wconstab:
> also, would it be possible to run the non-quantized benchmark in the same way and add the results here, so we can compare side to side and check if the numbers look reasonable?

If merged, this should just start running in CI automatically.  

There is also a unidash set up, but it only tracks the smaller set of models that are on the 'v0' branch.  We'd include this in a v1 dashboard whenever we get around to cutting that.

torchbenchmark/models/mobilenet_v2_qat/__init__.py
HDCharles: So the short answer is that I don't need to override set_train (that setup is located in the initialization) and that no matter what, there's not really a good way to handle eval before train for quantized models (that I know of) without reinitializing things.

As implemented it works with
i) eval only
ii) train only
ii) train then eval

there's a piece in test_bench that implies that if the order is an issue we  may be able to change the scope of the model to get around that

```Any pytest function that consumes a 'modeldef' arg will invoke this
    automatically, and reuse it for each test that takes that combination
    of arguments within the module.

    If reusing the module between tests isn't safe, change 'scope' parameter.```

but I wasn't sure how to do that for the model or if it would require changing it for the entire repo.

HDCharles:
> also, would it be possible to run the non-quantized benchmark in the same way and add the results here, so we can compare side to side and check if the numbers look reasonable?

Yeah, thats whats captured in the test plan, I only ran mobilenet_v2 and mobilenet_v2_qat and I was hoping you could sanity check those numbers for me. The info is setup so test_eval and test_train are for the training and eval runs while mobilenet_v2 vs mobilenet_v2_qat correspond to the quantized and unquantized models. Options like cuda and jit are added at the end.

I was a little surprised to see that qat ran like 3x slower than non qat for cuda training. However the cpu based eval for qat are faster than those for the non_quantized version which makes sense.

torchbenchmark/models/mobilenet_v2_qat/__init__.py
wconstab: IIUC it looks like you are able to run the test_eval and test_train in either order, due to initializing train first in the setup.  I guess that could be thought of as 'train then eval then train' in the worst case, where train happens in 'init', then test_eval runs, then test_train runs.  IF that's true, i don't have any issue with the approach you took.

wconstab (approved):
LGTM pending that last clarification i asked. Thanks for putting this in!

vkuzo:
> Yeah, thats whats captured in the test plan, I only ran mobilenet_v2 and mobilenet_v2_qat and I was hoping you could sanity check those numbers for me. the main thing to look out for is the difference between mobilenet_v2 vs mobilenet_v2_qat since they are named so simmilarly.

Thanks, I missed that.  One option would be to name the quantized model `mobilenet_v2_quantized_qat` to make it clear from the name.

> I was a little surprised to see that qat ran like 3x slower than non qat for cuda training. However the cpu based eval for qat are faster than those for the non_quantized version which makes sense.

The 3x is a bit slower than what we see in production, but not surprising.  I think we can save tuning the settings for a future PR.

torchbenchmark/models/mobilenet_v2_qat/__init__.py
HDCharles: > IIUC it looks like you are able to run the test_eval and test_train in either order, due to initializing train first in the setup. I guess that could be thought of as 'train then eval then train' in the worst case, where train happens in 'init', then test_eval runs, then test_train runs. IF that's true, i don't have any issue with the approach you took.

it will definitely throw an error, the prep_eval step turns it into a quantized form, which doesn't track gradients and doesn't support backprop for the training step. If for example you added self.prep_qat_train() and self.train()to the bottom of the eval function, it won't complete without error.

I'm not sure how to get around that though without entirely reinitializing each time the model is called.

wconstab: i think we can change 
@pytest.fixture(scope='class')
to scope=test, in order to have the fixture construct a new model object for train and then for eval, so it would work in either order.  

that said, i am not sure if i confirmed that understanding before;  also not sure if pytest promises to run tests in a certain order (Relative to how they are listed in the test class, which is train then eval).  If it's at least consistent, i think it's OK to merge.

HDCharles: scope='test' seems to be unrecognized, scope='function' works, would that be sufficient?

wconstab: sorry, yea that sounds right.

HDCharles:
Something weird happened to the PR that threw a bunch of mypy stuff into the diff, I made a clean PR: https://github.com/pytorch/benchmark/pull/323

torchbenchmark/models/mobilenet_v2_qat/__init__.py
HDCharles: alright, I put together another PR, this most recent one grabbed a bunch of mypy garbage and it was easier for me to make a new PR to fix it: https://github.com/pytorch/benchmark/pull/323

I re ran the test and added the name change recommend by @vkuzo, let me know if there's anything else

HDCharles:
See https://github.com/pytorch/benchmark/pull/323 instead

PR #320, opened 2021-03-13
Rewrite DLRM to be TorchScript-able

gmagogsfm:
* The only portion not scriptable is `parallel_forward` calling into `nn.parallel`, but it is only needed for distributed execution, which as far as I know isn't part of plan for benchmarking.


gmagogsfm:
no longer needed

PR #324, opened 2021-03-17
Add --cpu_only option for running benchmarks on cpu machines

qxy11:
Add an option for running benchmarks on laptops locally to prevent cuda-related errors when run on a cpu-only machine to make the tool easier to understand.

Testing:
Run `pytest test_bench.py --ignore_machine_config -v --cpu_only`
The cuda tests are no longer run with the --cpu_only tag included

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -9,11 +9,13 @@ def pytest_addoption(parser):
     parser.addoption("--ignore_machine_config",
                      action='store_true',
                      help="Disable checks/assertions for machine configuration for stable benchmarks")
-    parser.addoption("--disable_nograd", action='store_true', 
+    parser.addoption("--disable_nograd", action='store_true',
                     help="Disable no_grad for eval() runs")
     parser.addoption("--check_opt_vs_noopt_jit",
                      action='store_true',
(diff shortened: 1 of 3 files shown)


conftest.py
```diff
@@ -9,11 +9,13 @@ def pytest_addoption(parser):
     parser.addoption("--ignore_machine_config",
                      action='store_true',
                      help="Disable checks/assertions for machine configuration for stable benchmarks")
-    parser.addoption("--disable_nograd", action='store_true', 
+    parser.addoption("--disable_nograd", action='store_true',
                     help="Disable no_grad for eval() runs")
     parser.addoption("--check_opt_vs_noopt_jit",
                      action='store_true',
                      help="The best attempt to check results for inference runs. Not all models support this!")
+    parser.addoption("--localrun", action='store_true',
```
xuzhao9, line 17: Since the "--localrun" only runs cpu benchmarks, maybe it makes more sense to rename it to "--cpuonly". Also, please mention in the help and doc that specifying this option will automatically ignore the machine config check.

xuzhao9 (approved):
Looking good! Please address the comment so that the meaning of the option is more explicit.

PR #325, opened 2021-03-22
Add script to upload the benchmark result to S3 buckets

xuzhao9:
After the nightly CI finishes, push results to S3 bucket so that [Pytorch CI HUD](https://ezyang.github.io/pytorch-ci-hud/torchbench-v0-nightly) can reads results from S3 buckets.

### DIFF
--- a/scripts/upload_s3.py
+++ b/scripts/upload_s3.py
@@ -1,4 +1,13 @@
 #!/usr/bin/env python
+"""
+The script to upload TorchBench nightly CI result to Amazon S3.
+
+The command
+`upload_s3.py --torchbench-result-dir benchmark-results/result-directory --gen-index --upload-s3`
+will index all directories under `benchmark-results` and generate the `index.json` file.
+Then it will upload the `result-directory` and `index.json` to Amazon S3 bucket.
+
+"""
 
 import argparse
 import json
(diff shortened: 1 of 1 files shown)


scripts/upload_s3.py
```diff
@@ -0,0 +1,104 @@
...
+        # Upload index file
+        index_key = "torchbench_v0_nightly/index.json"
+        upload_s3_file(index_key, index_path)
+        # Upload the result directory
+        basedir = os.path.basename(args.torchbench_result_dir)
+        result_dir = f"torchbench_v0_nightly/{basedir}/"
+        for result_file in os.listdir(args.torchbench_result_dir):
+            result_key = f"{result_dir}/{result_file}"
+            result_path = os.path.join(args.torchbench_result_dir, result_file)
+            if should_upload(result_path):
```
wconstab, line 103: suggest naming 'is_nonempty_json' instead of 'should_upload', as it is more clear what it does

scripts/upload_s3.py
```diff
@@ -0,0 +1,104 @@
...
+        print(f"Uploading file {body_file} to S3 key: {key}")
+        obj = get_S3_object_from_bucket(S3_BUCKET, key)
+        obj.put(Body=body.read())
+
+def should_upload(json_path):
+    # Upload non-empty json files
+    return (json_path.endswith(".json") and os.path.exists(json_path) and os.stat(json_path).st_size)
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description=__doc__)
```
wconstab, line 77: probably good to add some details in __doc__ at the top, explaining what the hierarchy of index and other files is supposed to be and other things about when to use this script.

wconstab (approved):

scripts/upload_s3.py
xuzhao9: Good point!

PR #326, opened 2021-03-22
Added a step to upload to Amazon S3.

xuzhao9:


.github/workflows/main.yml
```diff
@@ -28,3 +30,7 @@ jobs:
         with:
           name: Benchmark result
           path: benchmark-result-*.json
+      - name: Upload results to Amazon S3
+        run: |
+          LAST_MODIFIED_DIR=$(find ${HOME}/benchmark-results-v0.1/ -mindepth 1 -maxdepth 1 -type d -exec stat -t {} \; | sort -r -n -k 13,13 | head -1 | cut -d " " -f 1)
```
wconstab, line 35: is this LAST_MODIFED_DIR command very safe?  would the uploader do something bad if LAST_MODIFIED_DIR is a duplicate of one uploaded before?  or if it is empty?

wconstab (approved):

.github/workflows/main.yml
xuzhao9: If LAST_MODIFIED_DIR is a duplicate, the uploader will upload the file again. Since `--torchbench-result-dir` is a required argument, if it is empty the upload_s3.py will fail with error.

PR #327, opened 2021-03-23
Gh/hd charles/2/base

HDCharles:
These changes add the quantized mobilenet v2 model as an MVP to demonstrate usecase of benchmarking quantization with this setup.


wconstab (approved):
just blindly stamping this time as this is a replica of previously accepted #323

PR #333, opened 2021-03-25
Improve usability of torch_nightly.py

xuzhao9:
Currently, nightly CI suffers problem when torch/torchtext/torchvision nightly build fails. Installing the failed packages from pip will silently fall back to the recent stable versions. For example, if pytorch nightly build fails, it will install torch stable, torchtext nightly, and torchvision nightly.

Therefore, I would like to add some checks before we actually run the nightly CI. The plan is to first call torch_nightly.py to get a list of latest nightly built packages, and only proceed if all packages have the latest build.

### DIFF
--- a/torchbenchmark/util/torch_nightly.py
+++ b/torchbenchmark/util/torch_nightly.py
@@ -1,6 +1,8 @@
 import os
 import re
 import requests
+import argparse
+import urllib.parse
 from datetime import date, timedelta
 from bs4 import BeautifulSoup
 from collections import defaultdict
@@ -26,7 +28,7 @@ def memoize(function):
 def get_wheel_index_data(py_version, platform_version, url=torch_nightly_wheel_index, override_file=torch_nightly_wheel_index_override):
     """
     """
-    if override_file:
+    if os.path.isfile(override_file) and os.stat(override_file).st_size:
         with open(override_file) as f:
             data = f.read()
     else:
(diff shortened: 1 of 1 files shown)


torchbenchmark/util/torch_nightly.py
```diff
@@ -64,32 +67,40 @@ def get_nightly_wheel_urls(packages:list, date:date,
...
 if __name__ == "__main__":
-    from tabulate import tabulate
-    # print(tabulate(get_n_prior_nightly_wheels(['torch', 'torchvision', 'torchtext'], 200)))
-    wheels = get_n_prior_nightly_wheels(['torch', 'torchvision', 'torchtext'], 1)
+    parser = argparse.ArgumentParser(description="Find recent pytorch nightly builds")
+    parser.add_argument("--pyver", type=str, default="cp37", help="PyTorch Python version")
+    parser.add_argument("--platform", type=str, default="linux_x86_64", help="PyTorch platform")
+    parser.add_argument("--priordays", type=int, default=0, help="Number of days")
+    parser.add_argument("--reverse", action="store_true", help="Return reversed result")
+    parser.add_argument("--packages", required=True, type=str, nargs="+", help="List of package names")
```
wconstab, line 97: it looks like you forgot to plumb args.packages into the function call below?

wconstab (approved):
the changes lgtm, although I don't know how it helps the problem of silently using old nightlys in CI?

torchbenchmark/util/torch_nightly.py
xuzhao9: Right, thanks!

xuzhao9:
> the changes lgtm, although I don't know how it helps the problem of silently using old nightlys in CI?

This script will return an empty string if one of the latest packages is missing. If the stdout size is non-zero, we know all packages exist.

PR #335, opened 2021-03-27
Add equation_of_state benchmark from pyhpc-benchmarks

bertmaher:
This is one of the three benchmarks from https://github.com/dionhaefner/pyhpc-benchmarks.  (I'm interested in adding the other 2 as well, this is just where I started in case there's discussion).  It's a scientific computing, rather than deep learning, benchmark, but it's interesting nonetheless, and has good comparisons with tensorflow, jax, etc.

It's really intended to run single-threaded; I'm not sure whether that's easily achievable using our automation.

I used the largest size sample data from the benchmark, which takes about 6 seconds (single threaded).  Or 0.7 seconds, if you enable the CPU fuser ;-)


torchbenchmark/models/pyhpc_equation_of_state/__init__.py
```diff
@@ -0,0 +1,53 @@
...
+
+    def forward(self, s, t, p):
+        return eos_pytorch.gsw_dHdT(s, t, p)
+
+class Model(BenchmarkModel):
+    def __init__(self, device=None, jit=False):
+        super().__init__()
+        self.device = device
+        self.jit = jit
+        self.model = EquationOfState()
```
Chillee, line 33: You need to manually move the model to the appropriate device (this has been the source of a bunch of issues recently).

Chillee: On that note, I'd also be interested in knowing what the CUDA results are :P

torchbenchmark/models/pyhpc_equation_of_state/eos_pytorch.py
```diff
@@ -0,0 +1,199 @@
...
+  Temperature, using the computationally-efficient 48-term expression for
+  density in terms of SA, CT and p (IOC et al., 2010).
+==========================================================================
+"""
+
+import numpy as np
+import torch
+
+
+@torch.jit.script
```
Chillee, line 14: Won't this force it to always script? We should probably remove it.

torchbenchmark/models/pyhpc_equation_of_state/eos_pytorch.py
```diff
@@ -0,0 +1,199 @@
...
+def gsw_dHdT(sa, ct, p):
```
Chillee, line 15: Why don't these guys use autograd? 🤔 I guess in this case they need an AOT autograd to make it run reasonably fast.

Chillee (changes requested):
Some oversights, I think, to make it run correctly.

bertmaher:
With Cuda, I'm seeing 175ms (eager) vs 26 ms (jit).  I also tested with `--fuser old` and saw 54ms.

I did, however notice an odd behavior.  Those numbers are based on running just the `-cuda` tests.  If I run the cpu and cuda variants in the same test run, I see 175ms for both (eager) and (jit), which suggests some unfused graph is being cached (maybe from the cpu variant?).  That... kinda sucks.  I don't know how our CI runs these but it might run into that issue.  I wonder if we could do something to fix this in jit itself...

torchbenchmark/models/pyhpc_equation_of_state/__init__.py
bertmaher: Done, and done!

torchbenchmark/models/pyhpc_equation_of_state/eos_pytorch.py
wconstab: good lord. i would love to have to code-review that formula..

wconstab:
@bertmaher i'm curious, do you know why this model is intended to run single-threaded?  is it just really small tensors?

torchbenchmark/models/pyhpc_equation_of_state/eos_pytorch.py
Krovatkin: @wconstab  @bertmaher  do we want to create a new category for this type of benchmarks?

Krovatkin (approved):
:shipit:

wconstab:
@Krovatkin  Yea i think a new category (HPC? Scientific? is warranted).

torchbenchmark/models/pyhpc_equation_of_state/eos_pytorch.py
bertmaher: > @wconstab @bertmaher do we want to create a new category for this type of benchmarks?

Sure!  It's like HPC, or "scientific computing"

bertmaher:
> @bertmaher i'm curious, do you know why this model is intended to run single-threaded? is it just really small tensors?
> 
> I see they show using taskset -c 0 for this model, but they don't say why.
> 
> I also see they describe the problem as 'embarrassingly parallel, and a GPU's dream'
> https://github.com/dionhaefner/pyhpc-benchmarks/tree/master/benchmarks/equation_of_state
> 
> So what gives?

They give some rationale for the single-thread choice, but it's not totally convincing to me: https://github.com/dionhaefner/pyhpc-benchmarks#why-only-test-sequential-cpu-performance.  Reason (1) sort of boils down to "I just want to know single thread perf", and (2) is "these are easy enough to parallelize with MPI anyways"

It's certainly interesting to know single-thread perf, but I don't see a fundamental reason why it's more meaningful than multi-thread.

wconstab:
@bertmaher interesting i didn't even parse that part of their document being about single-thread but now it makes sense. 

I think that rationale assumes that multiprocess parallelism == multithread parallelism, which for e.g. an Intel CPU with DL related wide vector stuff and libs like MKL probably? isn't true, but I'm not sure.

PR #336, opened 2021-03-28
Add pyhpc-benchmark isoneutral_mixing

bertmaher:
Some perf results on one core (in ms):

device | eager | jit | jit+fuser
-|-|-|-
cuda | 238 | 176 | --
cpu | 3652 | 3341 | 3046

### DIFF
--- a/torchbenchmark/models/pyhpc_isoneutral_mixing/isoneutral_pytorch.py
+++ b/torchbenchmark/models/pyhpc_isoneutral_mixing/isoneutral_pytorch.py
@@ -1,0 +1,196 @@
+import torch
+
+
+def get_drhodT(salt, temp, p):
+    rho0 = 1024.0
+    z0 = 0.0
+    theta0 = 283.0 - 273.15
+    grav = 9.81
+    betaT = 1.67e-4
+    betaTs = 1e-5
+    gammas = 1.1e-8
+
+    zz = -p - z0
+    thetas = temp - theta0
+    return -(betaTs * thetas + betaT * (1 - gammas * grav * zz * rho0)) * rho0
+
+
+def get_drhodS(salt, temp, p):
+    betaS = 0.78e-3
+    rho0 = 1024.
+    return betaS * rho0 * torch.ones_like(temp)
+
+
+def dm_taper(sx):
+    """
+    tapering function for isopycnal slopes
+    """
+    iso_slopec = 1e-3
+    iso_dslope = 1e-3
+    return 0.5 * (1. + torch.tanh((-torch.abs(sx) + iso_slopec) / iso_dslope))
+
+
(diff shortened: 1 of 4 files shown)


torchbenchmark/models/pyhpc_isoneutral_mixing/isoneutral_pytorch.py
```diff
@@ -0,0 +1,213 @@
...
+    if device == 'gpu':
+        torch.cuda.synchronize()
+    return out
+
+
+def run(*inputs, device='cpu'):
+    with torch.no_grad():
+        if device == 'gpu':
+            outputs = isoneutral_diffusion_pre_gpu(*inputs)
+            torch.cuda.synchronize()
```
wconstab, line 210: hmm this is good but we also do it again in the benchmark wrapper so i'm not sure if it adds extra overhead we care about.

wconstab (approved):
awesome! props for finding a cool and different use case.

from my quick skimming, this is basically a numerical simulator for ocean mixing, and so 'training' is not even a relevant concept- i assume that's the reason for you not implementing it?

xuzhao9 (approved):
LGTM

torchbenchmark/models/pyhpc_isoneutral_mixing/isoneutral_pytorch.py
```diff
@@ -0,0 +1,213 @@
...
+
+    K_33[2:-2, 2:-2, :-1] = sumx / (4 * dxt[2:-2, None, None]) + \
+        sumy / (4 * dyt[None, 2:-2, None]
+                * cost[None, 2:-2, None])
+    K_33[2:-2, 2:-2, -1] = 0.
+
+    return K_11, K_22, K_33, Ai_ez, Ai_nz, Ai_bx, Ai_by
+
+
+def prepare_inputs(*inputs, device):
```
Krovatkin, line 199: is this used anywhere?

Krovatkin (approved):
:shipit:

bertmaher:
> awesome! props for finding a cool and different use case.
> 
> from my quick skimming, this is basically a numerical simulator for ocean mixing, and so 'training' is not even a relevant concept- i assume that's the reason for you not implementing it?

Yep, the benchmark suite I'm pulling this from is just using PT for numerical calculations, there's no training stage.

torchbenchmark/models/pyhpc_isoneutral_mixing/isoneutral_pytorch.py
bertmaher: No, I copied this file verbatim from the pyhpc benchmark harness.  I can delete this for clarity.

torchbenchmark/models/pyhpc_isoneutral_mixing/isoneutral_pytorch.py
bertmaher: I don't think this path is actually run, so I'll delete it (and `prepare_inputs` as above)

PR #339, opened 2021-04-06
Wire callgrind CI into TorchBench harness

robieta:
Scripts to actually run https://github.com/pytorch/pytorch/pull/55428


robieta:
I'm not sure how to test this. Do I just SSH into a machine?

.github/scripts/instr-count-benchmark/create-nightly-env.sh
```diff
@@ -10,6 +10,11 @@ set +a;
...
 . activate ${CONDA_ENV_NAME}
 
+# We need Valgrind to collect instructions.
+conda install -y valgrind -c conda-forge
+
 # Install PyTorch nightly from pip
 pip install --pre torch \
     -f https://download.pytorch.org/whl/nightly/${CUDA_VERSION}/torch_nightly.html
+
+git clone https://github.com/pytorch/pytorch.git ${REPO_CHECKOUT}
```
xuzhao9, line 20: If historical data is not required, maybe use `--depth 1` to only checkout the latest master commit as an optimization?

.github/scripts/instr-count-benchmark/run-benchmark.sh
```diff
@@ -13,3 +13,6 @@ PYTORCH_GIT_VERSION=$(python -c "import torch; print(torch.version.git_version)"
 echo "Running instruction benchmark for pytorch-${PYTORCH_VERSION}, commit SHA: ${PYTORCH_GIT_VERSION}"
 
 # Run the instruction count benchmark
+pushd ${REPO_CHECKOUT}
+python main.py --mode ci --destination ${RESULT_JSON}
```
xuzhao9, line 17: Assuming `REPO_CHECKOUT` is the pytorch repository base, I think we should run `benchmarks/instruction_counts/main.py`?

.github/scripts/instr-count-benchmark/create-nightly-env.sh
```diff
@@ -10,6 +10,11 @@ set +a;
 conda create -y -q --name ${CONDA_ENV_NAME} python=${PYTHON_VERSION}
 . activate ${CONDA_ENV_NAME}
 
+# We need Valgrind to collect instructions.
+conda install -y valgrind -c conda-forge
```
xuzhao9, line 14: It's nice to know valgrind is in the conda-forge repo!

.github/scripts/instr-count-benchmark/upload.py
```diff
@@ -0,0 +1,121 @@
...
+            ],
+            'normal': [
+                # Unique ID for the run.
+                'run_id',
+
+                # Metadata for ad-hoc analysis.
+                'task_key', 'counts_json',
+
+                # TODO: More environment info.
+                'pytorch_version',
```
xuzhao9, line 44: We should also record the cuda version here.

.github/scripts/instr-count-benchmark/upload.py
```diff
@@ -0,0 +1,121 @@
...
+
+                # We don't need to record as many summary statistics for counts
+                # as we do for times, because we retain them all in
+                # `counts_json`.
+                'count_min', 'count_max',
+                'count_p25', 'count_median', 'count_p75',
+            ],
+            'normal': [
+                # Unique ID for the run.
+                'run_id',
```
xuzhao9, line 38: We could use the github actions workflow id as the run id. It will be set to the environment variable `GITHUB_RUN_ID`.

.github/scripts/instr-count-benchmark/upload.py
```diff
@@ -0,0 +1,121 @@
...
+        for key in result_json["values"].keys():
+            md5.update(key.encode("utf-8"))
+        assert result_json['md5'] == md5.hexdigest(), \
+            f"Data reports version MD5 of {result_json['md5']}, however computed {md5.hexdigest()}"
+
+        base_msg = {
+            'time': int(time.time()),
+            'start_time': result_json['start_time'],
+            'end_time': result_json['end_time'],
+            'benchmark_version': result_json['version'],
```
xuzhao9, line 73: Is "benchmark_version" the version of torchbench, or the version of instruction count benchmark? (I guess it is the latter, just wanna confirm). Also, I suggest adding another field called "benchmark_name" and set it to "instruction_count" to make it easier to filter all instruction count data in Scuba.

xuzhao9 (approved):
Looks good overall! Looking forward to the CI result.

xuzhao9:
> I'm not sure how to test this. Do I just SSH into a machine?

You may test the workflow by manually triggering it here: https://github.com/pytorch/benchmark/actions/workflows/instruction-count.yml. The log should be available in the GH Actions Web UI.

.github/scripts/instr-count-benchmark/config.env
```diff
@@ -3,3 +3,7 @@
 PYTHON_VERSION=3.7
 CUDA_VERSION=cu102
 CONDA_ENV_NAME="instr_cnt_benchmark"
+
+RESULT_JSON="${BASEDIR}/ubenchmark_results.json"
```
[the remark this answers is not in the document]
robieta: Good point. I was relying on each script having `BASEDIR=$(dirname $0)`, but that's not a great pattern. I switched it to `INSTRUCTION_COUNT_ROOT="$HOME/.torchbench/instruction-count-nightly-ci"` and keyed everything else on that. Thanks for the suggestion.

.github/scripts/instr-count-benchmark/config.env
```diff
@@ -3,3 +3,7 @@
...
+REPO_CHECKOUT="${BASEDIR}/pytorch"
+BENCHMARK_ROOT="${CHECKOUT}/benchmarks/instruction_counts"
```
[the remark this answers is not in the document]
robieta: Oops. That was supposed to be `REPO_CHECKOUT`. Hence why I'm keen to have a way to test it...

.github/scripts/instr-count-benchmark/upload.py
```diff
@@ -0,0 +1,121 @@
...
+import sys
+import time
+
+import numpy as np
+import torch
+
+from scripts.upload_scribe import ScribeUploader
+
+
+VERSIONS = [
```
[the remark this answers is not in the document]
robieta: It's to make sure this stays in sync with the benchmark definitions. (Which are in a separate repo.) I added a comment to that effect.

.github/scripts/instr-count-benchmark/create-nightly-env.sh
robieta: Done.

.github/scripts/instr-count-benchmark/run-benchmark.sh
robieta: They expect to be run from `benchmarks/instruction_counts`. (It affects import paths. e.g. `from core.api` rather than `from benchmarks.instruction_counts.core.api`)

.github/scripts/instr-count-benchmark/create-nightly-env.sh
robieta: Indeed!

.github/scripts/instr-count-benchmark/upload.py
robieta: Done.

.github/scripts/instr-count-benchmark/upload.py
robieta: Ohh, that's much better. Thanks!

.github/scripts/instr-count-benchmark/upload.py
robieta: Added a comment and a `benchmark_name` field. (These will live in a separate Scuba table so we don't have to worry about clashing with TorchBench, but still good to be explicit.)

.github/scripts/instr-count-benchmark/create-nightly-env.sh
```diff
@@ -7,9 +7,16 @@ set -a;
...
+mkdir -p ${INSTRUCTION_COUNT_ROOT}
+
 conda create -y -q --name ${CONDA_ENV_NAME} python=${PYTHON_VERSION}
 . activate ${CONDA_ENV_NAME}
 
+# We need Valgrind to collect instructions.
+conda install -y valgrind -c conda-forge
+
 # Install PyTorch nightly from pip
 pip install --pre torch \
```
xuzhao9, line 19: Sometimes pytorch nightly will fail to build, and we should cancel CI. Please add a check similar to here: https://github.com/pytorch/benchmark/blob/0.1/.github/scripts/run-nightly-nodocker.sh#L33
```suggestion
# Check if nightly builds are available
NIGHTLIES=$(python torchbenchmark/util/torch_nightly.py --packages torch)
# If failed, the script will generate empty result
if [ -z $NIGHTLIES ]; then
    echo "Torch, torchvision, or torchtext nightly build failed. Cancel the workflow."
    exit 1
fi
pip install --pre torch \
```

.github/scripts/instr-count-benchmark/run-benchmark.sh
xuzhao9: But then shouldn't we first enter the `${BENCHMARK_ROOT}` directory?
```suggestion
python main.py --mode ci --destination ${BENCHMARK_ROOT}
```

.github/scripts/instr-count-benchmark/run-benchmark.sh
```diff
@@ -13,3 +13,6 @@ PYTORCH_GIT_VERSION=$(python -c "import torch; print(torch.version.git_version)"
 echo "Running instruction benchmark for pytorch-${PYTORCH_VERSION}, commit SHA: ${PYTORCH_GIT_VERSION}"
 
 # Run the instruction count benchmark
+pushd ${REPO_CHECKOUT}
```
xuzhao9, line 16: But then shouldn't we first enter the directory?
```suggestion
pushd ${BENCHMARK_ROOT}
```

.github/scripts/instr-count-benchmark/create-nightly-env.sh
robieta: Thanks for the tip.

.github/scripts/instr-count-benchmark/run-benchmark.sh
robieta: Ah! You are correct.

robieta:
I didn't realize I could test branches until too late. Sorry. (It should only have broken the instruction count workflow, so I'm going to just fix forward.)

robieta:
Actually, it looks like even if I select a different branch it's running `master`. (Presumably because of `instruction-count.yml`?) @xuzhao9, could you advise?

xuzhao9:
> Actually, it looks like even if I select a different branch it's running `master`. (Presumably because of `instruction-count.yml`?) @xuzhao9, could you advise?

Right, currently the `instruction-count.yml` only runs on the master branch (https://github.com/pytorch/benchmark/blob/master/.github/workflows/instruction-count.yml#L2). If you would like the CI to run on another branch, you could submit PR to `instrunction-count.yml` and checkout the branch you would like to work on, similar to [here](https://github.com/pytorch/benchmark/blob/master/.github/workflows/main.yml#L19), where branch "0.1" is used.

PR #344, opened 2021-04-07
[hackathon] Add differentially-private cifar10 training

### LINKED ISSUE #340: Add opacus differentially-private cifar10 model
This is a proposal to add a model that does differentially-private training.

The specific model in question is the following:
https://github.com/pytorch/opacus/blob/master/examples/cifar10.py

zou3519:
Fixes #340.

This PR adds an example of differentially-private (DP) training on
the cifar10 dataset. In reality this uses resnet18 as the model and
example inputs with the size of those found in the cifar10 dataset.

The model is not tracable or scriptable. This is because the DP
training relies on autograd hooks and those cannot be traced through or
scripted yet.

Test Plan:

All of the following pass, is there anywhere else I should add a test?
- `python test.py -k test_opacus_cifar10_example_cpu`
- `python test.py -k test_opacus_cifar10_eval_cpu`
- `python test.py -k test_opacus_cifar10_train_cpu`


torchbenchmark/models/opacus_cifar10/__init__.py
```diff
@@ -0,0 +1,81 @@
...
+
+        optimizer = optim.Adam(model.parameters(), lr=0.001)
+        criterion = nn.CrossEntropyLoss()
+
+        # This is supposed to equal the number of data point.
+        # It is only to compute stats so dwai about the value.
+        sample_size = 64 * 100
+        clipping = {"clip_per_layer": False, "enable_stat": False}
+
+        privacy_engine = PrivacyEngine(
```
wconstab, line 47: is constructing the PrivacyEngine something you need to do every step?  I would suggest moving this part to __init__ and focusing 'train' on just the `for _  in range` part.

torchbenchmark/models/opacus_cifar10/__init__.py
```diff
@@ -0,0 +1,81 @@
...
+        )
+        self.example_target = torch.randint(0, 10, (64,), device=self.device)
+
+    def get_module(self):
+        if self.jit:
+            raise NotImplementedError()
+
+        return self.model, self.example_inputs
+
+    def train(self, niter=3):
```
wconstab, line 32: any reason you picked niter=3? 

in practice we've tended to not use niter, since we wrapped it in pytest which does figure out a number of iterations based on how long each takes.

wconstab:
@zou3519 you can run `pytest test_bench.py` with the similar -k filters and get a timing report.  as long as that also runs, it should be good.

We've had a breakage on CI master this week, but it's being worked on and we should be able to merge stuff again soon.

torchbenchmark/models/opacus_cifar10/__init__.py
zou3519: Copied it from https://github.com/pytorch/benchmark/blob/3015d62c6c8825521c45cbaf89cc498742bee65b/torchbenchmark/models/resnet18/__init__.py#L23. I'll delete it, having pytest figure out iterations sounds good

torchbenchmark/models/opacus_cifar10/__init__.py
zou3519: Hmm, privacy engine should just be constructed once. It needs to accept `optim` as an argument. Is it fine for the optimizer construction to also be in `__init__`?

wconstab: yea, the more the better (moved to __init__).  the stuff that should be in train() is typically just calling forward, backward, step.

zou3519:
Updated the PR as per comments. Could you take another look when you get a chance please, @wconstab ?

torchbenchmark/models/opacus_cifar10/__init__.py
```diff
@@ -0,0 +1,78 @@
...
+import torch.nn as nn
+import torchvision.models as models
+from opacus.utils.module_modification import convert_batchnorm_modules
+from opacus import PrivacyEngine
+
+from ...util.model import BenchmarkModel
+
+
+class Model(BenchmarkModel):
+    def __init__(self, device=None, jit=False):
```
zou3519, line 12: @xuzhao9 This model is differentially private training of an image classification problem. I think I should assign this the OTHER.OTHER_TASKS task due to the category being "differential privacy", what do you think?

xuzhao9: Yes, please set it to `OTHER.OTHER_TASKS`. We may have a better categorization later, but setting it to OTHER is good for now.

torchbenchmark/models/opacus_cifar10/__init__.py
```diff
@@ -0,0 +1,78 @@
...
+        loss.backward()
+        self.optimizer.step()
+        self.optimizer.zero_grad()
+
+    def eval(self):
+        if self.jit:
+            raise NotImplementedError()
+
+        model, (images,) = self.get_module()
+        targets = self.example_target
```
xuzhao9, line 69: Please set the model to eval() mode and disable auto_grad() in eval().

xuzhao9 (approved):
Overall LGTM. Please address the comments.

torchbenchmark/models/opacus_cifar10/__init__.py
zou3519: Resolved

torchbenchmark/models/opacus_cifar10/__init__.py
zou3519: Done

PR #348, opened 2021-04-07
Add a SSD model

desertfire:
Originated from https://github.com/sgrvinod/a-PyTorch-Tut……orial-to-Object-Detection.git

Test Plan:
```
python test.py -k "test_ssd_train_cuda"
python test.py -k "test_ssd_eval_cuda"
python test.py -k "test_ssd_train_cpu"
python test.py -k "test_ssd_eval_cpu"
```


Chillee:
Perhaps this implementation of SSD would be better? https://pytorch.org/hub/nvidia_deeplearningexamples_ssd/

wconstab:
i'm wondering if it'd make sense to move the dataset to a shared folder (torchbenchmark/datasets) given that it's a lot of files and may be used by other benchmarks.  Do we have any other benchmarks using VOC dataset yet?

desertfire:
> Is there any reason you chose that implementation and not say, this one? https://github.com/amdegroot/ssd.pytorch.
> Or the one on torch hub?

* Initially I tried https://github.com/amdegroot/ssd.pytorch, but it failed because its intermediate step uses a pre-trained vgg16_reducedfc.pth which is not compatible with the latest pytorch.
* https://pytorch.org/hub/nvidia_deeplearningexamples_ssd/, for this torch hub one, I was afraid it is highly optimized for CUDA which makes it not a good candidate for general benchmarking. Please let me know if it is not the case.

desertfire:
> i'm wondering if it'd make sense to move the dataset to a shared folder (torchbenchmark/datasets) given that it's a lot of files and may be used by other benchmarks. Do we have any other benchmarks using VOC dataset yet?

I do see COCO lives in that folder. Not sure if any other benchmark needs VOC at the moment, but I can move the directory.

wconstab (changes requested):
it looks like too much initialization stuff is going on inside your BenchmarkModel.train() function.  The BenchmarkModel.__init__ should take care of stuff like setting up dataloader, building the model, etc- and train() should just call model.forward/loss.backward/optimizer.step ideally.

Issue #328, opened 2021-03-23
Replace MAML with an actually correct implementation


Chillee:
https://github.com/dragen1860/MAML-Pytorch/issues/59

xuzhao9:
Closed, we keep this "incorrect" model implementation because it is useful for dynamo correctness testing.

PR #349, opened 2021-04-07
Add example of maml training on omniglot

zou3519:
This is related to #328. This PR adds an actually correct
implementation of maml to the repo. The previous implementation doesn't
actually compute higher order gradients where it is supposed to.

I'm not familiar with how torchbench works so please let me know if
there are additional files that need to be modified.

Test Plan:

Ran the following:
```
python test.py -k test_maml_omniglot_example_cpu
python test.py -k test_maml_omniglot_eval_cpu
python test.py -k test_maml_omniglot_train_cpu
```

Future work:
- Delete the maml example that is currently in this repo (or rename it
to make it clear that it's doing something different from the paper that
it is trying to reproduce).


Krovatkin (dismissed):
:shipit: this is probably the highest density of optimizers per 10 LOCs, I've seen O_o

zou3519:
@Krovatkin It's been a while since you initially reviewed this PR. I rebased on top of the master branch and the build+tests now fully succeed. Could I ask you to take another look at this please before I hit the merge button?

torchbenchmark/models/maml_omniglot/__init__.py
```diff
@@ -0,0 +1,110 @@
...
+import torch.nn.functional as F
+from pathlib import Path
+import higher
+
+from ...util.model import BenchmarkModel
+
+
+class Model(BenchmarkModel):
+    def __init__(self, device=None, jit=False):
+        super().__init__()
```
xuzhao9, line 31: Please specify a task to the model: https://github.com/pytorch/benchmark/blob/master/torchbenchmark/tasks.py.

torchbenchmark/models/maml_omniglot/__init__.py
```diff
@@ -0,0 +1,110 @@
...
+                    qry_loss = F.cross_entropy(qry_logits, y_qry[i])
+                    qry_loss.backward()
+
+            meta_opt.step()
+
+    def eval(self, niter=1):
+        if self.jit:
+            raise NotImplementedError()
+
+        model, (example_input,) = self.get_module()
```
xuzhao9, line 100: Do we need to set model.eval() or "with torch.no_grad()" in eval()?

torchbenchmark/models/maml_omniglot/__init__.py
zou3519: This model is metalearning over classification on an image dataset. I think this is either OTHER.OTHER_TASKS or COMPUTER_VISION.OTHER_COMPUTER_VISION. Do you have advice for which category this should go @xuzhao9? (Otherwise I'll pick OTHER.OTHER_TASKS)

xuzhao9: > This model is metalearning over classification on an image dataset. I think this is either OTHER.OTHER_TASKS or COMPUTER_VISION.OTHER_COMPUTER_VISION. Do you have advice for which category this should go @xuzhao9? (Otherwise I'll pick OTHER.OTHER_TASKS)

Since it is meta learning, please add it to OTHER.OTHER_TASKS, thanks!

xuzhao9:
Please help resolve the comments and I will be happy to merge your contribution!

torchbenchmark/models/maml_omniglot/__init__.py
zou3519: Added eval() and no_grad

torchbenchmark/models/maml_omniglot/__init__.py
zou3519: Added OTHER.OTHER_TASKS.

zou3519:
@xuzhao9 I resolved the comments, could you please take another look when you get a chance?

xuzhao9 (approved):
LGTM! Thanks for the contribution!

Issue #342, opened 2021-04-07
Add a bunch of timm models


wconstab:
https://github.com/rwightman/pytorch-image-models

pip install a specific version
then pick one or more models from the library based on popularity and interest, and wrap them in __init__

xuzhao9:
Created https://github.com/pytorch/benchmark/pull/350 to add NFNet-F0 (DeepMind compatible) model to TorchBench. It should support other variants of the nfnet models as well.

xuzhao9:
PR that adds a bunch of timm models: https://github.com/pytorch/benchmark/pull/352

xuzhao9:
Closed by merging https://github.com/pytorch/benchmark/pull/352. Feel free to reopen it if people need more timm models.

PR #350, opened 2021-04-07
Added the NFNet-F0 (DeepMind weight compatible) model to TorchBench.

xuzhao9:
Addresses https://github.com/pytorch/benchmark/issues/342
Test plan:
Run the model locally with the following command:

`python -m torchbenchmark.models.timm_nfnet.__init__`

Support both [CPU, CUDA], [JIT, eager] modes.


Krovatkin (commented):
@xuzhao9 could we consider adding the model stuff as a submodule or via a pip package? 
Also could we move configuration objects from  `__init__.py` into a separate file and have a smaller construction code in  `__init__.py`

xuzhao9:
> @xuzhao9 could we consider adding the model stuff as a submodule or via a pip package?
> Also could we move configuration objects from `__init__.py` into a separate file and have a smaller construction code in `__init__.py`

Thanks for the feedback! I have addressed your comments.

torchbenchmark/models/timm_nfnet/__init__.py
```diff
@@ -0,0 +1,59 @@
...
+class Model(BenchmarkModel):
+    task = COMPUTER_VISION.CLASSIFICATION
+
+    def __init__(self, device=None, jit=False, variant='dm_nfnet_f0', precision='float32'):
+        super().__init__()
+        self.device = device
+        self.jit = jit
+        self.model = timm.create_model(variant, pretrained=False, scriptable=True)
+        self.cfg = NFNetConfig(model = self.model, device = device, precision = precision)
+        self.model.train()
```
wconstab, line 17: what's the purpose of calling train() and then eval() here? 

note that the model does need to be set in 'train' or 'eval' mode at the right time (Before calling training or eval functions).  See this api:
https://github.com/pytorch/benchmark/blob/master/torchbenchmark/util/model.py#L47

you probably want to override it.  

its default impl actually would work for you by default, except you did not implement get_module().  That is supposed to be a required API, but for some reason, it is not included in the 'BenchmarkModel' base class.  looks like that was an oversight.  I'd suggest implementing get_module() and then you don't have to override set_mode.

wconstab (approved):

torchbenchmark/models/timm_nfnet/requirements.txt
```diff
@@ -0,0 +1 @@
+timm == 0.4.5
```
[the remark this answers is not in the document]
xuzhao9: Maybe, just want to add that the runner could easily run another model by specifying the "variant" variable. Available of variant values are [here](https://github.com/rwightman/pytorch-image-models/blob/master/timm/models/nfnet.py#L46)

torchbenchmark/models/timm_nfnet/__init__.py
xuzhao9: I borrowed this part of code from the timm benchmarking code [train](https://github.com/rwightman/pytorch-image-models/blob/master/benchmark.py#L248) and [inference](https://github.com/rwightman/pytorch-image-models/blob/master/benchmark.py#L193). You are right, I should remove this.

However, I am not very clear about the `get_module()` api's semantic. What is it supposed to do? For example, can you explain [the code here](https://github.com/pytorch/benchmark/blob/master/torchbenchmark/models/BERT_pytorch/__init__.py#L101)?

xuzhao9: Also, I have updated the code to use get_module(), please help check whether it looks correct. Thanks!

Chillee:
For `timm` models, I think it might make sense to do something similar to this (https://github.com/pytorch/benchmark/pull/351) and codegen a whole bunch of them.

xuzhao9:
> For `timm` models, I think it might make sense to do something similar to this (#351) and codegen a whole bunch of them. In addition to NFNets, I think it's worth adding ViT, EfficientNets, and potentially RegNet.

Based on my understanding, the timm models already contains a `create_model()` [method](https://github.com/rwightman/pytorch-image-models/blob/a2727c1bf78ba0d7b5727f5f95e37fb7f8866b1f/timm/models/factory.py#L25) that is used to generate different models (such as nfnet, efficientnet, and others). So after installing the timm package, we should already be able to codegen all timm models.

xuzhao9:
> For `timm` models, I think it might make sense to do something similar to this (#351) and codegen a whole bunch of them. In addition to NFNets, I think it's worth adding ViT, EfficientNets, and potentially RegNet.

Ah I see what you are talking about. I am working on another PR.

xuzhao9:
Closed and superseded by https://github.com/pytorch/benchmark/pull/352

PR #351, opened 2021-04-07
Added a bunch of transformers from HuggingFace

Chillee:
https://huggingface.co/transformers/model_doc/t5.html#training


torchbenchmark/models/hf_T5/__init__.py
```diff
@@ -0,0 +1,56 @@
...
+        input_ids = torch.randint(0, 5000, (batch_size, seq_length)).to(device)
+        decoder_ids = torch.randint(0, 5000, (batch_size, seq_length)).to(device)
+
+        self.model = T5ForConditionalGeneration.from_pretrained('t5-small').to(device)
+        self.train_inputs = {'input_ids': input_ids, 'labels': decoder_ids}
+        self.eval_inputs = {k: v[0:1] for k,v in self.train_inputs.items()}
+
+    def get_module(self):
+        if self.jit:
+            raise NotImplementedError()
```
Krovatkin, line 29: not even tracing 😿

torchbenchmark/models/hf_T5/__init__.py
```diff
@@ -0,0 +1,56 @@
...
+        self.device = device
+        self.jit = jit
+
+        batch_size = 8
+        seq_length = 512
+
+        input_ids = torch.randint(0, 5000, (batch_size, seq_length)).to(device)
+        decoder_ids = torch.randint(0, 5000, (batch_size, seq_length)).to(device)
+
+        self.model = T5ForConditionalGeneration.from_pretrained('t5-small').to(device)
```
Krovatkin, line 23: I don't know how small is `t5-small` we could consider checking in the pth file to pin it.

torchbenchmark/models/hf_T5/__init__.py
```diff
@@ -0,0 +1,56 @@
...
+
+    def get_module(self):
+        if self.jit:
+            raise NotImplementedError()
+        return self.model, self.eval_inputs
+
+    def train(self, niter=3):
+        if self.jit:
+            raise NotImplementedError()
+        optimizer = optim.Adam(self.model.parameters())
```
Krovatkin, line 35: this is beyond the scope of this PR, but @wconstab we should consider outlining constructors in `pre_train` or something. I'm not sure if we really want to measuring how fast we can construct an optimizer object.

Krovatkin (approved):
:shipit:

torchbenchmark/models/hf_T5/__init__.py
Chillee: It's a couple hundred megabytes I think.

torchbenchmark/models/hf_T5/__init__.py
Chillee: that's a good point - I'll move it to the init.

torchbenchmark/models/gen_hf_generative.py
```diff
@@ -0,0 +1,109 @@
...
+        self.jit = jit
+
+
+        config = {config}
+        self.model = {model}(config).to(device)
+        self.optimizer = optim.Adam(self.model.parameters(), lr=0.001)
+
+        num_tokens = 5
+
+        input_ids = torch.randint(0, config.vocab_size, {input_shape}).to(device)
```
wconstab, line 42: i'm curious about your motivation here.  if you're wanting to make the model exercise different codepaths run to run, this would come at the expense of less deterministic benchmarking.  I'm not sure how much impact the input data has on either, but it would have that impact on both.  My gut would be to just fix the inputs.

Chillee: well, I think it's common for us to use `randn` or `randint` for generating data.

I was quite worried about this, but luckily, I don't think any of the models I'm adding have *very* input-dependent runtimes.

The only exception is Reformer, which has a sort embedded within it. However, I think that sorting procedure's runtime isn't very input dependent, luckily.

torchbenchmark/models/hf_T5/__init__.py
stas00: You can see the size of each model by going to its files tab on the hub:
https://huggingface.co/t5-small/tree/main

so 230MB for t5-small

wconstab: wait so this 230MB is downloaded on setup? or in the repo? either way it sounds non-ideal.  do we need the real parameters?

torchbenchmark/models/gen_hf_generative.py
wconstab: ok. so my objection isn't to using randint, it's to using randint _on every startup_.  Instead, why not have your gen_ script call randint for you and hardcode the tensor in a pkl?  that way at least there wouldn't be run-to-run variation.

wconstab (approved):
I'd prefer to change the randint thing but otherwise LGTM. thanks for pulling this together!

torchbenchmark/models/hf_T5/__init__.py
Chillee: I've since changed it to not use the real parameters.

stas00: If you just want to get the config, it's `T5Config.from_pretrained("t5-small")` and then you can tweak it and then create a model, but not pretrained, that is:
```
config = T5Config.from_pretrained("t5-small")
# edit config if need be
# now create a new untrained model
model = T5ForConditionalGeneration(config)
```

You can use `AutoConfig` and `AutoModel` everywhere and it'll automatically figure out the right class to use internally.

stas00: so no download will happen.

and also if you run on the same machine the models are cached.

torchbenchmark/models/gen_hf_generative.py
Krovatkin: I think I share @wconstab's concern about a possible jitter, but I was also wondering if we could do some kind of a middle road solution here. We could still generate random inputs, we could just try generating them in a *consistent* way like so w/o affecting the global state.

```python
       # freeze the global rnd state
       with freeze_rng_state():
            torch.manual_seed(42) # set a new state 
            decoder_ids = torch.randint(0, config.vocab_size, {input_shape})
```

torchbenchmark/models/hf_Albert/__init__.py
```diff
@@ -0,0 +1,75 @@
+
+# Generated by gen_hf_generative.py
```
Krovatkin, line 2: nitpick: can we make this comment visually stand out somehow? Otherwise, inevitably, we'll start editing individual models.

torchbenchmark/models/hf_GPT2/requirements.txt
```diff
@@ -0,0 +1,3 @@
+transformers==4.5.0
+datasets
```
Krovatkin, line 2: is this package,  `datasets`  used? We seem to be generating all the inputs ourselves?

Krovatkin (approved):
:shipit:

torchbenchmark/models/gen_hf_generative.py
Chillee: If we're doing this, we should also make sure the model's parameters are generated consistently as well.

Actually, why don't we just enforce this at the benchmark level? Before you run `__init__`, just reset all of your random seeds.

torchbenchmark/models/hf_GPT2/requirements.txt
Chillee: not totally sure... these HF models do some weird imports internally. I didn't think we needed SentencePiece either, but the BigBird model imports it somewhere internally.

torchbenchmark/models/gen_hf_generative.py
Krovatkin: may not be a bad idea, actually. If we do it consistently, for each test.

wconstab: actually.. iirc we do this in the template, and (maybe?) suggest it in the adding models?  should make that more explicit and somehow enforce.  maybe just rely on doing it in the benchmark driver before each test instead?

wconstab: e.g. here:
https://github.com/pytorch/benchmark/blob/master/torchbenchmark/models/Background_Matting/__init__.py#L20

checked, adding_models is out of date. i'll PR to update now. missing some other best practices.. i'll ask you two to weigh in on the PR to get more stuff covered, i'm sure i'll miss something.

PR #352, opened 2021-04-08
Add a bunch of timm models.

xuzhao9:
Added six timm models, which are all in the computer vision category.

- Use RegnetY variant for Regnet
- Use MixNet variant for efficientnet
- Added ResNest model
- Use batch_size=64 as default for training,  use batch_size=1 as default for inference

Test plan:
Generate all the timm models by running:
```
python torchbenchmark/models/gen_timm_models.py
```

Run all the timm model locally with the following command:

```
python -m torchbenchmark.models.timm_nfnet.__init__
python -m torchbenchmark.models.timm_efficientnet.__init__
python -m torchbenchmark.models.timm_regnet.__init__
python -m torchbenchmark.models.timm_resnest.__init__
python -m torchbenchmark.models.timm_vovnet.__init__
python -m torchbenchmark.models.timm_vision_transformer.__init__
```


torchbenchmark/models/timm_efficientnet/config.py
```diff
@@ -0,0 +1,48 @@
...
+    def _init_input(self):
+        self.example_inputs = torch.randn(
+            (self.batch_size,) + self.input_size, device=self.device, dtype=self.data_dtype)
+
+    def __init__(self, model, device, precision):
+        self.model = model
+        self.device = device
+        self.use_amp, self.model_dtype, self.data_dtype = resolve_precision(precision)
+        # Configurations
+        self.batch_size = 2
```
wconstab, line 39: why use bs=2? i'm not familiar with the particular models, but it's generally important to pick batch sizes that make sense given the real use case.  

it's also common to use different batch sizes for training vs inference, and possibly different based on cpu/gpu.   we haven't done a good job across the suite with this.  

i wonder if you could implement a different batch size for train vs eval, based on what is common for that model in the original repo?

Chillee:
Some other models you might want to consider adding, from discussion with Ross (although perhaps we have enough vision models), are:
1. RegnetY instead of RegnetX (regnetY is more popular).
2. MixNet (has some unusual conv sizes, like 9x9 kernels)
3. ResNest (another model with attention that has some interesting properties).

torchbenchmark/models/timm_efficientnet/config.py
xuzhao9: The original repo uses batch_size=256 for both train and inference: https://github.com/rwightman/pytorch-image-models/blob/master/benchmark.py#L64. I think it would be too large for torchbench. Is there any numbers I can use for reference, e.g. in other models?

Chillee: I think it would be reasonable to choose the largest (power of 2) batch size you could fit on the GPU, and do a batch size of 1 for inference.

xuzhao9: > I think it would be reasonable to choose the largest (power of 2) batch size you could fit on the GPU, and do a batch size of 1 for inference.

Sorry but I am not very familiar with ML. How to estimate the largest batch size to fit on the GPU? The CI machine uses a Telsa T4 GPU with 16GB DDR6 memory.

wconstab: haha i don't think there is much to it,  just try 256, 128, ... until it doesn't give OOM error :)

xuzhao9: > haha i don't think there is much to it, just try 256, 128, ... until it doesn't give OOM error :)

I see. On the CI machine the highest available batch_size without OOM error is 64, so I am using 64 for train and 1 for inference.

bitfort (commented):
This is simply awesome. LOVE to see this many models being added at the same time!

torchbenchmark/models/gen_timm_models.py
```diff
@@ -0,0 +1,149 @@
...
+        self.model = timm.create_model(variant, pretrained=False, scriptable=True)
+        self.cfg = TimmConfig(model = self.model, device = device, precision = precision)
+        self.model.to(
+            device=self.device,
+            dtype=self.cfg.model_dtype
+        )
+        if device == 'cuda':
+            torch.cuda.empty_cache()
+        if jit:
+            self.model = torch.jit.script(self.model)
```
Chillee, line 38: Do you know how many of these models actually script?

xuzhao9: Sorry I don't. How to verify that the model actually scripts?

xuzhao9: The [benchmarking code](https://github.com/rwightman/pytorch-image-models/blob/master/benchmark.py#L166) from timms package shows all models created in the repository should be able to jit.

wconstab: if the line of code does not crash, then it scripts. but you might want to make sure it's actually hitting that line in one case if you're not sure.  (you can also confirm that self.model is a 'scriptmodule' object rather than a 'module' object.

xuzhao9: Makes sense, added the jit script check.

wconstab: oh- to be clear, the typecheck isn't necessary.  the only reason i suggested that is if you wanted to convince yourself in the debugger.  as long as you know the line of code is reached, it should be fine without the additional check.  (doesn't hurt anything to have it either though)

bitfort (approved):
Thanks Xu; great work here.

PR #354, opened 2021-04-09
[WIP] Configure fastNLP model

ansley:


torchbenchmark/models/fastNLP/fastNLP/models/bert.py
```diff
@@ -67,7 +67,7 @@ def forward(self, words: Tensor, offsets: Tensor) -> Tensor:
         :param torch.LongTensor words: [batch_size, seq_len]
         :return: { :attr:`fastNLP.Const.OUTPUT` : logits}: torch.Tensor [batch_size, num_labels]
         """
-        hidden = self.dropout(self.bert(words, offsets))
+        hidden = self.dropout(self.bert(words), offsets)
```
wconstab, line 70: i suspect this part isn't right, but i am not sure how offsets is used.  it'd be a little odd to pass a tuple of things to dropout.  i think offsets either should be used by bert, or removed entirely (e.g. the way it is on the trunk fastnlp repo).

Chillee: Does the current code run?

ansley: No, it doesn't. I need to do more to revert the EmbeddingBag change. This hasn't been one of my priorities since it was just a hackathon task. If it's mission critical, someone else should take it over!

wconstab: @Chillee I think offsets was added by the person who torchscripted this model and used embeddingbag instead of bert.  The challenging part for @ansley is to figure out the correct way to remove offsets and replace this with a real bert.  if you happen to be familiar with those APIs maybe you can show her the change that needs to be made, i think it's probably simple.

xuzhao9:
Closing this PR since the fastNLP model is not replaced by fastNLP_Bert.

PR #357, opened 2021-04-12
Fix Circle CI PR correctness check.

xuzhao9:
The PR ci meets a bunch of CUDA errors because it uses an incompatible version of Nvidia driver.
I tried to install the correct Nvidia driver, but the problem persists under CUDA 11.1.
After upgrading the version to CUDA 11.2, the problem disappears.

Also added other fixes:
- Use gpu.nvidia.medium runner, which has larger GPU memory
- Fixes the Huggingface model unittests by accepting kwargs in test.py

### DIFF
--- a/torchbenchmark/models/mobilenet_v2_quantized_qat/__init__.py
+++ b/torchbenchmark/models/mobilenet_v2_quantized_qat/__init__.py
@@ -14,7 +14,6 @@ class Model(BenchmarkModel):
         self.model = models.mobilenet_v2().to(self.device)
         self.example_inputs = (torch.randn((96, 3, 224, 224)).to(self.device),)
         self.prep_qat_train()  # config+prepare steps are required for both train and eval
-        torch.cuda.empty_cache()
 
     def prep_qat_train(self):
         qconfig_dict = {"": torch.quantization.get_default_qat_qconfig('fbgemm')}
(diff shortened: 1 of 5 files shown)


wconstab (approved):

torchbenchmark/models/mobilenet_v2_quantized_qat/__init__.py
```diff
@@ -14,6 +14,7 @@ def __init__(self, device=None, jit=False):
         self.model = models.mobilenet_v2().to(self.device)
         self.example_inputs = (torch.randn((96, 3, 224, 224)).to(self.device),)
         self.prep_qat_train()  # config+prepare steps are required for both train and eval
+        torch.cuda.empty_cache()
```
wconstab, line 17: hmm... makes me wonder why this helps?  i think if the cache can be emptied, the space would be reused automatically rather than throw oom.  maybe there is something wrong with our allocator?

xuzhao9: I was trying to use this to fix the OOM error but looks like it doesnot work. I am trying other methods to fix this.

xuzhao9: Should be fixed by using `gpu.nvidia.medium` runner.

Chillee:
Feels good having a passing master :)

PR #358, opened 2021-04-13
Remove warmups in correctness check and run each benchmark only once.

xuzhao9:


wconstab (approved):

wconstab (commented):
oh, _this_ was why I didnt disable before- because it helps check the score computes.

do you feel like we can get that coverage another way?

xuzhao9:
> oh, _this_ was why I didnt disable before- because it helps check the score computes.
> 
> do you feel like we can get that coverage another way?

If we are only checking v0 scores, we could use a json from the S3 bucket, such as https://s3.amazonaws.com/ossci-metrics/torchbench_v0_nightly/gh741259573/20210412_131220_1.json.

Krovatkin:
@xuzhao9 if we are going to use a score for anything, I think, we need to keep a warm up. Otherwise, we will be running every JIT test in a profiling mode. cc @wconstab

xuzhao9:
> @xuzhao9 if we are going to use a score for anything, I think, we need to keep a warm up. Otherwise, we will be running every JIT test in a profiling mode. cc @wconstab

In the correctness test, the score will only be used for testing the correctness of "compute_score.py".

PR #361, opened 2021-04-16
Various fixes to instruction count CI

robieta:
I was missing a bunch of minor things in the first pass:
* Need the top level `requirements.txt` for `torch_nightly.py`
* Need a bunch of extra conda installs for the ubenchmarks
* For some reason there are some files in `REPO_CHECKOUT`. (But not a full checkout.) So clean that up before cloning.
* I wasn't actually calling the Scribe upload script. Oops.
* I am missing `SCRIBE_GRAPHQL_ACCESS_TOKEN`

Right now the benchmark suite runs until it gets to `upload.py` and fails because of the token. So we're close!

### DIFF
--- a/.github/scripts/instr-count-benchmark/config.env
+++ b/.github/scripts/instr-count-benchmark/config.env
@@ -4,6 +4,7 @@ PYTHON_VERSION=3.7
 CUDA_VERSION=cu102
 CONDA_ENV_NAME="instr_cnt_benchmark"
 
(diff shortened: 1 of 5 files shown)


xuzhao9 (approved):
Nice work! LGTM. The secret "SCRIBE_GRAPHQL_ACCESS_TOKEN" should have already been set in the repository, so you should be ready to go :)

PR #363, opened 2021-04-22
Add bisection script and devbig local abtest scripts

xuzhao9:
This PR cherry-picks the bisector and devbig local abtest scripts from the 0.1 branch. So that people can local test models in the master branch.

Need to remove `import torch` from `__init__.py` because when running the bisector, pytorch is not compiled and installed yet.

### DIFF
--- a/.github/scripts/bisection-config.sample.yaml
+++ b/.github/scripts/bisection-config.sample.yaml
@@ -1,0 +1,15 @@
+# The sample bisection config that solves GH issue #51380
+
+# Start and end commits
+start: a87a1c1
+end: 0ead9d5
+# 10 percent regression
+threshold: 10
+# Support increase, decrease, or both
+# increase means performance regression, decrease means performance optimization
+direction: increase
+# Test timeout in minutes
+timeout: 60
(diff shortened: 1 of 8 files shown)


.github/scripts/bisection-config.sample.yaml
```diff
@@ -0,0 +1,15 @@
+# The sample bisection config that solves GH issue #51380
+
+# Start and end commits
+start: a87a1c1
+end: 0ead9d5
+# 10 percent regression
+threshold: 10
```
wconstab, line 7: what did you base this threshold on? is the 'noise level' about 5%?

wconstab (approved):

.github/scripts/bisection-config.sample.yaml
xuzhao9: Unfortunately, although the overall score doesn't change 1% for the same build, it is common that individual tests result changes 5% or even more. For example, in this [issue](https://github.com/pytorch/pytorch/issues/51384) where regression only happens on BackgroundMatting model, other models' performance could change up to 9% (e.g., demucs and pytorch_struct). Therefore, I am only confident it is a real signal with >10% delta.

PR #366, opened 2021-04-27
Add PyTorch PR CI support

xuzhao9:
This PR is to add TorchBench support in PyTorch CI.
It modifies bisection.py so that the bisector is also able to run abtests.
It also allows users to specify model names as test items. So model names like "yolov3" also works, not just "test_eval[yolov3-cpu-eager]".

This PR is the prerequisite of the TorchBench CI PR on PyTorch: https://github.com/pytorch/pytorch/pull/56957.

### DIFF
--- a/bisection.py
+++ b/bisection.py
@@ -147,8 +147,11 @@ class TorchSource:
             assert gitutils.update_git_repo(repo), f"Failed to update master branch of {repo}."
 
     # Get all commits between start and end, save them in self.commits
-    def init_commits(self, start: str, end: str) -> bool:
-        commits = gitutils.get_git_commits(self.srcpath, start, end)
+    def init_commits(self, start: str, end: str, abtest: bool) -> bool:
+        if not abtest:
+            commits = gitutils.get_git_commits(self.srcpath, start, end)
+        else:
+            commits = [start, end]
         if not commits or len(commits) < 2:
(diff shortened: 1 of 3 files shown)


bisection.py
```diff
@@ -220,26 +239,31 @@ class TorchBench:
...
         self.branch = branch
+        self.models = list()
 
     def prep(self) -> bool:
         # Verify the code in srcpath is pytorch/benchmark
         repo_origin_url = gitutils.get_git_origin(self.srcpath)
         if not repo_origin_url == TORCHBENCH_GITREPO:
             print(f"WARNING: Unmatched repo origin url: {repo_origin_url} with standard {TORCHBENCH_GITREPO}")
+        # get list of models
+        self.models = [ model for model in os.listdir(os.path.join(self.srcpath, "torchbenchmark", "models"))
```
wconstab, line 265: any reason to do this instead of importing 'list_models' function from torchbenchmark?

wconstab (approved):

bisection.py
xuzhao9: When this function is called, the pytorch/torchvision/torchtext packages are not built and installed yet (they are built later during bisection/abtesting). Therefore, the list_models function doesn't work at this moment.

PR #367, opened 2021-04-27
Fix a bug in the bisector filter.

xuzhao9:


bisection.py
```diff
@@ -47,15 +47,15 @@ def targets_to_bmfilter(targets: List[str], models: List[str]) -> str:
         return "(not slomo)"
     for test in targets:
         regex = re.compile("test_(train|eval)\[([a-zA-Z0-9_]+)-([a-z]+)-([a-z]+)\]")
-        m = regex.match(test).groups()
+        m = regex.match(test)
         if not m:
             if m in models:
```
wconstab, line 52: hmm, is line 52 supposed to be 'if test in models'?

wconstab (approved):
i'm not sure if my comment is correct or not, so i'll approve now and you can decide

bisection.py
xuzhao9: Ah, right! Thanks for pointing out!

PR #369, opened 2021-04-29
Relax the CPU frequency requirement.

xuzhao9:
In practice, the p-state driver often failed to pin the CPU frequency within the 10MHz threshold. We relax the threshold to 15MHz.

For example, I observe the following error message of the failed a few times (with no_turbo set to 1):

Specify frequency 2500 Mhz, find setting cpu30 scaling_cur_freq: 2489.495.


wconstab (approved):
i think this amount of change is fine.

PR #370, opened 2021-05-03
Ignore the empty json in the result directory.

xuzhao9:
When benchmarking with the PR CI, the workflow sometimes cancels or timeouts. In this case, a directory will be created but there won't be json result files, or the json result file is empty.

When we print the result, we simply ignore those files and load the latest non-empty json files only.

### DIFF
--- a/bisection.py
+++ b/bisection.py
@@ -58,10 +58,17 @@ def targets_to_bmfilter(targets: List[str], models: List[str]) -> str:
         bmfilter_names.append(f"({partial_name})")
     return "(" + " or ".join(bmfilter_names) + ")"
 
+# Find the latest non-empty json file in the directory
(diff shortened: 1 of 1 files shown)


wconstab (approved):
lgtm. is empty json happening often?  due to test crash or ctrl-c during debug?  could also try to avoid creating the empty files but not sure it's important

xuzhao9:
> lgtm. is empty json happening often? due to test crash or ctrl-c during debug? could also try to avoid creating the empty files but not sure it's important

It happens quite often, mainly because rerunning the workflow. For example, before the previous workflow finishes, the PR author pushes another update to the PR, then the GHA will cancel the running workflow and trigger a new one.

PR #371, opened 2021-05-06
Fix broken links in ADDING_MODELS.md

xuzhao9:
Fix the broken links and text in ADDING_MODELS.md.

The result after the fix:
https://github.com/xuzhao9/benchmark/blob/xz9/fix-add-models/torchbenchmark/models/ADDING_MODELS.md

Also fix a small bug in local run.py.

### DIFF
--- a/run.py
+++ b/run.py
@@ -40,7 +40,7 @@ if __name__ == "__main__":
 
     found = False
     for Model in list_models():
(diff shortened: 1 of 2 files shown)


eircfb (approved):

wconstab (approved):
thanks!

PR #372, opened 2021-05-18
Refine instruction count job

robieta:
This PR adds several finesses and fixes a breakage to the current job:
1) Use `BENCHMARK_USE_DEV_SHM=1` now that https://github.com/pytorch/pytorch/pull/56711 is in.
2) Extend `torch_nightly.py` to be able to generate a version spec, and pin the install: `pip install --pre torch=="${TORCH_VERSION_STR}" -f https://download.pytorch.org/whl/nightly/${CUDA_VERSION}/torch_nightly.html`
Previously TorchVision was installed first and picked up 1.8 release version, which made the "nightly" install a no-op.
3) Misc defensive changes to the install (e.g. `--no-default-packages`, checking that `import torch` is the right torch, etc) This is from when I was running down (2), but it's generally good practice so I'm leaving it in.
4) Clear out `/tmp/vgdb-pipe-*`, which comes from killed Valgrind processes.
5) Add an `is_debug_entry` to the scribe upload so I can test without contaminating the "real" data. (Basically, give a bit for the dashboard to filter with.)


scripts/upload_scribe.py
```diff
@@ -169,10 +169,10 @@ def post_torchbench_score(self, pytest_json, score):
     parser.add_argument("--torchbench_score", type=float,
                         help="optional torchbench score to include")
     args = parser.parse_args()
-    
+
```
xuzhao9, line 172: Thanks for fixing the style issue!

.github/scripts/instr-count-benchmark/create-nightly-env.sh
```diff
@@ -23,16 +24,22 @@ conda install -y valgrind -c conda-forge
...
+if [ -z "${NIGHTLIES}" ]; then
     echo "Torch nightly build failed. Cancel the workflow."
     exit 1
 fi
 
+TORCH_VERSION_STR=$(python torchbenchmark/util/torch_nightly.py --packages torch --pip_str)
+
 # Install PyTorch nightly from pip
-pip install --pre torch \
+pip install --pre torch=="${TORCH_VERSION_STR}" \
```
xuzhao9, line 38: May I ask why you are installing pytorch with the version number get from torch_nightly.py?

xuzhao9 (approved):
LGTM!

PR #373, opened 2021-05-18
NVidia Deep Recommender as PyTorch Benchmark

eircfb:
Uses generated 1 week data

Supports cpu and cuda, eval and train.

jit is not supported with this commit.

Output from the 4 supported configurations:

python run.py nvidia_deeprecommender -m eager -d cpu -t train
Running train method from nvidia_deeprecommender on cpu in eager mode
Ran in 55.664320945739746 seconds.

python run.py nvidia_deeprecommender -m eager -d cpu -t eval
Running eval method from nvidia_deeprecommender on cpu in eager mode
Ran in 17.123261213302612 seconds.

python run.py nvidia_deeprecommender -m eager -d cuda -t train
Running train method from nvidia_deeprecommender on cuda in eager mode
Ran in 12.558531999588013 seconds.

python run.py nvidia_deeprecommender -m eager -d cuda -t eval
Running eval method from nvidia_deeprecommender on cuda in eager mode
Ran in 9.746833801269531 seconds.


torchbenchmark/models/nvidia_deeprecommender/nvinfer.py
```diff
@@ -0,0 +1,239 @@
...
+    if self.args.jit:
+      self.rencoder = torch.jit.script(self.rencoder)
+  
+    self.inv_userIdMap = {v: k for k, v in self.data_layer.userIdMap.items()}
+    self.inv_itemIdMap = {v: k for k, v in self.data_layer.itemIdMap.items()}
+  
+    self.eval_data_layer.src_data = self.data_layer.data
+
+  def eval(self, niter=1):
+      for i, ((out, src), majorInd) in enumerate(self.eval_data_layer.iterate_one_epoch_eval(for_inf=True)):
```
wconstab, line 193: i didn't run the code, but i'm guessing the iterate_one_epoch_Eval function here is producing a whole sequence of training steps.

in our paradigm, we just want to take one step, and then we have benchmark driver logic calling your eval() function a bunch of times to takes statistical measurements - so we don't need or benefit from the averaging that goes on by running in a loop on the inside.  Instead, the benchmark driver will run this a minimum number of times, accruing a multiple of the slow 10s of sec runtime of this eval func.

Really, what you want to do in your wrapper is directly interact with the module.  in this case, the module is DeepRecommenderInferenceBenchmark.rencoder

Some folks take the approach of duplicating the setup code (code that builds rencoder) from this file in their own __init__ file.  It would also be fine and probably better if you just construct the DeepRecommenderInferenceBenchmark as you do, then reach into it and use the rencoder attribute directly in your wrapper.

torchbenchmark/models/nvidia_deeprecommender/nvtrain.py
```diff
@@ -0,0 +1,403 @@
...
+          log_var_and_grad_summaries(self.logger, self.rencoder.encode_b, self.global_step, "Encode_b")
+          if not self.rencoder.is_constrained:
+            log_var_and_grad_summaries(self.logger, self.rencoder.decode_w, self.global_step, "Decode_W")
+          log_var_and_grad_summaries(self.logger, self.rencoder.decode_b, self.global_step, "Decode_b")
+  
+      self.total_epoch_loss += loss.item()
+      self.denom += 1
+  
+      #if args.aug_step > 0 and i % args.aug_step == 0 and i > 0:
+      if self.args.aug_step > 0:
```
wconstab, line 330: -this block can also be omitted.  it's just adding random noise to the same input aug_step times, and doing another training step for each of those samples.  we can just benchmark the time to do one step.  you'll note this block is a copypaste of the block just below which also computes model output, loss, and steps the optimizer.

torchbenchmark/models/nvidia_deeprecommender/nvtrain.py
```diff
@@ -0,0 +1,403 @@
...
+  
+      loss, num_ratings = model.MSEloss(outputs, inputs)
+      loss = loss / num_ratings
+      loss.backward()
+      self.optimizer.step()
+      self.global_step += 1
+      self.t_loss += loss.item()
+      self.t_loss_denom += 1
+  
+      if not self.args.nooutput:
```
wconstab, line 314: this block can be omitted

torchbenchmark/models/nvidia_deeprecommender/nvtrain.py
```diff
@@ -0,0 +1,403 @@
...
+    if self.args.noise_prob > 0.0:
+      self.dp = nn.Dropout(p=self.args.noise_prob)
+
+  def DoTrain(self):
+  
+    self.rencoder.train()
+    #if self.args.optimizer == "momentum":
+    #  self.scheduler.step()
+  
+    for i, mb in enumerate(self.data_layer.iterate_one_epoch()):
```
wconstab, line 299: the training function will be more complex of course.. but the same trick.  you need to end up with the model, so you can just construct an object of the train benchmark for that purpose.  but you'll have to duplicate the DoTrain code and cut out some parts.

- the for loop should not exist, instead you'll just want to intercept (in pdb) the first output from it, save that to a pickle, then prepare it in your __init__() so that it's already unpickled and on cuda and in dense format, then can just be used directly by the train function

torchbenchmark/models/nvidia_deeprecommender/__init__.py
```diff
@@ -0,0 +1,88 @@
...
+    self.infer_obj = DeepRecommenderInferenceBenchmark(device = device, jit = jit)
+    if jit:
+      self.notimplementedreason = "Jit Not Supported"
+    
+    elif self.devicename != "cpu" and self.devicename != "cuda":
+      self.notimplementedreason = "device type not supported"
+
+    self.evalMode = True
+
+  def get_module(self):
```
wconstab, line 29: once you change your eval() to invoke self.infer_obj.rencoder directly, you'll be able to just return self.infer_obj.rencoder here.  likewise for train mode, you can just return self.train_obj.model here, i think.  although i'm not sure if other benchmarks implemented get_module for both train and eval, or just eval.

wconstab (changes requested):
made some comments inline- basically, the benchmark will run too slowly as is, since internally it takes more steps and also does some stuff that will add noise such as invoking the dataloader, transferring stuff to the gpu, and possibly printing stuff.

torchbenchmark/models/nvidia_deeprecommender/__init__.py
```diff
@@ -0,0 +1,93 @@
...
+      self.notimplementedreason = "Jit Not Supported"
+
+    elif self.devicename != "cpu" and self.devicename != "cuda":
+      self.notimplementedreason = "device type not supported"
+
+    elif self.devicename == "cuda" and torch.cuda.is_available() == False:
+      self.notimplementedreason = "cuda not available on this device"
+
+    else:
+      self.train_obj = DeepRecommenderTrainBenchmark(device = device, jit = jit)
```
xuzhao9, line 31: Could you please unify the style of naming of instance variables? The style of pytorch is to use underscore style.

torchbenchmark/models/nvidia_deeprecommender/__init__.py
```diff
@@ -0,0 +1,93 @@
...
+
+  def get_module(self):
+    if self.evalMode:
+        return lambda x: self.eval(), [0]
+
+    return lambda x: self.train(), [0]
+
+  def set_eval(self):
+    self.evalMode = True
+    return
```
xuzhao9, line 42: The "return" statement can be removed

torchbenchmark/models/nvidia_deeprecommender/__init__.py
```diff
@@ -0,0 +1,93 @@
...
+    self.evalMode = False
+    return
+
+  def train(self, niter=1):
+    self.checkimplemented()
+
+    for i in range(niter):
+      self.train_obj.train(self.train_obj.args.num_epochs)
+  
+    self.cleanup()
```
xuzhao9, line 54: The `cleanup()` call is not needed because it is handled by the framework:  https://github.com/pytorch/benchmark/blob/master/test_bench.py#L48

torchbenchmark/models/nvidia_deeprecommender/__init__.py
```diff
@@ -0,0 +1,93 @@
...
+      raise NotImplementedError(self.notimplementedreason)
+
+  def cleanup(self):
+    self.train = 0
+    self.infer = 0
+    gc.collect()
+    if self.devicename == "cuda":
+      torch.cuda.empty_cache()
+
+  def timedInfer(self):
```
xuzhao9, line 75: Please rename it to "timed_infer". Similarly, rename "timedTrain" to "timed_train".

torchbenchmark/models/nvidia_deeprecommender/__init__.py
```diff
@@ -0,0 +1,93 @@
+import torch
```
xuzhao9, line 1: Please add a documentation string at the beginning of this file. It should contain the following information:
- Which files are copied from the original repository
- The release or commit hash of the original repository you copied it from
- Clarify which configuration the model supports. For example, DeepRecommender model supports only eager mode, and supports both CPU and GPU

torchbenchmark/models/nvidia_deeprecommender/origin
```diff
@@ -0,0 +1 @@
+origin	https://github.com/nvidia/deeprecommender
```
xuzhao9, line 1: Looks like this file can be removed?

torchbenchmark/models/nvidia_deeprecommender/__init__.py
xuzhao9: In my case I only return the train model, because the eval one just loads the model constructed from train.

torchbenchmark/models/nvidia_deeprecommender/nvinfer.py
xuzhao9: Can you clarify what do you mean by "a whole sequence of training steps"? First, the `iterate_one_epoch_eval` function is to run one epoch in eval, not train. For example, do you mean we should only run one iteration of the loop here(https://github.com/pytorch/benchmark/pull/373/files#diff-24c21d850287bf146bc126be1a92b29f448b6f4d55f621bd91defa367dc9a35cR111)?

torchbenchmark/models/nvidia_deeprecommender/nvtrain.py
xuzhao9: I understand the pickle part. Could you please articulate what should be in the `train()` function (suppose the data is already unpickled)?

xuzhao9 (changes requested):
Let's address all the comments and I will take another look.

torchbenchmark/models/nvidia_deeprecommender/__init__.py
eircfb: I got memory complaints before I added this. The memory cleanup is called out in the ADDING_MODELS doc. Also isn't the line above is unreached?

torchbenchmark/models/nvidia_deeprecommender/origin
eircfb: The ADD_MODELS doc says to add an origin file with original source of test code. I made it the same as the attention_is_all_you_need_pytorch model. Is there some other format we're supposed to use?

wconstab: it looks good to me. why remove @xuzhao9?

(it's possible that some people haven't been following that suggestion, but it is something i wanted done for every model)

torchbenchmark/models/nvidia_deeprecommender/__init__.py
wconstab: in many cases there will only be one model obj.  it's fine to just return that.

torchbenchmark/models/nvidia_deeprecommender/__init__.py
wconstab: what were the memory complaints? is it still happening after the other changes you made?  

agreed this shouldn't be necessary, possible there is some underlying bug it's masking

torchbenchmark/models/nvidia_deeprecommender/__init__.py
eircfb: done

torchbenchmark/models/nvidia_deeprecommender/origin
xuzhao9: > it looks good to me. why remove @xuzhao9?
> 
> (it's possible that some people haven't been following that suggestion, but it is something i wanted done for every model)

Sorry I missed the instruction in ADD_MODELS. Please ignore my comment.

xuzhao9:
Closed as superseded by #403

PR #378, opened 2021-05-21
update template file; add optimize_for_inference

Krovatkin:
This change is the first set of models we are switching to be run in the optimized_for_inference mode.


torchbenchmark/models/alexnet/__init__.py
```diff
@@ -8,13 +8,13 @@
...
     task = COMPUTER_VISION.CLASSIFICATION
-    def __init__(self, device="cpu", jit=False):
+    def __init__(self, device=None, jit=False):
         super().__init__()
         self.device = device
         self.jit = jit
         self.model = models.alexnet().to(self.device)
         if self.jit:
-            self.model = torch.jit.script(self.model)
+            self.model = torch.jit.optimize_for_inference(torch.jit.script(self.model))
```
xuzhao9, line 17: self.model is also used for train() and get_module(), could you please create another model instance and name it jit_freeze_model? And in eval(), use self.jit_freeze_model to evaluate. For `get_modules()`, I prefer to use the train model, so no need to change it.

xuzhao9 (changes requested):
Please see my inline comments, and if you agree, we should change it for every model using jit freeze. Also, for the models enabling jit_freeze, we should change their test names.

torchbenchmark/models/gen_torchvision_benchmarks.py
```diff
@@ -7,38 +7,62 @@
...
+            self.eval_model = torch.jit.optimize_for_inference(self.eval_model)
         self.example_inputs = {example_inputs}
 
     def get_module(self):
         return self.model, self.example_inputs
 
+    # vision models have another model
+    # instance for inference that has
+    # already been optimized for inference
+    def set_eval(self):
```
xuzhao9, line 50: May I ask what is the purpose of this function?

xuzhao9 (approved):
LGTM overall, looking forward to the result!

torchbenchmark/models/gen_torchvision_benchmarks.py
Krovatkin: `set_eval` is defined `BenchmarkModel` and by default it runs `get_module()[0].eval()` and it's overriden for the models that don't use `get_module()` for `test_eval()`. Here's I'm just overriding it for clarity to do nothing since we have a whole new model instance for inference.

PR #379, opened 2021-05-22
Add the Deep Reinforcement Learning (DQN) model from Rainbow

xuzhao9:
The model supports train and eval mode running with both CPU and CUDA. It doesn't support JIT.
The benchmark plays only "Tetris" Atari game.

The performance data is shown below.
```
$ python run.py -d cpu -m eager -t train deep_reinforcement_learning
Running train method from deep_reinforcement_learning on cpu in eager mode
Ran in 0.019966840744018555 seconds.
$ python run.py -d cpu -m eager -t eval deep_reinforcement_learning
Running eval method from deep_reinforcement_learning on cpu in eager mode
Ran in 0.019075393676757812 seconds.
$ python run.py -d cuda -m eager -t eval deep_reinforcement_learning
Running eval method from deep_reinforcement_learning on cuda in eager mode
Ran in 0.23907780647277832 seconds.
$ python run.py -d cuda -m eager -t train deep_reinforcement_learning
Running train method from deep_reinforcement_learning on cuda in eager mode
Ran in 0.24699902534484863 seconds.
```


torchbenchmark/models/deep_reinforcement_learning/__init__.py
```diff
@@ -0,0 +1,74 @@
...
+            if done:
+                state = self.env.reset()
+            if T % self.args.replay_frequency == 0:
+                self.dqn.reset_noise()  # Draw a new set of noisy weights
+            action = self.dqn.act(state)  # Choose an action greedily (with noisy weights)
+            next_state, reward, done = self.env.step(action)  # Step
+            if self.args.reward_clip > 0:
+                reward = max(min(reward, self.args.reward_clip), -self.args.reward_clip)  # Clip rewards
+            self.mem.append(state, action, reward, done)  # Append transition to memory
+            if T >= self.args.learn_start:
```
wconstab, line 47: are you setting learn_start to 0, so this line is covered by the benchmark?

torchbenchmark/models/deep_reinforcement_learning/__init__.py
```diff
@@ -0,0 +1,74 @@
...
+                self.mem.priority_weight = min(self.mem.priority_weight + priority_weight_increase, 1)  # Anneal importance sampling weight β to 1
+                if T % self.args.replay_frequency == 0:
+                    dqn.learn(self.mem)  # Train with n-step distributional double-Q learning
```
wconstab, line 50: same question for all the stuff that's conditional on T.  

I guess there are 2 ways I could think to approach this:

(1)- simple way, but a little less faithful to the original benchmark:
  just remove all the conditions (or force them to True), so each call to train() does all of the optional stuff.
   This is 'worse' since it distorts/magnifies the importance of the less frequent parts, leading to inaccurate benchmarking
  
(2)- more faithful way:
  increase niter to the point where you are sure you're covering all the branches of the conditionals.  This is 'better' since it would more heavily weight the computations that happen more frequently in real training,
  
(0) - current way:
  if niter == 1, then the current code would skip some steps, which is another form of distortion 
  
I guess my vote would be to try (2) and see how slow it is.  If it's too slow, maybe lean towards (1) or consider increasing frequencies a bit but not all the way to 'every time', so it can be a compromise

torchbenchmark/models/deep_reinforcement_learning/__init__.py
```diff
@@ -0,0 +1,74 @@
...
+            return NotImplemented
+        # Set DQN to evaluation mode
+        self.dqn.eval()
+        self.env.eval()
+        done = True
+        for iter in range(niter):
+            if done:
+                state, reward_sum, done = self.env.reset(), 0, False
+            action = self.dqn.act_e_greedy(state)
+            state, reward, done = self.env.step(action)
```
wconstab, line 67: i'm not sure what the best policy is for the env.  The env can be a very cpu-hungry portion of the workload, so leaving it out with a 'mock' is distorting.  Otoh, the env isn't part of pytorch at all, so if we find it is a bottleneck in the workload, we'll still do nothing about it.  I think on balance I lean towards mocking the env.  

another factor is where the env runs.  in this setup, is the env making any network calls or it's strictly local?  hopefully the latter.

xuzhao9: In this model, the env is atari_py, the Atari Arcade Learning Environment. It is a local package and doesn't make network call. I agree that simulation is a better way to go. I will create a MockEnv to simulate it.

PR #381, opened 2021-05-25
Revert to use CUDA 11.1 supported by pytorch nightly.

xuzhao9:
Revert to use CUDA 11.1 because of the official pytorch nightly build target only supports CUDA 11.1.


Krovatkin (approved):
:shipit:

PR #382, opened 2021-05-25
Add Soft Actor Critic (SAC) reinforcement learning model

xuzhao9:
Original repository: https://github.com/jakegrigsby/deep_control/tree/2b02646c1586fd5baa23d1ee078539ccb29c16d8

It supports CUDA and CPU, train and eval.


torchbenchmark/models/soft_actor_critic/__init__.py
```diff
@@ -0,0 +1,243 @@
...
+class Model(BenchmarkModel):
+    task = REINFORCEMENT_LEARNING.OTHER_RL
+    def __init__(self, device=None, jit=False):
+        super(Model, self).__init__()
+        self.device = device
+        self.jit = jit
+        self.args = SACConfig()
+        # Construct agent
+        current_dir = os.path.dirname(os.path.abspath(__file__))
+        self.train_env = load_gym(self.args.env_id, self.args.seed)
```
wconstab, line 123: wait, are you actually using the gym for this? i thought you were going to mock the env

torchbenchmark/models/soft_actor_critic/__init__.py
```diff
@@ -0,0 +1,243 @@
...
+                    actor_clip=self.args.actor_clip,
+                    update_policy=step % self.args.actor_delay == 0,
+                    device=self.device
+                )
+
+            # move target model towards training model
+            if not self.args.self_regularized and (step % self.args.target_delay == 0):
+                soft_update(self.target_agent.critic1, self.agent.critic1, self.args.tau)
+                soft_update(self.target_agent.critic2, self.agent.critic2, self.args.tau)
+    # Set default ninter to 5 because latency of one iteration is too small (~6ms)
```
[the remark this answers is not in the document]
xuzhao9: I remember Nikita mentioned 1~5ms is too small as a measure result because it is close to kernel scheduling jiffies (we don't have this problem on CI machine, but other people may want to run TorchBench locally). Also, I hope to avoid the problem like (https://github.com/pytorch/pytorch/issues/56419) where we are too sensitive to a small regression in the code path.

torchbenchmark/models/soft_actor_critic/__init__.py
xuzhao9: I cut the "simulation boundary" at mujoco, not gym. This model uses gym but doesn't use mujoco, so I don't mock the env here. Similarly in https://github.com/pytorch/benchmark/pull/387, I mock the mujoco environment but still use gym.spaces. Please let me know whether this is reasonable, or you suggest we mock both mujoco and gym, and get rid of the gym dependency.

xuzhao9 (commented):
@wconstab Please let me know your feedbacks, thanks!

torchbenchmark/models/soft_actor_critic/__init__.py
wconstab: I think if you can make a clear argument for why you set that boundary, so that others can follow the precedent in the future, I am ok with it.

torchbenchmark/models/soft_actor_critic/__init__.py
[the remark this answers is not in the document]
wconstab: If the solution to fast tests is to loop them more, 

and given that we already have 'smart' stuff set up in pytest to determine how many times to repeat each test,

I would suggest just setting the minimum time per test in pytest settings to something you're happy with, rather than adding another place where you might tweak it.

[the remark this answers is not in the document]
xuzhao9: I see, I will set the number of iterations to be the minimum that can cover all the code in train().

torchbenchmark/models/soft_actor_critic/__init__.py
```diff
@@ -0,0 +1,243 @@
...
+            if done:
+                state = self.train_env.reset()
+                steps_this_ep = 0
+                done = False
+            action = self.agent.sample_action(state)
+            next_state, reward, done, info = self.train_env.step(action)
+            self.buffer.push(state, action, reward, next_state, done)
+            state = next_state
+            steps_this_ep += 1
+            if steps_this_ep >= self.args.max_episode_steps:
```
bitfort, line 202: Nice job checking end conditions :)

torchbenchmark/models/soft_actor_critic/__init__.py
```diff
@@ -0,0 +1,243 @@
...
+
+    def eval(self, niter=1):
+        with torch.no_grad():
+            discount= 1.0
+            episode_return_history = []
+            for episode in range(self.args.eval_episodes):
+                episode_return = 0.0
+                state = self.test_env.reset()
+                done, info = False, {}
+                for step_num in range(niter):
```
bitfort, line 235: Is there any difference between having the niter loop inside the episode loop versus outside of it?

torchbenchmark/models/soft_actor_critic/requirements.txt
```diff
@@ -0,0 +1 @@
+gym
```
bitfort, line 1: Is this a new torchbench requirement too?

bitfort (approved):
Thanks Xu! I left a few comments.

torchbenchmark/models/soft_actor_critic/requirements.txt
xuzhao9: Yes, it will be a new torchbench requirement because `install.py` will recursively install the dependencies of all models.

torchbenchmark/models/soft_actor_critic/__init__.py
xuzhao9: My guess is that the outer loop is the epoch, and the inner loop specifies how many steps we would like to run in an epoch. Therefore, I update the code so that we run `niter` epochs and each epoch we run `self.args.max_episode_steps` steps.

Original code reference:
https://github.com/jakegrigsby/deep_control/blob/2b02646c1586fd5baa23d1ee078539ccb29c16d8/deep_control/run.py#L12

PR #385, opened 2021-05-27
Enable profile directed typing using _script_pdt API

nikithamalgifb:
Summary:
========
`torch.jit._script_pdt` is an API similar to `torch.jit.script` which infers types automatically using a tool called MonkeyType.
Under the hood of `_script_pdt` API, some sample example inputs are run with the `method` or `nn.Module` obj which needs to be scripted and then using those types, the types are inferred. This helps in eliminating TorchScript's much required type annotation to be used while scripting a `nn.Module` or a function argument. 

This PR replaces `torch.jit.script` API with `torch.jit._script_pdt` API for few of the torchbenhcmark models.


wconstab (approved):
lgtm! As we discussed offline, jit.script is still fully covered by script_pdt so nothing is lost here.  It doesn't seem super important to be 'consistent' (all models use either script or pdt), it's OK to just have the ones that benefit from _pdt using it now, and then later the APIs will be merged anyway.

PR #387, opened 2021-05-27
Add the DrQ model

xuzhao9:
Original repository: https://github.com/denisyarats/drq
Removed the dependency of mujoco by pickling the input array `obs` and mock the environment.

```
$ python run.py -t train -d cuda drq
Running train method from drq on cuda in eager mode
Ran in 0.6877460479736328 seconds.
$ python run.py -t eval -d cuda drq
Running eval method from drq on cuda in eager mode
Ran in 0.22950458526611328 seconds.
$ python run.py -t eval -d cpu drq
Running eval method from drq on cpu in eager mode
Ran in 0.0064280033111572266 seconds.
$ python run.py -t train -d cpu drq
Running train method from drq on cpu in eager mode
Ran in 0.7551465034484863 seconds.
```


torchbenchmark/models/drq/__init__.py
```diff
@@ -0,0 +1,154 @@
...
+
+    def train(self, niter=2):
+        episode, episode_reward, episode_step, done = 0, 0, 1, True
+        for step in range(niter):
+            obs = self.env.reset()
+            done = False
+            episode_reward = 0
+            episode_step = 0
+            episode += 1
+            if step < self.cfg.num_seed_steps:
```
bitfort, line 121: It looks like by default with niters = 2, this will always take one path? Should we have a fork based on step number for benchmarking? Do we just want to benchmark one, the other, or both?

torchbenchmark/models/drq/origin
```diff
@@ -0,0 +1 @@
+https://github.com/denisyarats/drq/tree/365e5969f9a748ad63d854ce8c8cc8f90c1de4e0
```
bitfort, line 1: Is this a submodule? Do we actually want this committed?

bitfort (approved):
Thanks Xu for pushing forward on this!

torchbenchmark/models/drq/origin
xuzhao9: It is not a submodule, in [ADD_MODEL]`https://github.com/pytorch/benchmark/blob/master/torchbenchmark/models/ADDING_MODELS.md` instructions we would like to add this file so that we can track where the model code is originally from.

torchbenchmark/models/drq/__init__.py
xuzhao9: We would like all the code path to be covered by the default niters. In the [config](https://github.com/pytorch/benchmark/pull/387/files#diff-1c5eb1e60aafbb155ddfd1a8291eab9d24200fd71ae402157b100f4647d18093R17) file we set `num_seed_steps` to be 1. Because niters = 2 by default, both paths will be executed (once when niter=0, the other is when niter=1).

wconstab: I just want to point out that making each path execute once per benchmark is a dangerous assumption.  do you know how big num_seed_steps was by default?  

if the behavior of the model is basically to do the first path N times, and then do the second path for the rest of the time, then in the limit, probably the second path is all we really need to benchmark, or at least making them 1:1 is pretty skewing in terms of faithful representation of the workload.

PR #389, opened 2021-06-02
use mobilenet_v3 from torchvision

Krovatkin:


xuzhao9 (changes requested):
I think it is a nice improvement. Thanks! 

However, could you please keep files in the  `torchbenchmark/score` directory as-is (with the exception of `config-v0.md`, where we should explicitly say pytorch_mobilenet_v3 is replaced by mobilenet_v3_large)? Changing the yml files will essentially redefine the v0 score, which is not what we want.
Although the active v0 `yml` config files are on the `0.1` branch, currently we would like to keep a copy of it on the master branch as well.

xuzhao9 (approved):
LGTM

PR #391, opened 2021-06-03
Add freeze suffix to test names

Krovatkin:


xuzhao9 (approved):
Looks good to me.

It will be better to add a a test in the test.py file that checks the `optimized_for_inference` field is set to `True` for all Models that enabled freeze and has the `eval_model` field.

xuzhao9 (changes requested):
I am wondering if `optimized_for_inference` field is set for all models that enabled freeze?

xuzhao9 (approved):

bitfort (approved):
Thanks!

Krovatkin:
> I am wondering if optimized_for_inference field is set for all models that enabled freeze?

@xuzhao9  it's not yet. I just wanted to get your feedback first.
I'll add it to all frozen models now.

Krovatkin:
> It will be better to add a a test in the test.py file that checks the optimized_for_inference field is set to True for all Models that enabled freeze and has the eval_model field.

that's a great idea! I'll add a check. I wish I could check for `eval_model` when constructing tests but at that point we don't construct model objects yet (((((

PR #396, opened 2021-06-08
Preload the data to train_data in __init__ for Background_Matting

### LINKED ISSUE #388: Background_Matting misconfigured for training
In the Backgound_Matting model, the train is bottlenecked by dataloader, not the actual computation.
We should remove the use of dataloader in train().

xuzhao9:
The idea of torchbenchmark is to measure the computation intensive part of a model. Therefore, we should move the data loading process from `train()` to `__init__` to avoid measuring the data loading speed.

Fixes #388

### DIFF
--- a/torchbenchmark/models/Background_Matting/__init__.py
+++ b/torchbenchmark/models/Background_Matting/__init__.py
@@ -51,8 +51,11 @@ class Model(BenchmarkModel):
             'reso': (self.opt.resolution, self.opt.resolution)}
         traindata = VideoData(csv_file=csv_file,
                               data_config=data_config_train, transform=None)
-        self.train_loader = torch.utils.data.DataLoader(
+        train_loader = torch.utils.data.DataLoader(
             traindata, batch_size=self.opt.batch_size, shuffle=True, num_workers=self.opt.batch_size, collate_fn=_collate_filter_none)
+        self.train_data = []
+        for data in train_loader:
+            self.train_data.append(data)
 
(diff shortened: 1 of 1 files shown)


torchbenchmark/models/Background_Matting/__init__.py
```diff
@@ -51,8 +51,11 @@ def __init__(self, device=None, jit=False):
             'reso': (self.opt.resolution, self.opt.resolution)}
         traindata = VideoData(csv_file=csv_file,
                               data_config=data_config_train, transform=None)
-        self.train_loader = torch.utils.data.DataLoader(
+        train_loader = torch.utils.data.DataLoader(
             traindata, batch_size=self.opt.batch_size, shuffle=True, num_workers=self.opt.batch_size, collate_fn=_collate_filter_none)
+        self.train_data = []
+        for data in train_loader:
+            self.train_data.append(data)
```
wconstab, line 58: do you need to do `.to(device)` here?

wconstab (approved):
lgtm except for device issue if needed

torchbenchmark/models/Background_Matting/__init__.py
xuzhao9: Good point! Originally we copy data to cuda in `train()`. We should do that in `__init__()` as well.

xuzhao9: Removed the redundant `cuda()` calls in `train()` because they are already copied to device in `__init__()`.

PR #397, opened 2021-06-08
Quantization test model for resnet50

DevSatpathy:
Summary: This is the resnet50 model with additional functions to enable prep and conversion to qat, emulating the work done in #323. The benchmark tests run by the command 
with-proxy pytest test_bench.py -k "resnet50_quantized_qat" --ignore_machine_config
hang when they get to the eager tests, but this behavior is also exhibited by running the similar command
with-proxy pytest test_bench.py -k "mobilenet_v2_quantized" --ignore_machine_config 
that is described in #323 which this PR emulates.

Test Plan:
python test.py -k "test_resnet50_quantized_qat_train_cpu"
python test.py -k "test_resnet50_quantized_qat_eval_cpu"


jerryzh168 (approved):
lgtm, cc @vkuzo as well.

vkuzo (approved):

PR #400, opened 2021-06-10
Adding TTS Angular Loss Encoder

bitfort:
To test:
python run.py tts_angular


torchbenchmark/models/tts_angular/__init__.py
```diff
@@ -0,0 +1,38 @@
...
+        return self.model, []
+
+    def set_train(self):
+        # another model instance is used for training
+        # and the train mode is on by default
+        pass
+
+    def train(self, niterations=1):
+        # the training process is not patched to use scripted models
+        self.model.train()
```
xuzhao9, line 27: `niterations` is not used, we need to pass this parameter to train to control how many epochs we want to run for training and inference.

torchbenchmark/models/tts_angular/angular_tts_main.py
```diff
@@ -0,0 +1,324 @@
...
+import os
+import sys
+import time
+import traceback
+import math
+
+import torch
+import torch as T
+from .model import SpeakerEncoder, AngleProtoLoss
+from .audio import AudioProcessor
```
xuzhao9, line 14: We are using generated input instead of actual audio file, so all the `audio.py` can be removed, I guess?

xuzhao9 (approved):
Overall looks good, we can further simplify the code by removing the audio file input and parsing part.

PR #401, opened 2021-06-11
Add V1 score

xuzhao9:
TorchBench V1 score defines an overall score with four subscores, and a jit speedup score.
The JIT speedup score compares all the JIT-enabled tests with its eager counterparts, and calculate the average speedup.
The overall score calculates the weighted average of all tests. The weight is determined by the following formula:

`weight = domain_weight * config_weight`

The `domain_weight` definition is the same as V0. The `config_weight` is defined based on the tests' configuration available, with the following scale:
- CPU Training: 1x
- CPU Inference: 2x
- GPU Training: 2x
- GPU Inference: 2x

Similar to V0, we use [geometric average formula](https://en.wikipedia.org/wiki/Weighted_geometric_mean) to calculate the weighted average.

Here is an example run of getting V1 score:

```
$ python compute_score.py --benchmark_data_file input.json --relative

[
    {
        "file": "input.json",
        "pytorch_version": "1.7.0.dev20200626",
        "score": {
            "subscore[jit]": 1041.6367342612677,
            "subscore[cpu-train]": 1000.0,
            "subscore[cpu-eval]": 1000.0,
            "subscore[cuda-train]": 1000.0,
            "subscore[cuda-eval]": 1000.0,
            "total": 1000.0
        }
    }
]
```


torchbenchmark/score/compute_score_v0.py
```diff
@@ -0,0 +1,135 @@
...
+import os
+import re
+import yaml
+import importlib
+
+from tabulate import tabulate
+from pathlib import Path
+from collections import defaultdict
+
+from .generate_score_config import generate_bench_cfg
```
bitfort, line 17: Out of curiosity, how come we use relative imports?

torchbenchmark/score/compute_score_v1.py
```diff
@@ -0,0 +1,193 @@
+"""
+Compute the benchmark score given a frozen score configuration and current benchmark data.
```
bitfort, line 2: Ideally we wouldn't need 2 files... but looking at it, I don't see a better solution so I guess 2 files is a fine solution for now :)

torchbenchmark/score/compute_score_v1.py
```diff
@@ -0,0 +1,193 @@
...
+    def _config_to_weight(self, config):
+        # config weight rule in V1: 1x CPU Training, 2x GPU Training, 2x CPU Inference, 2x GPU Inference
+        test, model_name, device, mode = config
+        if test == "train" and device == "cpu":
+            return 1.0
+        else:
+            return 2.0
+        
+    # Generate the domain weights from the ref object
+    def _setup_weights(self, ref):
```
bitfort, line 71: The implementation of this is not easy to understand from the code. Could you include a brief explanation somewhere about how weights work for v1.0?

torchbenchmark/score/compute_score_v1.py
```diff
@@ -0,0 +1,193 @@
...
+            config_weights[name] = config_dict[model_name][name] / sum(config_dict[model_name].values())
+        for name in name_dict:
+            total_weight = config_weights[name] * domain_weights[name]
+            sum_domain += total_weight
+        assert(abs(sum_domain - 1.0) < 1e-6), "The total task weights sum is not 1.0, please submit a bug report."
+        return (domain_weights, config_weights)
+ 
+    def _setup_benchmark_norms(self, ref_data):
+        """
+        Helper function which gets the normalization values per benchmark
```
bitfort, line 104: Is normalization = speedup over reference value?

torchbenchmark/score/compute_score_v1.py
```diff
@@ -0,0 +1,193 @@
+"""
+Compute the benchmark score given a frozen score configuration and current benchmark data.
+"""
```
bitfort, line 3: Can you explain at the top how many scores are generated and what they mean?

torchbenchmark/score/compute_score_v1.py
```diff
@@ -0,0 +1,193 @@
...
+            score += benchmark_score
+        return math.exp(score)
+
+    def data_in_list(self, n, l):
+        for e in l:
+            if e not in n:
+                return False
+        return True
+
+    def _get_subscore(self, data, ref_norm, ref_weights, filters):
```
bitfort, line 143: Could you clarify in comments what a subscore is; is it just Eager/JIT, CPU/GPU, /Train/Eval, Doamins, tasks?

torchbenchmark/score/compute_score_v1.py
```diff
@@ -0,0 +1,193 @@
...
+        assert "train" in filters or "eval" in filters, error_msg
+        score = 0.0
+        (domain_weights, _) = ref_weights
+        for name in filter(lambda x: self.data_in_list(x, filters), data['benchmarks']):
+            norm = data['benchmarks'][name]['norm']
+            benchmark_score = domain_weights[name] * math.log(norm / ref_norm['benchmarks'][name]['norm'])
+            score += benchmark_score
+        return math.exp(score)
+    
+    def compute_jit_speedup_score(self, data):
```
bitfort, line 155: Just checking terminology, is the JIT speedup a "subscore" or different score?

torchbenchmark/score/configs/v1/config-v1.yaml
```diff
@@ -0,0 +1,124 @@
+benchmarks:
+  test_eval[BERT_pytorch-cpu-eager]:
+    norm: 0.05529004814998188
```
bitfort, line 3: What is a "norm" here?

torchbenchmark/score/compute_score_v1.py
xuzhao9: Using two files give us flexibility to design completely different score algorithms. For example, V1 obsoletes the spec file in V0 because now each model has a "task" field so we can extract the domain hierarchy at runtime. It also makes easier to maintain and deprecate the old score versions.

torchbenchmark/score/configs/v1/config-v1.yaml
xuzhao9: It is actually the mean latency of the test. The term "norm" is inherited from V0.

torchbenchmark/score/compute_score_v1.py
xuzhao9: It is a different score because it doesn't use the reference data. Instead, it only compares the jit speedup relative to the eager version of the test.

torchbenchmark/score/compute_score_v1.py
xuzhao9: A V1 score has four config subscores: cpu-train, cpu-eval, cuda-train, cuda-eval. For each subscore, we collect all tasks belong to that config, recalculate the weights, and generate the number.
This pr doesn't provide subscore for jit/eager because I believe that should be addressed by the jit speedup score. I will add more details in the comments.

torchbenchmark/score/compute_score_v1.py
xuzhao9: Here, the norm is only the absolute latency number in seconds. However, the relative score is the weighted average of the speedup over reference value: https://github.com/pytorch/benchmark/pull/401/files#diff-64db27bada97956276b540db8c8c4056a13f748f27508e2608fae8b3be60c7b5R133

torchbenchmark/score/compute_score_v1.py
xuzhao9: Sure, will do.

torchbenchmark/score/compute_score_v0.py
xuzhao9: Because in V0 the code to generate the norm yaml file is in generate_score_cfg.py, so we use relative import to import that. This is for compatibility purpose.

torchbenchmark/score/configs/v1/config-v1.yaml
xuzhao9: It is just the mean latency of the test. The term `norm` is inherited from v0.

bitfort (approved):
Looking great Xu!

torchbenchmark/score/compute_score_v1.py
```diff
@@ -0,0 +1,193 @@
...
+        return (domain_weights, config_weights)
+ 
+    def _setup_benchmark_norms(self, ref_data):
+        """
+        Helper function which gets the normalization values per benchmark
+        by going through the reference data file.
+        """
+        if ref_data == TORCHBENCH_V1_REF_DATA:
+            with open(ref_data) as ref_file:
+                ref = yaml.full_load(ref_file)
```
wconstab, line 109: it's not obvious to me why we handle the TORCHBENCH_V1_REF_DATA case differently from other files?

torchbenchmark/score/compute_score_v1.py
```diff
@@ -0,0 +1,193 @@
...
+        else:
+            ref = self._get_norm_from_ref_data(ref_data)
+        return ref
+
+    def _get_norm_from_ref_data(self, ref_data):
```
wconstab, line 114: its a little confusing that you use the same name (ref_data) to refer to a json dict in this function, and a filename in the above function.

torchbenchmark/score/compute_score_v1.py
```diff
@@ -0,0 +1,193 @@
...
+        """
+        This function iterates over the reference data (json object)
+        and calculates the normalization values based on the reference data.
+        It also sets up the domain weights of the score.
+        """
+        norm = {}
+        norm['benchmarks'] = {}
```
wconstab, line 121: are all the complexities of this data structure necessary?  

for instance, does `norm` need to have a top-level `benchmarks` entry,  or can it just start by indexing into names?

does `d` need to be another dict, or can it just be a scalar 'mean' value?

torchbenchmark/score/compute_score_v1.py
```diff
@@ -0,0 +1,193 @@
...
+
+    def data_in_list(self, n, l):
+        for e in l:
+            if e not in n:
+                return False
+        return True
+
+    def _get_subscore(self, data, ref_norm, ref_weights, filters):
+        assert len(filters) == 2, error_msg
+        assert "cpu" in filters or "cuda" in filters, error_msg
```
wconstab, line 145: where is `error_msg` defined?  

it would be more readable if it were defined here.  You should explain the expected format and requirements for `filters` in a comment.

torchbenchmark/score/compute_score_v1.py
```diff
@@ -0,0 +1,193 @@
...
+        score = 0.0
+        norm = self._setup_benchmark_norms(data)
+        norm_jit = self._filter_jit_tests(norm)
+        (domain_weights, config_weights) = self._setup_weights(norm_jit)
+        for name in norm_jit['benchmarks']:
+            eager_norm = norm_jit['benchmarks'][name]['eager_norm']
+            jit_norm = norm_jit['benchmarks'][name]['jit_norm']
+            jit_speedup_score = domain_weights[name] * config_weights[name] * math.log(eager_norm / jit_norm)
+            score += jit_speedup_score
+        return math.exp(score) * self.target
```
wconstab, line 173: does it make sense to use the target for the JIT speedup score?  I think if you're making the speedup score a special 'relative' computation, you can just leave it normalized to 1, so you'd get explainable numbers like 1.2x speedup or something.

torchbenchmark/score/compute_score_v1.py
wconstab: i agree, it seems like there is too much complexity here, and I wonder if it can be refactored so it is more readable and less lines of code.  

I guess it wouldn't change things that much to combine the 3 for loops, but it seems like you could do that.  

another possibility to consider is creating a data structure or two.  

Maybe one data-structure that wraps a reference_data file and exposes convenient APIs on it (e.g. to get particular numbers based on names or configs) so you don't have to do as much of the filtering and extracting in the TorchBenchScoreV1 class.

Maybe another (simpler?) data structure that replaces all the defaultdicts you construct directly here- Seems like what you're really constructing here is a 'weights' object, which probably could also include the 'norms'.  maybe could be called a 'scoremeta' object or something.  and have accessor methods that get the norm or weight for a particular benchmark name. 

this is a bigger change, so i'm not sure if it's critical to do now, but it would be nice to make this more readable- I doubt others could easily maintain or understand it currently.

torchbenchmark/score/compute_score_v1.py
xuzhao9: `ref_data` could be two things: either the reference `TORCHBENCH_V1_REF_DATA`, which is in yaml, or the reference benchmark result json object. So we are handling them differently here. Maybe we should only check the extension is `yaml` or `yml`, if so, we use yaml to load, otherwise we treat them as json objects.

torchbenchmark/score/compute_score_v1.py
xuzhao9: Sorry it is a bug. I will add the expectations to the filters here.

torchbenchmark/score/compute_score_v1.py
xuzhao9: Sounds good to me. JIT speedup score is an absolute score and doesn't rely on reference data, so it can be normalized to 1.

torchbenchmark/score/compute_score_v1.py
xuzhao9: The top-level benchmarks is not really needed. I will remove it later.
`d` has to be another dict here, but I can further simplify this by using `defaultdict`.

torchbenchmark/score/compute_score_v1.py
xuzhao9: Very good point! I will create a data structure and some APIs to expose the numbers needed for weights.

torchbenchmark/score/compute_score_v1.py
xuzhao9: Right, will rename it to "ref_json_obj"

torchbenchmark/score/compute_score_v1.py
xuzhao9: There are six scores generated:
1. Total score
2. CPU-train subscore
3. CPU-infer subscore
4. GPU-train subscore
5. GPU-infer subscore
6. JIT speedup score

torchbenchmark/score/compute_score_v1.py
xuzhao9: Changed the input to be always an obj, and differentiate json and yaml obj using keys.

PR #403, opened 2021-06-11
NVidia DeepRecommender as Benchmark

eircfb:
Supports eval and train mode, cpu and cuda. Both run a single instance of model. Jit is not supported.


torchbenchmark/models/nvidia_deeprecommender/__init__.py
```diff
@@ -0,0 +1,100 @@
...
+    self.not_implemented_reason = "Implemented"
+    self.eval_mode = True #default to inference
+
+    if jit:
+      self.not_implemented_reason = "Jit Not Supported"
+
+    elif self.device_name != "cpu" and self.device_name != "cuda":
+      self.not_implemented_reason = "device type not supported"
+
+    elif self.device_name == "cuda" and torch.cuda.is_available() == False:
```
xuzhao9, line 36: Since we are not supporting cuda on this model because of utilization issue, we should return "cuda is not available on this model" when `self.device_name == "cuda"`

torchbenchmark/models/nvidia_deeprecommender/__init__.py
```diff
@@ -0,0 +1,100 @@
...
+  def set_train(self):
+    self.eval_mode = False
+
+  def train(self, niter=1):
+    self.check_implemented()
+
+    for i in range(niter):
+      self.train_obj.train(self.train_obj.args.num_epochs)
+  
+    self.cleanup()
```
xuzhao9, line 61: `cleanup()` is not needed because they are handled by the framework: https://github.com/pytorch/benchmark/blob/master/test_bench.py#L50

torchbenchmark/models/nvidia_deeprecommender/nvtrain.py
```diff
@@ -0,0 +1,419 @@
...
+          self.optimizer.step()
+  
+  def train(self, niter=1) :
+    for self.epoch in range(niter):
+
+      if self.toytest:
+        self.rencoder.train()
+        self.optimizer.zero_grad()  
+        outputs = self.rencoder(self.toyinputs)
+        return
```
xuzhao9, line 358: The toytest runs only one iteration no matter what we set for `niter`, could you please fix that?
Also, I see the `DoTrain()` function runs the `model.MSEloss()`, `loss.backward()`, and `self.optimizer.step()` for training, but not in the toytest, could you please add them back?

torchbenchmark/models/nvidia_deeprecommender/nvinfer.py
```diff
@@ -0,0 +1,254 @@
...
+  return args
+
+class DeepRecommenderInferenceBenchmark:
+
+  def __init__(self, device = 'cpu', jit=False, usecommandlineargs = False) :
+
+    self.toytest = True
+
+    if self.toytest:
+      self.toyinputs = torch.randn(1,15178).to(device)
```
xuzhao9, line 123: If we are using randomly generated toyinputs, why we are still checking in the input files such as n1w.test.txt?

xuzhao9 (changes requested):
See the inlined comments. Please remove the unused files such as logger.py and the Netflix data, since we are using randomly generated mock inputs.
Also, please double check that the toyinput data shape matches the real input.

torchbenchmark/models/nvidia_deeprecommender/nvtrain.py
eircfb: Yikes.

xuzhao9 (approved):
LGTM. Thanks for the contribution!

PR #408, opened 2021-06-22
Fix pytorch_struct broken by more strict in-place operation restrictions

xuzhao9:
Recent CI fails because pytorch restricts the usage of in-place operations. This PR fixes the failure by using copy of data structures instead of in-place operations.

### DIFF
--- a/torchbenchmark/models/pytorch_struct/torch_struct/helpers.py
+++ b/torchbenchmark/models/pytorch_struct/torch_struct/helpers.py
@@ -15,8 +15,9 @@ class Get(torch.autograd.Function):
     @staticmethod
     def backward(ctx, grad_output):
         (grad_chart,) = ctx.saved_tensors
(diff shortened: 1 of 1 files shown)


bitfort (approved):
Good fix

malfet:
Most likely caused by https://github.com/pytorch/pytorch/pull/60195
cc: @albanD

albanD:
Yes that was a big on our end were we were not detecting these inplace ops properly :/
This is the right fix!

xuzhao9:
> Yes that was a big on our end were we were not detecting these inplace ops properly :/
> This is the right fix!

Thanks for the response. This PR fixes the broken run, but will it also break the correctness? I guess in this case the ctx.saved_tensors won't be updated because we are using a copy.

albanD:
Well, in this case, not modifying the saved_tensors inplace is the right fix and the code before had correctness issue.

PR #413, opened 2021-06-22
Add the sweep workflow for v1.

xuzhao9:
This PR adds the V1 sweep workflow that runs a group of pytorch configs and generate all their benchmarking results.


bitfort (approved):
Thanks

PR #415, opened 2021-06-29
Improve the logs printed by the bisector.

xuzhao9:
This PR fixes the new dependency of "iopath" of recent torchtext package, which causes CI failures.

It also improves the logging of bisector by printing the logs to both standard output and the log file. This should be very helpful for the on-demand PR CI users.


bitfort (approved):
Great!

PR #423, opened 2021-08-05
Run tests and benchmarks in subprocesses for isolation

robieta:
#### Foreword: Let me apologize in advance for the behemoth that this PR has become.

# Process Isolation in TorchBench

This PR adapts several components from https://github.com/pytorch/pytorch/pull/62458 to run benchmarks in isolated subprocesses.

## Motivation:

Currently TorchBench operates by importing all ~50 models into a single process, and then calls them one at a time. This presents several issues:

#### 1) Non-Hermetic setup

A model in TorchBench can be thought of as executing more-or-less arbitrary code. The models are audited when added, but many are forks of large and complicated frameworks and it's not practical to go through them with a fine tooth'd comb. @Krovatkin recently discovered that some models set global configuration flags (e.g. `torch.backends.cudnn.deterministic`) without restoring them to their original state, which makes benchmark results dependent on the order of execution.

#### 2) Resource contention

For various reasons, models have a habit of checking out resources on import rather than waiting for concrete model instantiation or lazily initializing. This means that by the time the suite is loaded key resources (namely GPU memory) is bloated and fragmented with 50 models worth of cruft, and unable to run realistic workloads without OOMing. (@eircfb has done a number of experiments on this recently)

## High level approach:

Instead of importing all models in the main process, this PR allows one to launch a subprocess specifically for one model. It's not possible for models to interfere with each other, because no matter what a model does to the global state it will be erased at teardown and the next model will see a clean, consistent initial state. Freeing the process also guarantees that any system resources are freed, and the next model to run enjoys sole access. (Apart from the parent process.)

### Workers

The primary execution abstraction is a (generally clean) Python process with three core APIs:
`store`: Take a value, and assign a copy to a name inside the worker. (Generally crossing a serialization boundary.)
`load`: Fetch a variable from the worker, and materialize a copy in the main process. (Again, crossing a serialization boundary.)
`run`: Execute a snippet of Python code in the worker. <-- *This is where the magic happens.*

API: `components/_impl/workers/base.py`
Simple (in process) worker: `components/_impl/workers/in_process_worker.py`

### SubprocessWorker

Moving from in-process to subprocess introduces significant complexity. `in_process_worker` is 50 lines with liberal comments, while `subprocess_worker` (and the helper module `subprocess_rpc`) is just shy of 800. The high level flow is that `SubprocessWorker` spawns an interactive Python process, bootstraps it into an event loop, and then writes work to a work queue and waits for results from the worker.

#### Key features:
* Robust communication: The `Pipe` class wraps `os.pipe`, and robustly handles the low level details of sending messages back and forth between the parent and child.
* Error handling: When the subprocess fails to execute a command, great care is given to plumb as much information as possible back to the caller. To this effect, there is machinery to re-raise a faithful copy of the thrown exception in the parent, provide a stack trace, and display stdout and stderr for the offending command. In summary, a SubprocessWorker may fail but it strives to always do so in a reasonable manner. (Rather than hang or silently crash.)
* Tidy: When exiting, the SubprocessWorker goes to great lengths to make sure the worker is given a chance to gracefully exit, and that it doesn't orphan any processes or system resources. (e.g. files and file descriptors)

### Integration

We borrow the concept of a `Task` from the initial PyTorch Timer refactor design. In summary, `run_in_worker` is an optional decorator which can handle the details of shipping code and values to and from the worker, and allow one to write code *almost* as though it was executing locally.

#### torchbenchmark/__init__.py

Here we define a `ModelTask` class, which is specialized to a *single* benchmark and loads that benchmark in a SubprocessWorker. (As well as implementing various glue code to manipulate the `model` in the underlying subprocess.)

#### test.py and test_bench.py

The changes to these files are fairly modest. Instead of operating on a `Model` which is imported from `torchbench`, each test case instantiates a `ModelTask` which loads the benchmark of interest and forwards the calls.

## Performance

The PyTest benchmark plugin simply takes a callable and a timer, so there is no opportunity to hide the interprocess communication latency; if communication with SubprocessWorker is slow and/or highly variable it will degrade the quality of the benchmarks. For this reason, SubprocessWorker is optimized for low overhead, and in most cases is expected to simply jump from syscall to syscall with very little Python on the hot path. (Most of the heavyweight code is in error handling, and at that point we don't care about overhead.)

Test command: `pytest test_bench.py --ignore_machine_config -k "test_worker"`

Screen Shot 2021-08-11 at 7 02 43 PM

### DIFF
--- a/components/_impl/tasks/base.py
+++ b/components/_impl/tasks/base.py
@@ -16,7 +16,133 @@ class TaskBase(abc.ABC):
         ...
 
 
-# TODO: This needs to be factored into components that can be unit tested.
+def parse_f(f: typing.Callable) -> typing.Tuple[inspect.Signature, str]:
+    """Handle source code extraction for `run_in_worker`."""
+    if not inspect.isfunction(f):
+        raise TypeError(f"Expected function, got {type(f)}. ({f})")
+
+    signature = inspect.signature(f)
+
+    # It isn't strictly necessary for `f` to be completely type annotated,
+    # however one of the key advantages of `run_in_worker` over manually
+    # passing strings is that it gives the type checker an opportunity to catch
+    # errors. (Which may become more difficult once the code has been shipped)
+    # to the worker. And because this is a developer rather than user facing
+    # API, it isn't a problem to enforce this strict criteria. This also
+    # provides some weak protection against decorators. (See below.)
+    for arg, arg_parameter in signature.parameters.items():
+        if arg_parameter.kind == inspect.Parameter.VAR_POSITIONAL:
+            raise TypeError(
+                f"Variadic positional argument `*{arg}` not permitted "
+                "for `run_in_worker` function.")
+
+        if arg_parameter.kind == inspect.Parameter.VAR_KEYWORD:
+            raise TypeError(
+                f"Variadic keywork argument `**{arg}` not permitted "
+                "for `run_in_worker` function.")
+
+        if arg_parameter.annotation == inspect.Parameter.empty:
+            raise TypeError(f"Missing type annotation for parameter `{arg}`")
+
+    if signature.return_annotation == inspect.Parameter.empty:
+        raise TypeError("Missing return annotation.")
+
+    #   We serialize the function by grabbing source from the body, so
+    # decorators pose a correctness problem. The structure of a decorator is:
+    # ```
+    #   def my_decorator(wrapped_f: typing.Callable) -> typing.Callable:
+    #
+    #       @functools.wraps(wrapped_f)  # <- Optional
+    #       def inner(*args, **kwargs):  # Most decorators don't know the signature of `wrapped_f`
+    #           # Generally a call to `wrapped_f(*args, **kwargs)` appears
+    #           # somewhere, though not required.
+    #           ...
+    #
+    #       return g
+    # ```
+    #   The inclusion or omission of `functools.wraps` is rather important.
+    # If included, it will provide the breadcrumbs to map `inner` back to
+    # `wrapped_f`, and the `inspect` library (namely `signature` and
+    # `getsource`) will parse the "True" function: `wrapped_f`. (This allows,
+    # among other things, type checkers to analyze decorated code.) Otherwise
+    # `inspect` will stop at `inner`.
+    #
+    #   In the case that a function is decorated but does not use
+    # `functools.wraps`, it is HIGHLY likely that it uses variadic arguments
+    # so that it can forward them. (And thus will fail the signature checks)
+    # above. If we are passed a function which is decorated with a "proper"
+    # decorator, we catch it here by checking the `__wrapped__` property.
+    #
+    #   The final case of a decorator with a concrete signature but no
+    # `functools.wraps` is not detectable (by design) and is thus caveat
+    # emptor.
+    if getattr(f, "__wrapped__", None):
+        raise TypeError(textwrap.dedent("""
+            `f` cannot be decorated below `@run_in_worker` (except for
+            @staticmethod) because the extraction logic would not carry through
+            said decorator(s).
+
+            Ok:
+                @my_decorator
+                @run_in_worker()
+                def foo() -> None:
+                    ...
+
+            Not ok:
+                @run_in_worker()
+                @my_decorator
+                def foo() -> None:
+                    ...
+        """).strip())
+
+    # Dedent, as `f` may have been defined in a scoped context.
+    f_src = textwrap.dedent(inspect.getsource(f))
+
+    #   We don't want to be in the business of writing a Python parser.
+    # Fortunately our needs are relatively modest: we simply need to run a few
+    # sanity checks, and get the position where the body starts. This means
+    # that we can rely on the `ast` library to do the heavy lifting, and grab
+    # a few key values at the end.
+    f_ast = ast.parse(f_src)
+    assert len(f_ast.body) == 1
+    assert isinstance(f_ast.body[0], ast.FunctionDef)
+    assert f_ast.body[0].body
+
+    # For some reason ast one indexes lineno.
+    src_lines = f_src.splitlines(keepends=False)
+
+    node: ast.AST
+    for node in f_ast.body[0].body:
+        # In Python 3.7, there is a bug in `ast` that causes it to incorrectly
+        # report the start line of bare multi-line strings:
+        #   https://bugs.python.org/issue16806
+        # Given that the only use for such strings is a docstring (or multi
+        # line comment), we simply elect to skip over them and index on the
+        # first node that will give valid indices.
+        if node.col_offset == -1:
+            assert isinstance(node.value, ast.Str), f"Expected `ast.Str`, got {type(node)}. ({node}) {node.lineno}"
+            continue
+
+        raw_body_lines = src_lines[node.lineno - 1:]
+        col_offset = node.col_offset
+        break
+
+    else:
+        raise TypeError("Could not find valid start of body.")
+
+    body_lines: typing.List[str] = []
+    for i, l in enumerate(raw_body_lines):
+        prefix, suffix = l[:col_offset], l[col_offset:]
+
+        # The first line of the body may overlap with the signature.
+        #   e.g. `def f(): pass`
+        # For all other lines, the prefix must only be indentation.
+        assert not i or not l.strip() or not prefix.strip(), f"{l}, {col_offset}"
+
+        body_lines.append(suffix)
+    return signature, "\n".join(body_lines)
+
+
 def run_in_worker(scoped: bool = True) -> typing.Callable[..., typing.Any]:
     """Decorator to run Task method in worker rather than the caller.
 
@@ -95,48 +221,11 @@ def run_in_worker(scoped: bool = True) -> typing.Callable[..., typing.Any]:
         except AttributeError:
             pass
 
-        signature = inspect.signature(f)
-        for arg, arg_parameter in signature.parameters.items():
-            if arg_parameter.kind == inspect.Parameter.VAR_POSITIONAL:
-                raise ValueError(
-                    f"Variadic positional argument `*{arg}` not permitted "
-                    "for `run_in_worker` function.")
-            if arg_parameter.kind == inspect.Parameter.VAR_KEYWORD:
-                raise ValueError(
-                    f"Variadic keywork argument `**{arg}` not permitted "
-                    "for `run_in_worker` function.")
-            if arg_parameter.annotation == inspect.Parameter.empty:
-                raise ValueError(f"Missing type annotation for parameter `{arg}`")
-
-        if signature.return_annotation == inspect.Parameter.empty:
-            raise ValueError("Missing return annotation.")
-
+        signature, f_body = parse_f(f)
         has_return_value = (signature.return_annotation is not None)
         if has_return_value and not scoped:
-            raise ValueError(
+            raise TypeError(
                 "Unscoped (globally executed) call can not have a return value.")
-
-        f_src = textwrap.dedent(inspect.getsource(f))
-        f_ast = ast.parse(f_src)
-        assert len(f_ast.body) == 1
-        assert isinstance(f_ast.body[0], ast.FunctionDef)
-        assert f_ast.body[0].body
-
-        # For some reason ast one indexes lineno.
-        src_lines = f_src.splitlines(keepends=False)
-        raw_body_lines = src_lines[f_ast.body[0].body[0].lineno - 1:]
-        col_offset = f_ast.body[0].body[0].col_offset
-
-        body_lines: typing.List[str] = []
-        for i, l in enumerate(raw_body_lines):
-            prefix, suffix = l[:col_offset], l[col_offset:]
-
-            # The first line of the body may overlap with the signature.
-            #   e.g. `def f(): pass`
-            # For all other lines, the prefix must only be indentation.
-            assert not i or not prefix.strip()
-
-            body_lines.append(suffix)
 
         @functools.wraps(f)
         def inner(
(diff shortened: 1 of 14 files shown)


Krovatkin:
@robieta I'll try to get to it this afternoon!

components/_impl/workers/subprocess_rpc.py
```diff
@@ -0,0 +1,488 @@
...
+
+    def from_handle(handle: typing.Optional[int], mode: int) -> typing.Optional[int]:
+        return None if handle is None else msvcrt.open_osfhandle(handle, mode)
+
+else:
+    to_handle = lambda fd: fd
+    from_handle = lambda fd, _: fd
+
+
+class _TimeoutPIPE:
```
bitfort, line 75: This is somewhat complicated, I suspect if you wanted you could punt most of this complexity to using the thread safe queue.Queue; the singleton becomes a single consumer and you can multiple producers. But the solution you have here seems to work, and seems to have some extra features like deduplication.

components/_impl/workers/subprocess_rpc.py
```diff
@@ -0,0 +1,488 @@
...
+    def __del__(self) -> None:
+        if self._owns_pipe:
+            self._close_fds()
+
+
+# =============================================================================
+# == Exception Propagation  ===================================================
+# =============================================================================
+
+class ExceptionUnpickler(pickle.Unpickler):
```
bitfort, line 246: Great eye for detail. I would have just throwing "SubprocessException(Foo)" and made people unpack it.

bitfort: This is true dedication to transparency.

components/_impl/workers/subprocess_rpc.py
```diff
@@ -0,0 +1,488 @@
...
+            singleton._active_reads[w_fd] = (timeout, time.time())
+
+        try:
+            yield
+
+        finally:
+            singleton.pop(w_fd)
+
+
+class Pipe:
```
bitfort, line 164: I suspect much of this could be simplified using some combination of select / poll / peek:

https://man7.org/linux/man-pages/man2/poll.2.html

Such as:
https://docs.python.org/3/library/select.html

bitfort: This is assuming your main concern is blocking on an empty pipe.

bitfort (commented):
I finished my first half of the review, coming back for the second tomorrow.

components/_impl/workers/in_process_worker.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def __init__(self, globals: typing.Dict[str, typing.Any]):
+        super().__init__()
+        self._globals: typing.Dict[str, typing.Any] = globals
+
+    @property
+    def in_process(self) -> bool:
+        return True
+
+    def run(self, snippet: str) -> None:
```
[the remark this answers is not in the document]
robieta: I guess the idea is that it would look something like: (?)

```
worker = Worker()

@worker.dispatch
def f(x: int, y: int) -> int:
  return x + y
  
z = f(x=1, y=2)
```

That's kind of what `run_in_worker` does now. It's not clear to me how you would handle state between calls. (Could a later dispatch access `x` and `y`?) And I'm worried about both the cognitive and runtime overhead of such a scheme. Maybe something like: (?)
```
def f_store(x: int) -> None:
  pass

def add() -> None:
  # y cannot cross the worker boundary
  y = torch.ones((1,)) + x
  
def f_load() -> int:
  return int(y)

worker = Worker()
worker.run(f_store, x=1)
worker.run(add)
y = worker.run(f_load)
```

Part of the reason for the strong `store`, `run`, `load` separation is that in most cases we expect that we're dealing with objects that we can't pull back into the parent process. So the majority of calls are just manipulating the env, and then occasionally we grab a result. Two things do strike me though, as I think through this again:
1) Maybe `run_in_worker` should be `worker.dispatch`. I feel like I may have already made what you're asking for, but I'm exposing it at the wrong level. (I still think there are cases where the simpler primitives are better just for comprehensibility.)

2) Maybe `store` and `load` should be implemented with `@property` and `.setter`. Instead of:
```
worker.store("x", 1)
worker.run("y = x + 1")
worker.load("y")
```
it's much more Pythonic to write something like:
```
worker.__globals__.x = 1
worker.run("y = x + 1")
y = worker.__globals__.y
```
The key detail that the latter obscures though is that there's a copy, so:
```
worker.__globals__.x = {}
worker.__globals__.x["foo"] = "bar"  # This runs, but does not do what you intend.
```

To wrap up this kind of rambling defense of my beautiful API: it's as pared down as I think it can be before it starts to become confusing (by papering over details of the caller/worker divide), but definitely open to discussion.

components/_impl/workers/subprocess_rpc.py
robieta: I experimented with putting the read on a side thread, and then timing out communication with that thread. (Am I understanding correctly that that's the gist of what you're getting at?) The issue was that it's horribly slow. (About 1000x slower than keeping the `os.read` on the main thread, and unacceptably high based on my discussions with Eric.)

components/_impl/workers/subprocess_rpc.py
robieta: Remember it has to be cross platform, and Windows has a pretty different I/O model. (In particular, peeks / non-blocking reads aren't really a thing.) And I'd rather not have a polling loop going if 99.9% of the time I can rely on OS interrupts.

I do share your concern with the Pipe machinery in general. I'd love to brainstorm if there's a simpler and better way to do it.

components/_impl/tasks/base.py
```diff
@@ -0,0 +1,274 @@
...
+import functools
+import inspect
+import marshal
+import textwrap
+import typing
+
+from components._impl.workers import base
+
+
+class TaskBase(abc.ABC):
```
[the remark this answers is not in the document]
robieta: I think you may have misread `worker.in_process` as `worker.in_progress`

robieta:
@eircfb I changed ModelTask so only one could exist at a time. I had to make some tweaks to the tests since we can no longer inspect ahead of time, but nothing major. The most significant is that we can't contextually add "-freeze" without benchmarks taking a minute to setup, which doesn't seem worth it.

components/_impl/workers/subprocess_worker.py
```diff
@@ -0,0 +1,346 @@
...
+                sys.exit(1)
+        """).strip()
+
+        if self._proc.poll() is not None:
+            raise ValueError("Process has already exited.")
+
+        proc_stdin = self._proc.stdin
+        assert proc_stdin is not None
+
+        self._log_cmd(bootstrap_command)
```
bitfort, line 212: I'm trying to understand why _log_cmd -> read doesn't go through a common shared functionality with _run?

components/_impl/workers/subprocess_worker.py
```diff
@@ -0,0 +1,346 @@
...
+                extra_context=(
+                    f"    working_dir: {self.working_dir}\n"
+                    f"    stdout:\n{textwrap.indent(stdout, ' ' * 8)}\n\n"
+                    f"    stderr:\n{textwrap.indent(stderr, ' ' * 8)}"
+                )
+            )
+
+    def _kill_proc(self) -> None:
+        """Best effort to kill subprocess."""
+        if getattr(self, "_proc", None) is None:
```
bitfort, line 302: Can we just set self._proc = None at the start of the ctor? Seems more intuitive.

torchbenchmark/__init__.py
```diff
@@ -103,6 +111,230 @@ def setup(verbose: bool = True, continue_on_fail: bool = False) -> bool:
...
+    _diagnostic_msg: str
+
+    @property
+    def name(self) -> str:
+        return os.path.basename(self.path)
+
+
+class ModelTask(base_task.TaskBase):
+
+    # The worker may (and often does) consume significant system resources.
```
bitfort, line 141: I am curious about this approach of running multiple torchbench tasks at the same time. There is fear in me that it would interfere with the results. But there is no data I have to support this fear - not that emotions need be rational :)

Can you share more as to the impact you think parallelization of benchmarks might have? Do we have any early numbers?

bitfort (approved):
Approved contingent on CI and comments resolved.

components/_impl/tasks/base.py
```diff
@@ -0,0 +1,274 @@
+import abc
```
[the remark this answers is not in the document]
robieta: Done.

components/_impl/tasks/base.py
```diff
@@ -0,0 +1,274 @@
...
+import marshal
+import textwrap
+import typing
+
+from components._impl.workers import base
+
+
+class TaskBase(abc.ABC):
+
+    @abc.abstractproperty
```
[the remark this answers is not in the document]
robieta: Thanks!

components/_impl/tasks/base.py
```diff
@@ -0,0 +1,274 @@
...
+    # `functools.wraps`, it is HIGHLY likely that it uses variadic arguments
+    # so that it can forward them. (And thus will fail the signature checks)
+    # above. If we are passed a function which is decorated with a "proper"
+    # decorator, we catch it here by checking the `__wrapped__` property.
+    #
+    #   The final case of a decorator with a concrete signature but no
+    # `functools.wraps` is not detectable (by design) and is thus caveat
+    # emptor.
+    if getattr(f, "__wrapped__", None):
+        raise TypeError(textwrap.dedent("""
```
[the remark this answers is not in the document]
robieta: Thanks!

components/_impl/workers/base.py
```diff
@@ -0,0 +1,99 @@
+import abc
+import ast
+import typing
+
+
+class WorkerBase(abc.ABC):
```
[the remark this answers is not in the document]
robieta: I started writing something out, but wound up just referring folks to `in_process_worker.InProcessWorker`. (I think your experience of understanding the semantics by just looking at a simple implementation will be pretty common.)

components/_impl/workers/in_process_worker.py
```diff
@@ -0,0 +1,49 @@
+import marshal
+import textwrap
+import typing
+
+from components._impl.workers import base
+
+
+class InProcessWorker(base.WorkerBase):
```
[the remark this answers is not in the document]
robieta: I'm glad.

components/_impl/workers/in_process_worker.py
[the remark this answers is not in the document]
robieta: After some offline discussion, we decided to leave as-is.

components/_impl/workers/subprocess_worker.py
robieta: The hang up is that `_run` requires that the work loop is already running on the worker, so we can't use it in the bootstrapping phase.

components/_impl/workers/subprocess_worker.py
robieta: Ah, but then it would be an `Optional[subprocess.Popen]` and we'd have to none check everywhere when it goes back to Pytorch proper and has to pass the type checker...

I would consider this getattr pattern at least as Pythonic as setting a default at the start of the ctor.

torchbenchmark/__init__.py
robieta: It's mostly more for the context of ubenchmarking, where you're mostly running single core (or maybe a couple cores) and you can strongly isolate the work units. For full model benchmarks it's less likely to be applicable. And as you say, there are resource contention concerns.

components/_impl/workers/subprocess_rpc.py
robieta: After offline discussion, we decided to keep this. (And if we think of a better way later it's not hard to swap.)

components/_impl/workers/subprocess_rpc.py
```diff
@@ -0,0 +1,488 @@
...
+    will not work because they rely on the awaited function returning control
+    to the event loop. An alternate formulation uses `run_in_executor` and
+    `asyncio.wait`, which places the read on a side thread under the hood.
+    However this is also not suitable, because:
+
+        1)  This additional machinery increases the cost when data is already
+            present in the Pipe (most common case) ~1000x, from O(us) to O(ms)
+        2)  We have to poll the future, which wastes the awaitable nature `read`
+
+    Instead of trying to interrupt the pipe read, we can cause it terminate by
```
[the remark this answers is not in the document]
robieta: Done.

components/_impl/workers/in_process_worker.py
```diff
@@ -0,0 +1,49 @@
...
+        )
+        exec(code, self._globals)  # noqa: P204
+
+    # Serialize and deserialize during store and load to match the behavior of
+    # workers with `in_process=False`.
+    def store(self, name: str, value: typing.Any, in_memory: bool = False) -> None:
+        if not in_memory:
+            value = marshal.loads(marshal.dumps(value))
+
+        self._globals[name] = value
```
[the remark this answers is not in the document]
robieta: :) A sufficiently astute observer in indistinguishable from good documentation.
- Arthur C. Clark (probably)

components/_impl/tasks/base.py
```diff
@@ -0,0 +1,274 @@
...
+
+class TaskBase(abc.ABC):
+
+    @abc.abstractproperty
+    def worker(self) -> base.WorkerBase:
+        ...
+
+
+def parse_f(f: typing.Callable) -> typing.Tuple[inspect.Signature, str]:
+    """Handle source code extraction for `run_in_worker`."""
```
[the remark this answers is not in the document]
robieta: Done.

components/_impl/workers/subprocess_rpc.py
robieta: Thanks! The rethrowing might be my favorite part of the entire PR.

robieta: Thanks! The rethrowing might be my favorite part of the entire PR.

components/_impl/workers/subprocess_worker.py
```diff
@@ -0,0 +1,346 @@
...
+            def args(self) -> typing.List[str]:
+                return ["taskset", "--cpu-list", "0"] + super().args
+    ```
+    """
+
+    _working_dir: str
+    _alive: bool = False
+    _bootstrap_timeout: int = 10  # seconds
+
+    def __init__(self, timeout: typing.Optional[float] = None) -> None:
```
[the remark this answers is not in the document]
robieta: In Python the convention is to use `timeout` specifically. (see asyncio, threading, subprocess, etc.)

PR #425, opened 2021-08-17
Address broken CUDA cases in TorchBench

robieta:
There are two issues with TorchBench CUDA right now:
1) YoloV3 is misconfigured, and is failing for a simple reason.
2) hf_DistilBert is failing and contaminating the GPU.

(1) is easy to solve, and I do so here. (2) needs to be diagnosed, but this should restore most of the models while we do. (Process isolation will make it moot, but it's still WIP)


bitfort (approved):
Exciting!!

PR #430, opened 2021-08-24
Use taskset when GOMP_CPU_AFFINITY is set

robieta:
Turns out if we don't do this, TorchBench CI will run with a single CPU thread.

### DIFF
--- a/torchbenchmark/__init__.py
+++ b/torchbenchmark/__init__.py
@@ -136,6 +136,28 @@ class ModelDetails:
         return os.path.basename(self.path)
 
 
+class Worker(subprocess_worker.SubprocessWorker):
+    """Run subprocess using taskset if CPU affinity is set.
+
+    When GOMP_CPU_AFFINITY is set, importing `torch` in the main process has
+    the very surprising effect of changing the threading behavior in the
+    subprocess. (See https://github.com/pytorch/pytorch/issues/49971 for
+    details.) This is a problem, because it means that the worker is not
+    hermetic and also tends to force the subprocess torch to run in single
+    threaded mode which drastically skews results.
+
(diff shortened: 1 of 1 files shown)


torchbenchmark/__init__.py
```diff
@@ -136,6 +136,28 @@ def name(self) -> str:
         return os.path.basename(self.path)
 
 
+class Worker(subprocess_worker.SubprocessWorker):
+    """Run subprocess using taskset if CPU affinity is set.
+
+    When GOMP_CPU_AFFINITY is set, importing `torch` in the main process has
+    the very surprising effect of changing the threading behavior in the
+    subprocess. (See https://github.com/pytorch/pytorch/issues/49971 for
```
xuzhao9, line 144: Maybe it is also worth mentioning using "torch.set_num_threads()" won't saturate the CPU resources?

torchbenchmark/__init__.py
```diff
@@ -136,6 +136,28 @@ def name(self) -> str:
         return os.path.basename(self.path)
 
 
+class Worker(subprocess_worker.SubprocessWorker):
+    """Run subprocess using taskset if CPU affinity is set.
```
xuzhao9, line 140: This is a LONG docstring! Thanks for the clarification.

xuzhao9 (approved):
Nicely done!

torchbenchmark/__init__.py
robieta: Oh, that's a great point!

torchbenchmark/__init__.py
robieta: And about to get a bit longer.

PR #433, opened 2021-08-30
Merge `torch.jit.script` and `torch.jit._script_pdt` API.

xuzhao9:
`torch.jit._script_pdt` API is removed upstream in this PR:
https://github.com/pytorch/pytorch/commit/510d2ece81c18ba3f9bbcbc509effe281a77206c


robieta (approved):
LGTM. Thanks Xu!

xuzhao9:
Need to add `import torch` even if torch is not used, because by default the JIT compiler will assume torch is already imported as torch and use symbols such as `torch.Tensor` directly. If torch is not imported, JIT compiler will complain `torch.Tensor` symbol is undefined.

PR #434, opened 2021-08-30
BERT_pytorch: Increase corpus and batch for adequate gpu utilization

eircfb:
Uses parameterized generated random corpus instead of giant file. Profile corresponds with running actual 620mb trained model.

<eval, train>
On gpu: <30ms, 150ms>
On cpu: <4.1s, 8.3s>

### DIFF
--- a/torchbenchmark/models/BERT_pytorch/__init__.py
+++ b/torchbenchmark/models/BERT_pytorch/__init__.py
@@ -10,18 +10,63 @@ from .bert_pytorch.dataset import BERTDataset, WordVocab
 from .bert_pytorch.model import BERT
 from torch.utils.data import DataLoader
 
-torch.manual_seed(1337)
-random.seed(1337)
-np.random.seed(1337)
-torch.backends.cudnn.deterministic = True
+torch.backends.cudnn.deterministic = False
 torch.backends.cudnn.benchmark = False
 from pathlib import Path
 from ...util.model import BenchmarkModel
 from torchbenchmark.tasks import NLP
+
+import io
+
+class CorpusGenerator(io.TextIOBase):
+
+    def __init__(self, words, lines):
+        self.lines_read = 0
+        self.lines = lines
+        self.words = words
+
+    def reset(self):
+        self.lines_read = 0
+
+    def readable(self):
+        if (self.lines_read > self.lines):
+            return False
+        return True
+
+    def readline(self):
+
+        self.lines_read = self.lines_read + 1
+        if (self.lines_read > self.lines):
+          return ""
+
+        newline = ""
+        for j in range(random.randrange(1,4)):
+            newline += str(random.randrange(self.words)) + " "
+
+        newline += "\\t "
+
+        for j in range(random.randrange(1,4)):
+            newline += str(random.randrange(self.words)) + " "
+
+        newline += "\n"
+
+        #print(newline)
+
+        return newline
+
+
+
 class Model(BenchmarkModel):
     task = NLP.LANGUAGE_MODELING
+
+    def reset_seeds(self, seed):
+        torch.manual_seed(seed)
+        random.seed(seed)
+        np.random.seed(seed)
+
     def __init__(self, device=None, jit=False):
         super().__init__()
+
         self.device = device
         self.jit = jit
         root = str(Path(__file__).parent)
@@ -33,12 +78,54 @@ class Model(BenchmarkModel):
         ]) # Avoid reading sys.argv here
         args.with_cuda = self.device == 'cuda'
         args.script = self.jit
-        vocab = WordVocab.load_vocab(args.vocab_path)
 
+        # Example effect of batch size on eval time(ms)
+        # bs     cpu       cuda
+        # 1      330       15.5
+        # 2      660       15.5
+        # 4     1200       15.2
+        # 8     2200       20
+        # 16    4350       33
+        # 32    8000       58
+        #
+        # Issue is that with small batch sizes the gpu is starved for work.
+        # Ideally doubling work would double execution time.
+
+        # parameters for work size
+        args.batch_size = 16
+        vocab_size = 20000
+        args.corpus_lines = 50000
+
+        # generate random corpus from parameters
+        self.reset_seeds(1337)
+        vocab = WordVocab(CorpusGenerator(vocab_size, args.corpus_lines))
+
+        #with open(args.train_dataset, "r", encoding="utf-8") as f:
+        #  vocab = WordVocab(f)
+        #vocab = WordVocab.load_vocab(args.vocab_path)
+
+        if False:
+            print("seq_len:")
+            print(args.seq_len)
+            print("batch size:")
+            print(args.batch_size)
+            print("layers")
+            print(args.layers)
+            print("args hidden:")
+            print(args.hidden)
+            print("len vocab:")
+            print(len(vocab))
+            print(type(vocab))
+
+        self.reset_seeds(1337)
         train_dataset = BERTDataset(args.train_dataset, vocab, seq_len=args.seq_len,
-                                    corpus_lines=args.corpus_lines, on_memory=args.on_memory)
-        test_dataset = BERTDataset(args.test_dataset, vocab, seq_len=args.seq_len, on_memory=args.on_memory) \
+                                    corpus_lines=args.corpus_lines, on_memory=args.on_memory, generator = CorpusGenerator(vocab_size, args.corpus_lines))
+
+        self.reset_seeds(1337)
+        test_dataset = BERTDataset(args.test_dataset, vocab, seq_len=args.seq_len, on_memory=args.on_memory, generator = CorpusGenerator(vocab_size, args.corpus_lines)) \
             if args.test_dataset is not None else None
+
+        self.reset_seeds(1337)
 
         train_data_loader = DataLoader(train_dataset, batch_size=args.batch_size, num_workers=args.num_workers)
         test_data_loader = DataLoader(test_dataset, batch_size=args.batch_size, num_workers=args.num_workers) \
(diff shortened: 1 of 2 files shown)


torchbenchmark/models/BERT_pytorch/bert_pytorch/dataset/dataset.py
```diff
@@ -16,7 +16,8 @@ def __init__(self, corpus_path, vocab, seq_len, encoding="utf-8", corpus_lines=N
         self.corpus_path = corpus_path
         self.encoding = encoding
 
-        with open(corpus_path, "r", encoding=encoding) as f:
+        with generator as f:
```
xuzhao9, line 19: Does `tqdm.tqdm()` correctly handle the case where generator is `None`? If not, please add an assert here that `f` is not `None`.

torchbenchmark/models/BERT_pytorch/__init__.py
```diff
@@ -10,18 +10,63 @@
 from .bert_pytorch.model import BERT
 from torch.utils.data import DataLoader
 
-torch.manual_seed(1337)
-random.seed(1337)
-np.random.seed(1337)
-torch.backends.cudnn.deterministic = True
+torch.backends.cudnn.deterministic = False
```
xuzhao9, line 13: I guess setting `cudnn.deterministic` to `False` helps improve GPU utilization numbers? How much does it help?

torchbenchmark/models/BERT_pytorch/__init__.py
```diff
@@ -10,18 +10,63 @@
...
+
+    def reset(self):
+        self.lines_read = 0
+
+    def readable(self):
+        if (self.lines_read > self.lines):
+            return False
+        return True
+
+    def readline(self):
```
xuzhao9, line 36: Maybe remove empty lines in this function to make it more compact? Also, it will be helpful to add comment on what each line looks like.

torchbenchmark/models/BERT_pytorch/__init__.py
```diff
@@ -33,13 +78,55 @@ def __init__(self, device=None, jit=False):
...
+
+        # generate random corpus from parameters
+        self.reset_seeds(1337)
+        vocab = WordVocab(CorpusGenerator(vocab_size, args.corpus_lines))
+
+        #with open(args.train_dataset, "r", encoding="utf-8") as f:
+        #  vocab = WordVocab(f)
+        #vocab = WordVocab.load_vocab(args.vocab_path)
+
+        if False:
```
xuzhao9, line 107: Maybe add a variable `debug = False` and `if debug` here.

torchbenchmark/models/BERT_pytorch/__init__.py
```diff
@@ -33,13 +78,55 @@ def __init__(self, device=None, jit=False):
...
+            print(args.batch_size)
+            print("layers")
+            print(args.layers)
+            print("args hidden:")
+            print(args.hidden)
+            print("len vocab:")
+            print(len(vocab))
+            print(type(vocab))
+
+        self.reset_seeds(1337)
```
xuzhao9, line 120: Why reset_seeds multiple times here and on line 100, 124?

torchbenchmark/models/BERT_pytorch/__init__.py
```diff
@@ -50,13 +137,14 @@ def __init__(self, device=None, jit=False):
                                    lr=args.lr, betas=(args.adam_beta1, args.adam_beta2), weight_decay=args.adam_weight_decay,
                                    with_cuda=args.with_cuda, cuda_devices=args.cuda_devices, log_freq=args.log_freq, debug=args.debug)
 
-        INFERENCE_BATCH_SIZE = 8
+        INFERENCE_BATCH_SIZE = args.batch_size
         example_batch = next(iter(train_data_loader))
-        self.example_inputs = example_batch['bert_input'].to(self.device), example_batch['segment_label'].to(self.device)[:INFERENCE_BATCH_SIZE]
-        self.is_next = example_batch['is_next'].to(self.device)
-        self.bert_label = example_batch['bert_label'].to(self.device)
+        self.example_inputs = example_batch['bert_input'].to(self.device)[:INFERENCE_BATCH_SIZE], example_batch['segment_label'].to(self.device)[:INFERENCE_BATCH_SIZE]
```
xuzhao9, line 142: Sorry I don't quite understand this line, does that mean we are truncating the input to `INFERENCE_BATCH_SIZE`? What is the reason behind that?

torchbenchmark/models/BERT_pytorch/__init__.py
```diff
@@ -33,13 +78,55 @@ def __init__(self, device=None, jit=False):
...
+            print(args.hidden)
+            print("len vocab:")
+            print(len(vocab))
+            print(type(vocab))
+
+        self.reset_seeds(1337)
         train_dataset = BERTDataset(args.train_dataset, vocab, seq_len=args.seq_len,
-                                    corpus_lines=args.corpus_lines, on_memory=args.on_memory)
-        test_dataset = BERTDataset(args.test_dataset, vocab, seq_len=args.seq_len, on_memory=args.on_memory) \
+                                    corpus_lines=args.corpus_lines, on_memory=args.on_memory, generator = CorpusGenerator(vocab_size, args.corpus_lines))
```
xuzhao9, line 122: I don't see `args.on_memory` is set anywhere in `__init__()`, maybe we explicitly set `args.on_memory=False`?

torchbenchmark/models/BERT_pytorch/__init__.py
```diff
@@ -10,18 +10,63 @@
...
+torch.backends.cudnn.deterministic = False
 torch.backends.cudnn.benchmark = False
 from pathlib import Path
 from ...util.model import BenchmarkModel
 from torchbenchmark.tasks import NLP
+
+import io
+
+class CorpusGenerator(io.TextIOBase):
+
```
xuzhao9, line 22: Please add a docstring here explain the usage of this class.

torchbenchmark/models/BERT_pytorch/bert_pytorch/dataset/dataset.py
```diff
@@ -16,7 +16,8 @@ def __init__(self, corpus_path, vocab, seq_len, encoding="utf-8", corpus_lines=N
...
+        #with open(corpus_path, "r", encoding=encoding) as f:
             if self.corpus_lines is None and not on_memory:
                 for _ in tqdm.tqdm(f, desc="Loading Dataset", total=corpus_lines, disable=QUIET):
```
xuzhao9, line 22: Sorry I don't quite get this line and line 122. On line 122, `on_memory` seems to be `False` and `copus_lines=50000`. Therefore, both this and the [next](https://github.com/pytorch/benchmark/blob/c60023d39556ee1d7e2bddde346309068626e02e/torchbenchmark/models/BERT_pytorch/bert_pytorch/dataset/dataset.py#L25) code block will be skipped. Is this the correct behavior?

xuzhao9 (changes requested):
Thanks for the PR! Sorry I don't quite understand the details in a few places, therefore please address my concerns and I will review it again, thanks!

torchbenchmark/models/BERT_pytorch/__init__.py
eircfb: I thought we agreed that deterministic was only to be used when there was a reason. We want the cudnn to be able to run in its best mode which is unconstrained.

torchbenchmark/models/BERT_pytorch/__init__.py
eircfb: Its so when data size changes the init of earlier dataset wont change the first part of next dataset. Keeps distribution the same. It was useful when developing the dataset and setting the parameters.

torchbenchmark/models/BERT_pytorch/__init__.py
eircfb: Yes, setting inference data to a smaller batch size. There is large training data and small inference data. We want batch size to be as small as possible because in real life it decreases latency, but needs to be bigger than 1 for sufficient utilization.

torchbenchmark/models/BERT_pytorch/__init__.py
eircfb: on_memory was supposed to be true! Not sure why it was working, some python default behavior? Anyway needs to be set to true.

torchbenchmark/models/BERT_pytorch/bert_pytorch/dataset/dataset.py
eircfb: Like above, on_memory is supposed to be true.

torchbenchmark/models/BERT_pytorch/__init__.py
xuzhao9: Gotcha!

torchbenchmark/models/BERT_pytorch/__init__.py
xuzhao9: Ahh, that makes more sense. Thanks!

xuzhao9 (approved):
LGTM!

torchbenchmark/models/BERT_pytorch/__init__.py
xuzhao9: Makes sense. I guess setting the batch size to 16 means when the batch size is 16 or larger, the job latency starts to increase.

eircfb:
> LGTM!
> One question: you mentioned that you are replacing the giant train file with generated data. I guess we can now remove that input file from the repository?

The corpus for this test was 58 bytes, now unreferenced. I don't think it does any harm to have a small example of the text input.

wconstab:
Sorry I didn't see this sooner.  I think it's a cool idea, we've talked about generating random data before, although if you do this it's important to make sure the statistics of the data are representative of the real data.  This matters more for longer-running benchmarks though.

Btw, if you are trying to cut down on the repo size, i'm afraid you'll have to do other stuff to actually prune out the file, or it will still be 'reachable' by historical git commits.

I'm also not sure I followed the connection between doing the generation of data, and increasing GPU util.  It should have been possible to use a larger sample of the 'real' data file to reach arbitrarily large batch sizes, right?

eircfb:
> Sorry I didn't see this sooner. I think it's a cool idea, we've talked about generating random data before, although if you do this it's important to make sure the statistics of the data are representative of the real data. This matters more for longer-running benchmarks though.
> 
> Btw, if you are trying to cut down on the repo size, i'm afraid you'll have to do other stuff to actually prune out the file, or it will still be 'reachable' by historical git commits.
> 
> I'm also not sure I followed the connection between doing the generation of data, and increasing GPU util. It should have been possible to use a larger sample of the 'real' data file to reach arbitrarily large batch sizes, right?

The purpose of this fix is to simply have an authentic load on bert. Previously utilization was impossibly low.

I started the redo with an actual 650mb corpus, then generated sentence pairs with similar vocab size and distribution. There's just no point using an actual corpus since we aren't testing correctness. In this case its not sufficient to simply increase batch size, there's no getting around vocab size of 16 when it should be 40k.

Cut down repo size: the current corpus is 2 lines, 2 sentence pairs each. 56 bytes.

wconstab:
Ah I see- i missed that you changed vocab not seq_len; and I was expecting to see you delete an existing data file from the repo, but I guess you were just avoiding adding a new one.  So, thats great.  

One point would be to use an actual vocab_size, batch_size, seq_len combination from a common academic paper, just so that our benchmarks stay 'comparable' to stuff other people are benchmarking.  There should be plenty of such configurations for BERT models that are large like the sizes you picked.

eircfb:
> Ah I see- i missed that you changed vocab not seq_len; and I was expecting to see you delete an existing data file from the repo, but I guess you were just avoiding adding a new one. So, thats great.
> 
> One point would be to use an actual vocab_size, batch_size, seq_len combination from a common academic paper, just so that our benchmarks stay 'comparable' to stuff other people are benchmarking. There should be plenty of such configurations for BERT models that are large like the sizes you picked.

Vocab and corpus sizes: I started with a common english to german corpus thats used in transformer papers. 450k lines and 45k words. I played with smaller sizes and found that 50k and 20k gave a cpu speedup without varying the runtime profile %.

I think its ok to use these smaller sizes, increasing them costs test throughput and I don't think it gains us anything.

PR #435, opened 2021-08-30
change resnet50 benchmark batch size from 1 to 32

erichan1:
Based on analysis of gpu throughput against batch size, change resnet50 benchmark batch size to 32 to better utilize gpu. 

Edit: This was Will's aws node and I forgot to change the git credentials to be mine. Don't think I can change the commit now so will just submit this.

### DIFF
--- a/torchbenchmark/models/resnet50/__init__.py
+++ b/torchbenchmark/models/resnet50/__init__.py
@@ -54,7 +54,7 @@ class Model(BenchmarkModel):
     def eval(self, niter=1):
         model = self.eval_model
         example_inputs = self.example_inputs
-        example_inputs = example_inputs[0][0].unsqueeze(0)
+        example_inputs = example_inputs[0]
         for i in range(niter):
             model(example_inputs)
 


bitfort (approved):
Great!

aaronenyeshi:
Hi @erichan1 , I noticed that this changed file is an auto-generated file where the generator is at `torchbenchmark/models/gen_torchvision_benchmarks.py`. And this line in particular may get introduced again if the generator is run, see: https://github.com/pytorch/benchmark/blob/master/torchbenchmark/models/gen_torchvision_benchmarks.py#L11
Snippet:
```
class_models = ['alexnet', 'vgg16', 'resnet18', 'resnet50', 'squeezenet1_1', 'densenet121', 'mobilenet_v2', 'mobilenet_v3_large', 'shufflenet_v2_x1_0', 'resnext50_32x4d', 'mnasnet1_0']
for class_model in class_models:
    folder = Path(class_model)
    if not os.path.isdir(folder):
        os.makedirs(folder)
    input_shape = (32, 3, 224, 224)
    example_inputs = f'(torch.randn({input_shape}).to(self.device),)'
    eval_inputs = 'example_inputs[0][0].unsqueeze(0)'
    init_program = f"""
```

Do we still use those gen_* scripts to create these `__init__.py` files? And should we remove `[0].unsqueeze(0)` from all the other models in `class_models`? + @xuzhao9 @robieta for that question too.

erichan1:
@aaronenyeshi You're right. Removing [0].unsqueeze(0) would make all the model benchmarks have batch size of 32 rather than 1. Eventually I'd guess we want it to be customizable by model since batch size is pretty important, but having a uniform way might be better for now. Thoughts @bitfort? I think making all batch sizes 32 is overall better than having them be 1.

PR #440, opened 2021-09-01
Disable hf_DistilBert model cuda tests because of GPU contamination 

xuzhao9:
hf_DistilBert CUDA test is disabled in PR#425 due to GPU contamination. We need to disable it in the v1 benchmark suite as well.


robieta (approved):
Oops, sorry about that. (Still getting up to speed on all the components.) Thanks!

PR #432, opened 2021-08-27
Update README.md for typos and code blocks

aaronenyeshi:
Small nitpick changes to README.md to improve this body of text for readers.

PR #443, opened 2021-09-01
Add test to verify models and inputs are using the expected device

aaronenyeshi:
The device attribute needs to be set in three locations: the BenchmarkModel,
the nn.Module model, and the model inputs. This test will verify that
the tensors and inputs are on the expected device set in BenchmarkModel.
Also, fix a few benchmark bugs where the device is incorrectly set.

### DIFF
--- a/test.py
+++ b/test.py
@@ -83,10 +83,21 @@ def _load_test(path, device):
             except NotImplementedError:
                 self.skipTest('Method eval is not implemented, skipping...')
 
+    def check_device_fn(self):
+        task = ModelTask(path, timeout=TIMEOUT)
+        with task.watch_cuda_memory(skip=(device != "cuda"), assert_equal=self.assertEqual):
+            try:
+                task.make_model_instance(device=device, jit=False)
+                task.check_device()
+                task.del_model_instance()
+            except NotImplementedError:
+                self.skipTest('Method check_device is not implemented, skipping...')
+
     name = os.path.basename(path)
     setattr(TestBenchmark, f'test_{name}_example_{device}', example)
     setattr(TestBenchmark, f'test_{name}_train_{device}', train)
     setattr(TestBenchmark, f'test_{name}_eval_{device}', eval_fn)
+    setattr(TestBenchmark, f'test_{name}_check_device_{device}', check_device_fn)
 
 
 def _load_tests():
--- a/torchbenchmark/__init__.py
+++ b/torchbenchmark/__init__.py
@@ -319,33 +319,43 @@ class ModelTask(base_task.TaskBase):
          # Check this BenchmarkModel has a device attribute.
         current_device = getattr(instance, 'device', None)
         if current_device is None:
-            raise Exception('Missing device in BenchmarkModel.')
+            raise RuntimeError('Missing device in BenchmarkModel.')
 
         model, inputs = instance.get_module()
-        model_name = model.name if hasattr(model, 'name') else None
+        model_name = getattr(model, 'name', None)
 
         # Check the model tensors are assigned to the expected device.
-        for t in list(model.parameters()):
+        for t in model.parameters():
             model_device = t.device.type
             if model_device != current_device:
-                raise Exception(f'Model {model_name} was not set to the' \
-                                f' expected device {current_device},' \
-                                f' found device {model_device}.')
+                raise RuntimeError(f'Model {model_name} was not set to the' \
+                                   f' expected device {current_device},' \
+                                   f' found device {model_device}.')
 
         # Check the inputs are assigned to the expected device.
-        if isinstance(inputs, dict):
-            # Huggingface models do not have a Key 0
-            inputs_device = next(iter(inputs.values())).device.type
-        elif isinstance(inputs, tuple):
-            # Some inputs are nested inside tuples, such as tacotron2
-            inputs_device = inputs[0][0].device.type
-        else:
-            inputs_device = inputs[0].device.type
+        def check_inputs(inputs):
+            if isinstance(inputs, torch.Tensor):
+                if inputs.dim() and current_device == "cuda":
+                    # Zero dim Tensors (Scalars) can be captured by CUDA
+                    # kernels and need not match device.
+                    return
 
-        if inputs_device != current_device:
-            raise Exception(f'Model {model_name} inputs were not set to the' \
-                            f' expected device {current_device},' \
-                            f' found device {inputs_device}.')
+                inputs_device = inputs[0].device.type
+                if inputs_device != current_device:
+                    raise RuntimeError(f'Model {model_name} inputs were' \
+                                       f' not set to the expected device' \
+                                       f' {current_device}, found device' \
+                                       f' {inputs_device}.')
+            elif isinstance(inputs, tuple):
+                # Some inputs are nested inside tuples, such as tacotron2
+                for i in inputs:
+                    check_inputs(i)
+            elif isinstance(inputs, dict):
+                # Huggingface models take inputs as kwargs
+                for i in inputs.values():
+                    check_inputs(i)
+
+        check_inputs(inputs)
 
     # =========================================================================
     # == Control `torch` state (in the subprocess) ============================
--- a/torchbenchmark/__init__.py
+++ b/torchbenchmark/__init__.py
@@ -311,6 +311,42 @@ class ModelTask(base_task.TaskBase):
         else:
             module(*example_inputs)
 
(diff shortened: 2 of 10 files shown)


aaronenyeshi:
Verified this test with `gpurun python test.py --verbose -k check_device` and all 52 models on cpu/cuda `check_device` tests passed:
```
Ran 104 tests in 584.110s

OK (skipped=8)
```

test.py
```diff
@@ -83,10 +83,21 @@ def eval_fn(self):
             except NotImplementedError:
                 self.skipTest('Method eval is not implemented, skipping...')
 
+    def check_device_fn(self):
+        task = ModelTask(path, timeout=TIMEOUT)
+        with task.watch_cuda_memory(skip=(device != "cuda"), assert_equal=self.assertEqual):
```
xuzhao9, line 88: I guess this means we only run the check_device test when the device is cuda and skip the test when device is CPU?

torchbenchmark/models/Super_SloMo/__init__.py
```diff
@@ -55,7 +55,7 @@ def __init__(self, device=None, jit=False):
         trainData = (frame0.to(device),
                      frameT.to(device),
                      frame1.to(device))
-        self.example_inputs = trainFrameIndex, *trainData
+        self.example_inputs = trainFrameIndex.to(self.device), *trainData
```
xuzhao9, line 58: Nice catch!

xuzhao9 (approved):
Overall looks good to me! Great to see the test is already capturing a few bugs!

test.py
aaronenyeshi: This test should also run for CPU. I think `watch_cuda_memory` is to check for memory leaks when running on `device == cuda`, otherwise `watch_cuda_memory` is a no-op.

torchbenchmark/__init__.py
```diff
@@ -311,6 +311,42 @@ def check_example() -> None:
...
+            raise Exception('Missing device in BenchmarkModel.')
+
+        model, inputs = instance.get_module()
+        model_name = model.name if hasattr(model, 'name') else None
+
+        # Check the model tensors are assigned to the expected device.
+        for t in list(model.parameters()):
+            model_device = t.device.type
+            if model_device != current_device:
+                raise Exception(f'Model {model_name} was not set to the' \
```
robieta, line 331: Same as above. Probably `RuntimeError`? (That's what we raise for when you try to call an op with incompatible arguments.)

torchbenchmark/__init__.py
```diff
@@ -311,6 +311,42 @@ def check_example() -> None:
...
+                                f' expected device {current_device},' \
+                                f' found device {model_device}.')
+
+        # Check the inputs are assigned to the expected device.
+        if isinstance(inputs, dict):
+            # Huggingface models do not have a Key 0
```
robieta, line 337: I think it's clearer to say "Huggingface models take inputs as kwargs"

robieta, line 337: I think it's clearer to say "Huggingface models take inputs as kwargs"

torchbenchmark/__init__.py
```diff
@@ -311,6 +311,42 @@ def check_example() -> None:
...
+
+        # Check the model tensors are assigned to the expected device.
+        for t in list(model.parameters()):
+            model_device = t.device.type
+            if model_device != current_device:
+                raise Exception(f'Model {model_name} was not set to the' \
+                                f' expected device {current_device},' \
+                                f' found device {model_device}.')
+
+        # Check the inputs are assigned to the expected device.
```
robieta, line 335: We may as well recursively check them all, no? e.g.

```
def check_inputs(inputs):
    if isinstance(inputs, torch.Tensor):
        if inputs.dim() and current_device == "cuda":
            # Zero dim Tensors (Scalars) can be captured by CUDA kernels and need not match device.
            return
            
        inputs_device = inputs.device.type
        if inputs_device != current_device:
            raise RuntimeError(f'Model {model_name} inputs were not set to the' \
                            f' expected device {current_device},' \
                            f' found device {inputs_device}.')

    elif isinstance(inputs, (list, tuple)):
        # Some inputs are nested inside tuples, such as tacotron2
        for i in inputs:
            check_inputs(i)

    elif isinstance(inputs, dict):
        # Huggingface models take inputs as kwargs
        for i in inputs.values():
            check_inputs(i)
```

test.py
```diff
@@ -83,10 +83,21 @@ def eval_fn(self):
             except NotImplementedError:
                 self.skipTest('Method eval is not implemented, skipping...')
 
+    def check_device_fn(self):
```
robieta, line 86: It's not clear to me that this should be a new test, as opposed to being added as a check in `example`.

robieta (changes requested):
I had some nitpicks, but overall looks quite reasonable. And that fact that it's already catching bugs speaks volumes! Thanks!

torchbenchmark/__init__.py
```diff
@@ -311,6 +311,42 @@ def check_example() -> None:
...
+         # Check this BenchmarkModel has a device attribute.
+        current_device = getattr(instance, 'device', None)
+        if current_device is None:
+            raise Exception('Missing device in BenchmarkModel.')
+
+        model, inputs = instance.get_module()
+        model_name = model.name if hasattr(model, 'name') else None
+
+        # Check the model tensors are assigned to the expected device.
+        for t in list(model.parameters()):
```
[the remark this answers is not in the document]
aaronenyeshi: Thanks this works. This originated from debugging, I used list() for some models that had a generator object, list helped to see the values inside. We can safely remove it!

torchbenchmark/__init__.py
```diff
@@ -311,6 +311,42 @@ def check_example() -> None:
...
 
+    @base_task.run_in_worker(scoped=True)
+    @staticmethod
+    def check_device() -> None:
+        instance = globals()["model"]
+
+         # Check this BenchmarkModel has a device attribute.
+        current_device = getattr(instance, 'device', None)
+        if current_device is None:
+            raise Exception('Missing device in BenchmarkModel.')
```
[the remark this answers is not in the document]
aaronenyeshi: Thanks, I will change it to RuntimeError.

torchbenchmark/__init__.py
aaronenyeshi: Thanks, I like this idea! I will change the check to this recursive method.

test.py
aaronenyeshi: I am not sure about this either. I thought that example tests were removed. Running `gpurun python test.py -k check_device` for all 104 check_device tests takes about 580sec (10mins). So would take less overall time if we combine this with `example` (skips many overheads).

robieta: > So it would take less overall time if we combine this with example (skips many overheads).

Yeah, that was my thinking. Where were example tests removed? (Seems like they're still being registered in `test.py`)

aaronenyeshi: It was mentioned in this PR, maybe by accident: https://github.com/pytorch/benchmark/pull/432

robieta: Interesting. CC @xuzhao9 to clarify, but no need to block this PR.

torchbenchmark/__init__.py
```diff
@@ -311,6 +311,52 @@ def check_example() -> None:
...
+
+        # Check the inputs are assigned to the expected device.
+        def check_inputs(inputs):
+            if isinstance(inputs, torch.Tensor):
+                if inputs.dim() and current_device == "cuda":
+                    # Zero dim Tensors (Scalars) can be captured by CUDA
+                    # kernels and need not match device.
+                    return
+
+                inputs_device = inputs[0].device.type
```
robieta, line 343: `inputs.device.type` I think. (The `[0]` is making a new Tensor)

robieta (approved):
One minor nit, but overall LGTM. Thanks!

torchbenchmark/__init__.py
aaronenyeshi: Good catch!

test.py
xuzhao9: > Interesting. CC @xuzhao9 to clarify, but no need to block this PR.

I actually think adding a new test here is clear that we are testing the correct target device of the models, and we may add more tests in the future to guarantee the correctness.

The name of `example` test is a bit vague, plus this is something specific to the need of JIT compiler, not all models. So I suggested we remove the `example` tests in `README.md`, and in the future we may rename it to `jit` test.

robieta: Ah, I see. Thanks!

PR #448, opened 2021-09-10
DCGan Sample As Benchmark

eircfb:
Port dcgan sample to be torchbench benchmark.

Uses data synthesized to profile like celeba image set.

### DIFF
--- a/torchbenchmark/models/dcgan/__init__.py
+++ b/torchbenchmark/models/dcgan/__init__.py
@@ -1,0 +1,334 @@
+# Ported from pytorch example:
+#   https://github.com/pytorch/examples/blob/master/dcgan/main.py
+
+
+from __future__ import print_function
+import argparse
+import os
+import random
+import torch
+import torch.nn as nn
+import torch.nn.parallel
+import torch.backends.cudnn as cudnn
+import torch.optim as optim
+import torch.utils.data
+import torchvision.datasets as dset
+import torchvision.transforms as transforms
+import torchvision.utils as vutils
+import numpy as np
+
+from pathlib import Path
+
+from ...util.model import BenchmarkModel
+from torchbenchmark.tasks import COMPUTER_VISION
+
+class DCGAN:
+ def __init__(self, bench):
+
+  # Set random seed for reproducibility
+  self.manualSeed = 999
+
+  random.seed(self.manualSeed)
+  torch.manual_seed(self.manualSeed)
+
+  # Spatial size of training images. All images will be resized to this
+  #   size using a transformer.
+  self.image_size = 64
+
+  # Number of channels in the training images. For color images this is 3
+  self.nc = 3
+
+  # Size of z latent vector (i.e. size of generator input)
+  self.nz = 100
+
+  # Size of feature maps in generator
+  self.ngf = 64
+
+  # Size of feature maps in discriminator
+  self.ndf = 64
+
+  # Number of training epochs
+  self.num_epochs = 5
+
+  # Learning rate for optimizers
+  self.lr = 0.0002
+
+  # Beta1 hyperparam for Adam optimizers
+  self.beta1 = 0.5
+
+  # Number of GPUs available. Use 0 for CPU mode.
+  self.ngpu = 1
+
+  self.device = bench.device
+
+# custom weights initialization called on netG and netD
+def weights_init(m):
+    classname = m.__class__.__name__
+    if classname.find('Conv') != -1:
+        nn.init.normal_(m.weight.data, 0.0, 0.02)
+    elif classname.find('BatchNorm') != -1:
+        nn.init.normal_(m.weight.data, 1.0, 0.02)
+        nn.init.constant_(m.bias.data, 0)
+
+class Generator(nn.Module):
+    def __init__(self, dcgan):
+        super(Generator, self).__init__()
+        self.ngpu = dcgan.ngpu
+        self.main = nn.Sequential(
+            # input is Z, going into a convolution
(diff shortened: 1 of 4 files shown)


torchbenchmark/models/dcgan/__init__.py
```diff
@@ -0,0 +1,380 @@
...
+        self.jt = None
+        self.jitshape = None
+        self.debug_print = False
+
+    def forward(self, input):
+
+      if self.debug_print:
+        print(input.shape)
+
+      # Not enabled
```
xuzhao9, line 118: May I ask why disabling this code block?

torchbenchmark/models/dcgan/__init__.py
```diff
@@ -0,0 +1,380 @@
...
+
+    def __init__(self, device=None, jit=False):
+        super().__init__()
+
+        self.debug_print = False
+
+        self.jit = jit
+        self.device = device
+        self.root = str(Path(__file__).parent)
+        self.jit = jit
```
xuzhao9, line 200: Why are we assigning `self.jit = jit` twice here? Also, if this model is not supporting JIT, please return `NotImplementedError` for JIT test.

torchbenchmark/models/dcgan/__init__.py
```diff
@@ -0,0 +1,380 @@
...
+  self.manualSeed = 999
+
+  random.seed(self.manualSeed)
+  torch.manual_seed(self.manualSeed)
+
+  # Number of workers for dataloader
+  self.workers = 2
+
+  # Batch size during training
+  self.batch_size = 32 #1024
```
xuzhao9, line 38: Is it better to rename this to `train_batch_size`?

torchbenchmark/models/dcgan/__init__.py
```diff
@@ -0,0 +1,380 @@
...
+        # Establish convention for real and fake labels during training
+        self.real_label = 1.
+        self.fake_label = 0.
+
+        # Setup Adam optimizers for both G and D
+        self.optimizerD = optim.Adam(self.netD.parameters(), lr=lr, betas=(beta1, 0.999))
+        self.optimizerG = optim.Adam(self.netG.parameters(), lr=lr, betas=(beta1, 0.999))
+
+        # Train batch size
+        self.train_b_size = 32
```
xuzhao9, line 259: Why not assigning `self.train_b_size` to `self.dcgan.batch_size`?

torchbenchmark/models/dcgan/__init__.py
```diff
@@ -0,0 +1,380 @@
...
+
+        # Inference Choice:
+        #   just run descriminator
+        #   run generator and descriminator
+
+        # inference would just run descriminator so thats what we'll do too.
+        self.inference_just_descriminator = True
+
+        # eval batch size
+        self.eval_b_size = 256
```
xuzhao9, line 269: Maybe it is better to store `eval_b_size` to `class DCGAN`, and assign `self.eval_b_size = dcgan.eval_b_size` here?

xuzhao9 (changes requested):

torchbenchmark/models/dcgan/__init__.py
eircfb: Jit works now.

torchbenchmark/models/dcgan/__init__.py
eircfb: Its for dealing with dynamic shapes. I've removed it.

xuzhao9:
Could you please rebase this PR on the latest main branch and make another commit?

torchbenchmark/models/dcgan/__init__.py
```diff
@@ -0,0 +1,334 @@
...
+        optimizerG = self.optimizerG
+
+        real_label = self.real_label
+        fake_label = self.fake_label
+
+        # Random values as surrogate for batch of photos
+        benchmark_pic = torch.randn(self.train_b_size, 3, 64, 64, device=self.device)
+
+        # For each epoch
+        for epoch in range(num_epochs):
```
xuzhao9, line 280: Here we are fix `num_epochs` to 5 and setting `num_train_batch` to `niter`. I am thinking whether it is needed to run 5 epochs. Maybe set num_epochs to `niter` and `num_train_batch` to 1, since we probably want to run only 1 batch?

torchbenchmark/models/dcgan/__init__.py
```diff
@@ -0,0 +1,334 @@
...
+        self.jt = None
+        self.jitshape = None
+
+    def forward(self, input):
+      return self.main(input)
+
+class Model(BenchmarkModel):
+    task = COMPUTER_VISION.GENERATION
+
+    def __init__(self, device=None, jit=False):
```
xuzhao9, line 148: In https://github.com/pytorch/benchmark/pull/535/files#diff-d6af0459a37d985953d7040c14f53feb3b9cc9e58b543aa3c2b80256d276c5e0R141 we introduced the `eval_bs` and `train_bs` to support adjusting batch size via `run.py`, can you add parameters to `__init__()` so that we can also test different batches sizes for dcgan through `run.py`?

torchbenchmark/models/dcgan/__init__.py
eircfb: It could be. I only uploaded this so I could profile on my other machine. The dcgan is crossing two models during training and there's significant cpu time between each epoch. I want to be sure to capture the damage between teh epochs but count need to be balanced against runtime.

PR #453, opened 2021-09-10
run.py cudastreams arg for testing utilization

eircfb:
new run.py argument --cudastreams causes benchmark to be run with 1-10 overlapped cuda streams.

Causes current benchmark to be run with same data but overlapped in time, in order to estimate time scaling in presence of overlapped work.

Expectation is that time per stream (reported time / streamcount) should be the same or increase as streamcount rises. If time per stream decreases when streamcount is increased this indicates that gpu has spare capacity with the load and benchmark data size should be investigated.

### DIFF
--- a/run.py
+++ b/run.py
@@ -16,6 +16,40 @@ from torchbenchmark import list_models
 
 import torch
 
+def run_one_step_with_cudastreams(func, streamcount):
+
+    print("Running Utilization Scaling Using Cuda Streams")
+
+    streamlist = []
+    for i in range(1,streamcount + 1,1):
+
+        # create additional streams and prime with load
+        while len(streamlist) < i :
+            s = torch.cuda.Stream()
+            streamlist.append(s)
+
+        for s in streamlist:
+            with torch.cuda.stream(s):
+                func()
+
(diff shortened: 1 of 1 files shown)


run.py
```diff
@@ -81,5 +120,7 @@ def profile_one_step(func, nwarmup=3):
 
     if args.profile:
         profile_one_step(test)
+    elif args.cudastreams:
+        run_one_step_with_cudastreams(test, 10)
```
xuzhao9, line 124: May I ask why the number of streams is set to 10? Is it better to specify the number of streams through the argument `--cudastreams`?

xuzhao9 (approved):
Looks good. Is the cuda stream test supported by all CUDA models? If not, do we need to blacklist/whitelist?

run.py
eircfb: > May I ask why the number of streams is set to 10? Is it better to specify the number of streams through the argument `--cudastreams`?

I thought an arg would make it more complicated and I don't think there's an actual benefit. If you can't learn from streams <= 10 then you're barking up the wrong tree. The few times I ran it with high stream count I wanted an iteration step, etc. And still it didn't show anything.

PR #454, opened 2021-09-10
Prevent eval of torchvision benchmarks in unsqueeze to dim 0

aaronenyeshi:
This line forced resnet50 benchmark to use batch size of 1.

This is regarding comment on PR https://github.com/pytorch/benchmark/pull/435#issuecomment-917194935. I am wondering if this should be removed from the generator as well as all torchvision benchmarks.


xuzhao9 (approved):
Thanks!

PR #455, opened 2021-09-10
Cleanup the garbage files

xuzhao9:
This is to reduce the total size of the repository, as we are trying to import this project to FB internal repo.


bitfort (approved):
Great!

PR #457, opened 2021-09-14
Remove the legacy directory that is not used anymore.

xuzhao9:
Further cleanup the unused files and reduce the repository size.


erichan1 (approved):
removal makes sense, lgtm!

PR #458, opened 2021-09-15
Remove the version.py file by force instead of generating it.

xuzhao9:
This is to fix the workflow failure in https://github.com/pytorch/pytorch/pull/64863/checks?check_run_id=3614506377


alanwaketan (approved):
LGTM. Thanks for fixing this so quickly.

PR #460, opened 2021-09-15
Remove the version.py file by force instead of generating it.

xuzhao9:
Since PyTorch doesn't update the CUDA version in `version.py` when running `generate_pytorch_version.py` anymore, we remove the `version.py` file and let the compilation script regenerate it.


alanwaketan (approved):
LGTM. Thanks for fixing the issue.

PR #461, opened 2021-09-16
Fix the bug in the V1 score calculation algorithm

xuzhao9:
In V1 score, we are using the weighted geometric average of the latency speedups across all tests comparing with the reference execution. When the execution runs faster, we should get a higher perf score.
The `norm` variable is the test latency, and `ref_norm` stores the test latency of reference execution.


bitfort (approved):
Great!

PR #462, opened 2021-09-16
Remove unnecessary blob files from the repository.

xuzhao9:
Binary files that are not needed for benchmarking should be removed.


erichan1 (approved):
lgtm! good that we're cleaning up TB.

PR #463, opened 2021-09-17
Remove the datasets input data and manage them in a separate branch.

xuzhao9:
To shrink the size of data being synced with FB Internal, we are removing the input data from the main branch and placing them in the [model-data](https://github.com/pytorch/benchmark/tree/model-data) branch.


erichan1 (approved):
lgtm! let's get oss shipped to internal soon.

PR #465, opened 2021-09-20
Maml update: synthesize data and parameterize

eircfb:
Generate model values instead of loading from 10mb file. Also model sizes are now partially parameterized.

This is a nop change wrt performance but simplifies future model investigation.

We can remove batch.pt from repo once we're happy with with model performance.


xuzhao9 (approved):
LGTM!

PR #466, opened 2021-09-21
Enable lfs to manage the blob files in the repository

xuzhao9:
Manage the blob files using Git LFS instead of plain git. This is helpful to reduce the repository size in FB Internal.


erichan1 (approved):
lgtm!

PR #467, opened 2021-09-22
Remove the files with case-sensitive names

xuzhao9:
This is because FB Internal doesn't support case-sensitive filenames. We need to remove the files with case-sensitive names under the same directory.

This PR also adds all files larger than 1M to git lfs.


erichan1 (approved):
lgtm!

PR #468, opened 2021-09-23
Remove fastNLP docs and tutorials to avoid non-ascii characters.

xuzhao9:
This PR is another preparation for FB Internal import.


erichan1 (approved):
lgtm!

PR #470, opened 2021-09-24
Revert "fbshipit-source-id: ef79c12fbdacd99d76143dc29bbfb7f51e226968"

erichan1:
This reverts commit 7ef95f4b0b144a568c83e57944475e7e019d16ab.

Restores all files that were unnecessarily deleted to sync internal and external repos.


xuzhao9 (approved):
LGTM. Thanks for fixing this!

xuzhao9:
Actually, we need to manually move the files back because this PR is only moving the stub files back, not the actual blob data.

PR #473, opened 2021-09-27
Modify run.py to use torch.profiler

aaronenyeshi:
Deprecate the use of torch.autograd.profiler, and use
torch.profiler. Clean up a few linting errors. Print
Tensorboard traces to a default folder, and add an
option for the user to configure the trace folder.

### DIFF
--- a/run.py
+++ b/run.py
@@ -82,7 +82,7 @@ def run_one_step(func):
 
 def profile_one_step(func, nwarmup=3):
     for i in range(nwarmup):
-            func()
+        func()
 
     activity_groups = []
     if args.device == "cuda":
@@ -93,7 +93,7 @@ def profile_one_step(func, nwarmup=3):
         schedule=profiler.schedule(wait=0, warmup=0, active=1),
         activities=activity_groups,
         record_shapes=True,
-        on_trace_ready=profiler.tensorboard_trace_handler('./logs')
+        on_trace_ready=profiler.tensorboard_trace_handler(args.profile_folder)
     ) as prof:
         func()
 
--- a/run.py
+++ b/run.py
@@ -10,18 +10,19 @@ Wall time provided for sanity but is not a sane benchmark measurement.
 """
 import argparse
 import time
-import torch.autograd.profiler as profiler
+import torch.profiler as profiler
 
 from torchbenchmark import list_models
 
 import torch
+
 
 def run_one_step_with_cudastreams(func, streamcount):
 
     print("Running Utilization Scaling Using Cuda Streams")
 
     streamlist = []
-    for i in range(1,streamcount + 1,1):
+    for i in range(1, streamcount + 1, 1):
 
         # create additional streams and prime with load
         while len(streamlist) < i :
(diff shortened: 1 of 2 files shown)


aaronenyeshi:
The CI seems to be failing on an un-related issue. Any idea how to get past this error?
```
+ git lfs fetch
fetch: Fetching reference refs/heads/aaron-cleanup-run
[e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855] Object does not exist on the server: [404] Object does not exist on the server
error: failed to fetch some objects from 'https://github.com/pytorch/benchmark.git/info/lfs'

Exited with code exit status 2
CircleCI received exit code 2
```

run.py
```diff
@@ -50,56 +51,68 @@ def run_one_step_with_cudastreams(func, streamcount):
...
+    activity_groups = []
+    if args.device == "cuda":
+        activity_groups.append(profiler.ProfilerActivity.CUDA)
+    activity_groups.append(profiler.ProfilerActivity.CPU)
+
+    with profiler.profile(
+        schedule=profiler.schedule(wait=0, warmup=0, active=1),
+        activities=activity_groups,
+        record_shapes=True,
+        on_trace_ready=profiler.tensorboard_trace_handler('./logs')
```
wconstab, line 96: shouldn't you take args.profile-folder here instead of './logs'?

wconstab (approved):
lgtm except the one issue I pointed out, stamping now to unblock but please fix that

run.py
aaronenyeshi: Good catch, thanks that was remains from my testing. fixing it!

aaronenyeshi:
> lgtm except the one issue I pointed out, stamping now to unblock but please fix that

Thanks, fixed the issue. Looks like this needs to be imported internally for merge. will do that once the import option comes back up.

aaronenyeshi:
Rebased due to CI errors yesterday. Also moved the warmup loop inside of prof block, as intended by profiler APIs.

run.py
```diff
@@ -50,56 +51,67 @@ def run_one_step_with_cudastreams(func, streamcount):
...
     parser = argparse.ArgumentParser(__doc__)
     parser.add_argument("model", help="Full or partial name of a model to run.  If partial, picks the first match.")
-    parser.add_argument("-d", "--device", choices=["cpu",  "cuda"], default="cpu", help="Which device to use.")
-    parser.add_argument("-m", "--mode", choices=["eager",  "jit"], default="eager", help="Which mode to run.")
-    parser.add_argument("-t", "--test", choices=["eval",  "train"], default="eval", help="Which test to run.")
+    parser.add_argument("-d", "--device", choices=["cpu", "cuda"], default="cpu", help="Which device to use.")
+    parser.add_argument("-m", "--mode", choices=["eager", "jit"], default="eager", help="Which mode to run.")
+    parser.add_argument("-t", "--test", choices=["eval", "train"], default="eval", help="Which test to run.")
     parser.add_argument("--profile", action="store_true", help="Run the profiler around the function")
+    parser.add_argument("--profile-folder", default="./logs", help="Save profiling model traces to this directory.")
```
xuzhao9, line 110: Please add "logs/" to `.gitignore` to prevent future possible commits of the directory.

run.py
```diff
@@ -50,56 +51,67 @@ def run_one_step_with_cudastreams(func, streamcount):
...
+
+    with profiler.profile(
+        schedule=profiler.schedule(wait=0, warmup=nwarmup, active=1),
+        activities=activity_groups,
+        record_shapes=True,
+        on_trace_ready=profiler.tensorboard_trace_handler(args.profile_folder)
+    ) as prof:
+        for _i in range(nwarmup + 1):
+            func()
+            prof.step()
```
xuzhao9, line 97: Just curious that you are now running `prof.step()` with warmup loops, is there a reason for that?

xuzhao9 (approved):
LGTM!

run.py
aaronenyeshi: Will do!

run.py
aaronenyeshi: I think both ways work the same, I need to look into if profiler does anything special for warmup steps. I think this is the expected way to use `torch.profiler` from the docs. This change reflects the docs better, but is not necessary. Do you have a preference? I don't mind changing it back.

xuzhao9: Either way looks okay to me too :)

aaronenyeshi:
@erichan1 and @xuzhao9 , this PR is merged internally, but not reflected on the `main` branch in Github. Any idea what went wrong? (In kineto, the main branch is updated with the commit before closing the PR).

PR #479, opened 2021-09-29
Fix get_module() API for 9 benchmarks

jansel:
I use the `get_module()` API for output correctness checking and other types of transformations (like FX).  These 10 benchmarks did not implement it according to spec.


wconstab (approved):

aaronenyeshi (approved):
LGTM!

xuzhao9:
Sorry can I ask one thing about `get_module()` API? Does it require `example_input` to be the data used in train? Or it doesn't matter whether the input is from train or eval?  @jansel

jansel:
@xuzhao9

AFAIK get_module() it is eval only.  It doesn't include the loss function or optimizer, so you can't really use it for training in a realistic way.

My main motivation for using it (over the train/eval APIs) is that it lets me check the correctness of the outputs and try transformations other than jit=True/jit=False.

xuzhao9:
@jansel I am asking this because some models (such as NLP) use different inputs in train and eval mode (e.g., train mode data include targets). Good to know `get_modules()` uses eval mode and data!

PR #486, opened 2021-10-01
Add util to collect profiler traces.

robieta:
While this is a large PR according to SLOC, most of the changes are mechanical. In order to collect traces, we need to do a bit of standardization:

1) Standardize to `niter` across all models. (Currently some use `niterations`)
2) Make `train` and `eval` take a `step_fn` argument so we can know where to draw appropriate boundaries. (e.g. for warmup.) Note that the model is free to define a "step" in whatever way makes sense for the context; we just care that the function is called. That said, most follow the pattern of `forward, backward, optimizer, step`.
3) *Optionally* models can annotate the forward, backward, and optimizer parts of the code. Unusual models like RL or GANs don't, but for most models this is straightforward and leads to a nicely annotated profile.

Interestingly, this project flushed out a bunch of bugs just by either running `niter` > 1 or inspection of the code during standardization.
* In DLRM it wasn't possible to run with `niter` > 1, because the forward pass was outside of the loop.
* We weren't zeroing grads for any of the huggingface models.
* `train` and `eval` are identical for `maml`, so we were just duplicating work.
* Some of the torchvision models had drifted from the gen'd version due to a merge conflict resolution 4 months ago.

Aside from that, the two main blocks of new code are the `collect_diagnostics` / `_collect_diagnostics` methods in `torchbenchmark/__init__.py` and the new top level `diagnostics.py` The high level flow is to do some runs to characterize the behavior, and then take a lightweight profile and a detailed profile. The former is for low distortion, while the latter makes it easy to follow up. (e.g. check the batch size.)

One of the interesting side effects of comparing normal step time to profiled step time is you can see which models are dispatch bound, or very close to dispatch bound. (All of these statistics are recorded in `traces/summary.txt`.) We should audit all of the traces as a team to sanity check which ones are providing useful signal and which ones need to be tweaked.


torchbenchmark/util/model.py
```diff
@@ -48,6 +50,18 @@ def _set_mode(self, train):
         (model, _) = self.get_module()
         model.train(train)
 
+    @staticmethod
```
gdankel, line 53: I've been considering whether we should provide a standard set of labels such as these that models can use and we can pick assume to have a certain meaning in the profiler / report - what do you think about such an idea?

torchbenchmark/__init__.py
```diff
@@ -288,6 +292,80 @@ def train(self) -> None:
...
+                ),
+                on_trace_ready=on_trace_ready,
+                record_shapes=detailed,
+                profile_memory=detailed,
+                with_stack=detailed,
+                with_flops=detailed,
+            ) as prof:
+                times = []
+                def step_fn() -> None:
+                    maybe_sync()
```
gdankel, line 355: What does this do exactly? When do we want to sync during profiling?

torchbenchmark/__init__.py
```diff
@@ -288,6 +292,80 @@ def train(self) -> None:
...
+        baseline_times = []
+        for _ in range(6):
+            t = [time.time()]
+            model_fn(step_fn=lambda: t.append(time.time()))
+            maybe_sync()
+            baseline_times.append(t + [time.time()])
+        count = len(baseline_times[0]) - 2
+        assert count > 0 and all(len(t) - 2 == count for t in baseline_times)
+
+        schedule = (max(count, 10), 5, 5)
```
gdankel, line 329: Add a comment explaining these values? Why warmup of 5?

gdankel (commented):
This is really awesome!!

xuzhao9 (commented):
The improvement looks great! Still looking into the details...
Here are my thoughts so far:
- We are now adding `step_fn` argument to the `train()` and `eval()` interface, and requires model contributors to insert `annotate_forward()`, `annotate_backward()`, `annotate_optimizer()` into the function. This will increase the work of model contributors and people may forget or insert them at incorrect locations. Can we use some unit tests to prevent those?
- After adding the annotation functions, does that mean we will always run the benchmark with profiler enabled? Can we add an option to make those annotations as no-ops and don't run with profilers?

PR #493, opened 2021-10-06
Fix run.py --profile options to support choosing device activity

aaronenyeshi:
When combining CPU and GPU trace activities into a profile
trace, the output trace hides all the GPU gaps inside of the
CPU profiling overhead. We need to collect GPU trace only when
running in -d cuda mode.

### DIFF
--- a/run.py
+++ b/run.py
@@ -82,14 +82,23 @@ def run_one_step(func):
 
 def profile_one_step(func, nwarmup=3):
     activity_groups = []
-    if args.device == "cuda":
+    if ((not args.profile_devices and args.device == 'cuda') or
+            (args.profile_devices and 'cuda' in args.profile_devices)):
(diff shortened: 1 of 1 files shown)


aaronenyeshi:
I've updated the patch to allow the user to toggle which activity_groups they want to trace using `--profile-devices`. Description to commit updated:
```
    When combining CPU and GPU trace activities into a profile
    trace, the output trace may hide some the GPU gaps inside of the
    CPU profiling overhead. Instead, we collect CUDA only trace when
    running in -d cuda. However, we provide an option --profile-devices
    which allows the user to choose which activity groups they want to
    collect such as cpu,cuda. Also, add --profile-detailed which
    collects record_shapes, profile_memory, with_stack, and with_flops.
```

PR #501, opened 2021-10-15
Enable writing stats to metrics.pytorch.org

driazati:
This copies over the pytorch/pytorch utils to write stats to metrics.pytorch.org where they can be graphed / displayed publicly


robieta:
> I noticed in the stats it reports individual data points along with a roll up (mean/median/etc). Is it worth reporting both here or should we just use the aggregated stats?

Those are included so we can check for noise. (Better to keep it than to want it later.) I figured it's not that much data, so wouldn't hurt. Is that correct or does that store too much?

PR #542, opened 2021-11-03
Fix the import issue of speech_transformer.

xuzhao9:
When adding the speech_transformer model, we modified the syspath import path. 
This is actually undesired because it will also affect the import path of other models. Although in CI environment we are running models in separate processes, this could still cause problem when locally running with `run.py`.

Therefore, here we are moving the model to a subdirectory `speech_transformer`, and change all the import styles to relative imports.


aaronenyeshi (approved):
LGTM!

PR #547, opened 2021-11-08
Adding maskrcnn model from torchvision

xuzhao9:
Add maskrcnn model from torchvision.
This is the first step towards adding dynamic shapes support for maskrcnn.


torchbenchmark/models/maskrcnn/__init__.py
```diff
@@ -0,0 +1,102 @@
...
+                self.optimizer.zero_grad()
+                self.scheduler.step()
+
+    def eval(self, niter=1):
+        if self.jit:
+            return NotImplementedError("JIT is not supported by this model")
+        self.model.eval()
+        for iter in range(niter):
+            for image in self.infer_example_inputs:
+                self.model(image)
```
xuzhao9, line 99: This line returns the following error:
```
    "images is expected to be a list of 3d tensors " "of shape [C, H, W], got {}".format(image.shape)
ValueError: images is expected to be a list of 3d tensors of shape [C, H, W], got torch.Size([400, 300])
```

xuzhao9: Resolved.

xuzhao9:
The eval GPU utilization looks horrible:

image

Looking into it.

xuzhao9:
Changed to using COCO 2017 data, the utilization for evaluation still looks bad:

image

xuzhao9:
After adding data prefetch, the gap in the front is gone but the gap at the tail still exists (for inference):

image

torchbenchmark/models/maskrcnn/__init__.py
```diff
@@ -0,0 +1,109 @@
...
+        for _, (images, targets) in zip(range(niter), self.train_data_loader):
+            # images = list(image.to(self.device) for image in images)
+            # targets = [{k: v.to(self.device) for k, v in t.items()} for t in targets]
+            loss_dict = self.model(images, targets)
+            losses = sum(loss for loss in loss_dict.values())
+            self.optimizer.zero_grad()
+            losses.backward()
+            self.optimizer.step()
+
+    def eval(self, niter=1):
```
Krovatkin, line 97: @xuzhao9 I'm a bit concerned we seem to be getting different batches out of `eval_data_loader` every time we run `eval`? Since maskrcnn seems to be data-dependent and bath sizes are so small, we might get running times for eval that vary greatly from one run to another (i.e. noise).
Is this how we usually use data loaders?

torchbenchmark/models/maskrcnn/__init__.py
```diff
@@ -0,0 +1,109 @@
...
+    def __init__(self, device=None, jit=False, train_bs=4, eval_bs=4):
+        self.device = device
+        self.jit = jit
+        self.model = torchvision.models.detection.maskrcnn_resnet50_fpn(pretrained=True).to(self.device)
+
+        # setup optimizer
+        lr = 0.02
+        momentum = 0.9
+        weight_decay = 1e-4
+        params = [p for p in self.model.parameters() if p.requires_grad]
```
Krovatkin, line 58: out of curiosity, is this some kind of an optimization? wouldn't autograd ignore parameters don't require gradients?

Krovatkin (commented):
a few qqs to understand this PR a bit better.

torchbenchmark/models/maskrcnn/__init__.py
xuzhao9: @Krovatkin I borrowed these default values from torchvision references: https://github.com/pytorch/vision/blob/main/references/detection/train.py#L77. These parameters seems to be default to torchvision maskrcnn training.

torchbenchmark/models/maskrcnn/__init__.py
xuzhao9: @Krovatkin Could you please explain why we are getting different batches out of `eval_data_loader` every time? I am thinking we are always getting the first batch because `self.eval_data_loader` is actually a python list that always start from index `0`. Even if it is a data loader object, I think in a for loop it also always start from the first batch?

torchbenchmark/models/maskrcnn/__init__.py
fmassa: I think we don't need this anymore, including in torchvision.

fmassa:
Hi @xuzhao9 

I believe the slowdown during inference of mask r-cnn is due [to this function](https://github.com/pytorch/vision/blob/a2f65656b2b9dd7cc4363c8ad4afcbf89a25952b/torchvision/models/detection/roi_heads.py#L472-L487)

Indeed, we perform many small operations in a for loop, which is not ideal for maximizing GPU utilization.
In particular, I believe [this line](https://github.com/pytorch/vision/blob/a2f65656b2b9dd7cc4363c8ad4afcbf89a25952b/torchvision/models/detection/roi_heads.py#L482) is the main culprit.

I think we can further optimize this function at the Python-level by leveraging a single call to `grid_sample`, but it would require rewriting the function. I believe detectron2 contains a more optimized version of this function in [here](https://github.com/facebookresearch/detectron2/blob/aa8ea943411a2e5cd616e4d517215779a2ea2dad/detectron2/layers/mask_ops.py#L74)

Also, ccing @datumbox  as this could be a good task to optimize the torchvision function

eircfb:
> After adding data prefetch, the gap in the front is gone but the gap at the end still exists for eval:
> 
> image
> 
> At the same time, train profiling looks terrific:
> 
> image

Question for model authors:

This still doesnt' look great. Is this how the model is used in nature?

What can be done here to improve utilization? Is this from a lack of prefetch from the second model? Its no good to have that idle section at the end.

eircfb:
> Hi @xuzhao9
> 
> I believe the slowdown during inference of mask r-cnn is due [to this function](https://github.com/pytorch/vision/blob/a2f65656b2b9dd7cc4363c8ad4afcbf89a25952b/torchvision/models/detection/roi_heads.py#L472-L487)
> 
> Indeed, we perform many small operations in a for loop, which is not ideal for maximizing GPU utilization. In particular, I believe [this line](https://github.com/pytorch/vision/blob/a2f65656b2b9dd7cc4363c8ad4afcbf89a25952b/torchvision/models/detection/roi_heads.py#L482) is the main culprit.
> 
> I think we can further optimize this function at the Python-level by leveraging a single call to `grid_sample`, but it would require rewriting the function. I believe detectron2 contains a more optimized version of this function in [here](https://github.com/facebookresearch/detectron2/blob/aa8ea943411a2e5cd616e4d517215779a2ea2dad/detectron2/layers/mask_ops.py#L74)
> 
> Also, ccing @datumbox as this could be a good task to optimize the torchvision function

Right, sending off all these tiny shaders against independent data. Thats a perf error, causes dispatch to dominate. Can maybe be patched up by using cuda graphs, but really not good.

Is this really how the model runs in nature?

fmassa:
@eircfb I believe I've answered your questions in my answer just before, let me know if you have further questions. It's the post-processing steps that are the culprit in here, and they can definitely be optimized (without having to resort to custom CUDA kernels)

This is normally how we use the model, yes.

eircfb:
> @eircfb I believe I've answered your questions in my answer just before, let me know if you have further questions. It's the post-processing steps that are the culprit in here, and they can definitely be optimized (without having to resort to custom CUDA kernels)
> 
> This is normally how we use the model, yes.

Right, we posted at same time. Isn't that loop of tiny dispatches just for the first section? What causes the low utilization towards the end of the trace?

fmassa:
@eircfb I believe the section I pointed out should correspond to the last part of the trace (although to be 100% sure I would need to check the full trace with further info, @xuzhao9 can you send it to me?)

For the first section with slowdown, I would need to check more carefully but my first guess would be that it comes from this section https://github.com/pytorch/vision/blob/0fa747eea4569f349fb0dacab44604b7b11c1a74/torchvision/models/detection/rpn.py#L253-L274

Maybe the slowdown could be attributed to [this function being called](https://github.com/pytorch/vision/blob/0fa747eea4569f349fb0dacab44604b7b11c1a74/torchvision/ops/boxes.py#L71) instead of [this one](https://github.com/pytorch/vision/blob/0fa747eea4569f349fb0dacab44604b7b11c1a74/torchvision/ops/boxes.py#L73), but again, I would need to see the full trace to be sure

xuzhao9:
> @eircfb I believe the section I pointed out should correspond to the last part of the trace (although to be 100% sure I would need to check the full trace with further info, @xuzhao9 can you send it to me?)
> 
> For the first section with slowdown, I would need to check more carefully but my first guess would be that it comes from this section https://github.com/pytorch/vision/blob/0fa747eea4569f349fb0dacab44604b7b11c1a74/torchvision/models/detection/rpn.py#L253-L274
> 
> Maybe the slowdown could be attributed to [this function being called](https://github.com/pytorch/vision/blob/0fa747eea4569f349fb0dacab44604b7b11c1a74/torchvision/ops/boxes.py#L71) instead of [this one](https://github.com/pytorch/vision/blob/0fa747eea4569f349fb0dacab44604b7b11c1a74/torchvision/ops/boxes.py#L73), but again, I would need to see the full trace to be sure

Here is the full trace of eval (bs==4, with prefetch): Eval trace [pytorch-benchmark-dev-0_1709.1636590532626.pt.trace.json.zip](https://github.com/pytorch/benchmark/files/7530875/pytorch-benchmark-dev-0_1709.1636590532626.pt.trace.json.zip)

torchbenchmark/models/maskrcnn/__init__.py
xuzhao9: Test failed when removing these parameters(https://app.circleci.com/pipelines/github/pytorch/benchmark/2710/workflows/26c92ded-69b3-4cdc-8c35-3cc4ab5f8e13/jobs/2793), so they are still needed.

xuzhao9:
> @xuzhao9 after looking at the full trace, my comments from before seems to be accurate.
> 
> The first slow-down is [due to this part](https://github.com/pytorch/vision/blob/0fa747eea4569f349fb0dacab44604b7b11c1a74/torchvision/models/detection/rpn.py#L253-L273) of the implementation. I have a gut-feeling that this might be due to bad hyperparameters in [here](https://github.com/pytorch/vision/blob/0fa747eea4569f349fb0dacab44604b7b11c1a74/torchvision/ops/boxes.py#L70), so if you could try always dispatching to [this codepath](https://github.com/pytorch/vision/blob/0fa747eea4569f349fb0dacab44604b7b11c1a74/torchvision/ops/boxes.py#L73) it would give hints if we can easily fix this.
> 
> For the second (and biggest slowdown), the problem is what I mentioned before on the mask pasting in images, and would require re-writing the function in a better way

Thank you @fmassa! I think your explanations make sense (though I am not an expert). I think the code is in decent shape now and we are confident it represents common use cases. We can use it as a baseline to help further improve the model quality.

Please help review and stamp it so that we can proceed on developing dynamic shapes.

Thanks!

torchbenchmark/models/maskrcnn/__init__.py
```diff
@@ -0,0 +1,112 @@
...
+    def get_module(self):
+        self.model.eval()
+        for (example_inputs, _example_targets) in self.eval_data_loader:
+            return self.model, (example_inputs, )
+
+    def train(self, niter=1):
+        if self.jit:
+            return NotImplementedError("JIT is not supported by this model")
+        if not self.device == "cuda":
+            return NotImplementedError("CPU is not supported by this model")
```
fmassa, line 89: Note that both torchscript and CPU are supported.
You can do
```python
torch.jit.script(model)
```
and it should work just fine. Same for CPU

torchbenchmark/models/maskrcnn/transforms.py
```diff
@@ -0,0 +1,289 @@
...
+
+            is_pil = F._is_pil_image(image)
+            if is_pil:
+                image = F.pil_to_tensor(image)
+                image = F.convert_image_dtype(image)
+            image = image[..., permutation, :, :]
+            if is_pil:
+                image = F.to_pil_image(image)
+
+        return image, target
```
fmassa, line 289: All those data augmentation transforms are only used in the data loader (are outside of the model) and are not used in the profiling example, so could probably be removed

torchbenchmark/models/maskrcnn/transforms.py
```diff
@@ -0,0 +1,289 @@
...
+
+def _flip_coco_person_keypoints(kps, width):
+    flip_inds = [0, 2, 1, 4, 3, 6, 5, 8, 7, 10, 9, 12, 11, 14, 13, 16, 15]
+    flipped_data = kps[:, flip_inds]
+    flipped_data[..., 0] = width - flipped_data[..., 0]
+    # Maintain COCO convention that if visibility == 0, then x, y = 0
+    inds = flipped_data[..., 2] == 0
+    flipped_data[inds] = 0
+    return flipped_data
+
```
fmassa, line 24: Same for this function, it's not currently used so could be removed if wanted.

torchbenchmark/models/maskrcnn/__init__.py
```diff
@@ -0,0 +1,112 @@
...
+        lr = 0.02
+        momentum = 0.9
+        weight_decay = 1e-4
+        params = [p for p in self.model.parameters() if p.requires_grad]
+        self.optimizer = torch.optim.SGD(params, lr=lr, momentum=momentum, weight_decay=weight_decay)
+
+        self.train_bs = train_bs
+        self.eval_bs = eval_bs
+        t = [ConvertCocoPolysToMask()]
+        transforms = Compose(t)
```
fmassa, line 66: Given that we only apply the `ConvertCocoPolysToMask()` transform in here, you could probably simplify this as follows:
```suggestion
        transforms = ConvertCocoPolysToMask()
```
and remove the whole `transforms.py` file

fmassa (approved):
Thanks!

I've made a few comments that could simplify the implementation (by removing unused functions), and also to mention that both torchscript as well as CPU are supported.

Otherwise the rest LGTM so I'm approving this as it can be merged as is

torchbenchmark/models/vision_maskrcnn/__init__.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+    def get_module(self):
+        self.model.eval()
+        for (example_inputs, _example_targets) in self.eval_data_loader:
+            return self.model, (example_inputs, )
+
+    def train(self, niter=1):
+        if not self.device == "cuda":
+            return NotImplementedError("CPU is not supported by this model")
```
fmassa, line 90: For my understanding, did you face any errors when running the code on the CPU?

xuzhao9: Sorry I haven't tested CPU. But the JIT run seems to be failed: https://app.circleci.com/pipelines/github/pytorch/benchmark/2729/workflows/474cf2b6-f2d5-4a13-b08d-7f634886d475/jobs/2812.

fmassa: Note that during scripting, the model always returns `(Losses, Detections)` (as pointed out in the error message).

So you would need to change the code in the training to always return `losses, detection = model(inputs)`, and go from there.

jdsgomes:
> I've made a few comments that could simplify the implementation (by removing unused functions), and also to mention that both torchscript as well as CPU are supported.

I re-ran the profiling with the code changes from [this PR](https://github.com/pytorch/vision/pull/4990/files) and it indeed it improves the overall figure (visible in the middle part where we are running nms). 

After changes 
Screenshot 2021-12-10 at 22 35 12

Before changes
Screenshot 2021-12-10 at 22 35 29

PR #558, opened 2021-11-11
Temporarily remove the attention model and fix pytorch_struct model.

xuzhao9:
torchtext removes the legacy dataset utilities(https://github.com/pytorch/text/pull/1437), therefore we need to migrate to the new dataset API or keep the old API but copy the related code here. This PR we still use the old API because it seems non-trivial to migrate to the new API.

I will re-add the attention model in a follow-up PR (and do the quality analysis there).

Also, pytorch_struct model runs an [unsupervised learning task](https://github.com/harvardnlp/pytorch-struct/blob/master/notebooks/Unsupervised_CFG.ipynb), therefore, it does not support eval test.

# batch size analysis

<google-sheets-html-origin>

Batch Size | GPU Time | CPU Dispatch Time | Walltime | GPU Delta
-- | -- | -- | -- | --
16 | 53.06 | 52.969 | 53.072 | -
32 | 85.123 | 75.824 | 85.126 | 0.6042781757
64 | 157.218 | 121.155 | 157.23 | 0.8469508828
128 | 315.678 | 242.213 | 315.681 | 1.007899859
256 | 568.102 | 428.017 | 568.098 | 0.7996249343

# Non-idleness analysis (train, bs=128)
image

GPU is mostly idle when bs=32, so I am testing with bs=128 instead.
Data is already prefetched to the device.


xuzhao9:
Disable JIT because of the following exception:

```
torch.jit.frontend.UnsupportedNodeError: function definitions aren't supported:
  File "/home/circleci/project/torchbenchmark/models/pytorch_struct/networks/NeuralCFG.py", line 46
        T, NT = self.T, self.NT
    
        def terms(words):
        ~~~ <--- HERE
            b, n = input.shape[:2]
            term_prob = (
```

eircfb (approved):
Got to make it work.

pytorch_struct utilization isn't great, all those tiny dispatches - might be a good place to try cudagraphs.

aaronenyeshi (approved):
Thanks for fixing this

PR #560, opened 2021-11-12
Correct LearningToPaint to match the wild

aaronenyeshi:
Previously this model was only running the Stroked-Based
Renderer model, which is a simple nn with linear and conv
layers. But the real LearningToPaint model consists of
WGAN, discriminator and renderer. Adding the correct model
implementation.

### DIFF
--- a/torchbenchmark/models/LearningToPaint/__init__.py
+++ b/torchbenchmark/models/LearningToPaint/__init__.py
@@ -1,14 +1,12 @@
 import cv2
 import torch
+import random
 import numpy as np
-import sys
+from .baseline.DRL.evaluator import Evaluator
+from .baseline.utils.util import *
+from .baseline.DRL.ddpg import DDPG
+from .baseline.DRL.multi import fastenv
 
-import torch.nn as nn
-import torch.nn.functional as F
-import torch.optim as optim
-from .baseline.utils.tensorboard import TensorBoard
-from .baseline.Renderer.model import FCN
-from .baseline.Renderer.stroke_gen import *
 from ...util.model import BenchmarkModel
 from torchbenchmark.tasks import REINFORCEMENT_LEARNING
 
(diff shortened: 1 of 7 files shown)


aaronenyeshi:
Here are the collected traces for updated PR:
Train:
image

Eval:
image

xuzhao9:
Thanks for improving the LearningToPaint model quality. One question I have is do we observe similar GPU utilization pattern in the original LearningToPaint paper code demo? It is the ground truth and we would like to confirm that our reproduction matches the ground truth.

torchbenchmark/models/LearningToPaint/__init__.py
```diff
@@ -1,99 +1,99 @@
...
 from torchbenchmark.tasks import REINFORCEMENT_LEARNING
 
 from argparse import Namespace
 
 torch.manual_seed(1337)
 np.random.seed(1337)
-torch.backends.cudnn.deterministic = True
-torch.backends.cudnn.benchmark = False
+torch.backends.cudnn.deterministic = False
+torch.backends.cudnn.benchmark = True
```
xuzhao9, line 18: Maybe we should set this config globally instead of setting it for individual models repeatedly. But we can do it in a separate PR.

torchbenchmark/models/LearningToPaint/__init__.py
```diff
@@ -1,99 +1,99 @@
...
 
     def get_module(self):
-        return self.module,[self.example_inputs]
-
-    # LearningToPaint has another model
-    # instance for inference that has
-    # already been optimized for inference
-    def set_eval(self):
-        pass
+        raise NotImplementedError()
```
xuzhao9, line 54: This API has some users (e.g., https://github.com/pytorch/benchmark/pull/479), can you implement this (e.g. using randomly generate inputs to the actor model)?

aaronenyeshi:
> Thanks for improving the LearningToPaint model quality. One question I have is do we observe similar GPU utilization pattern in the original LearningToPaint paper code demo? It is the ground truth and we would like to confirm that our reproduction matches the ground truth to help confirm that the bad utilization is a problem of user code, not our reproduction.

Here is a 1 second snippet of the eval trace in the original LearningToPaint code:
image

torchbenchmark/models/LearningToPaint/__init__.py
aaronenyeshi: Sounds good to me!

torchbenchmark/models/LearningToPaint/__init__.py
aaronenyeshi: Sure, I can take a further look into this.

aaronenyeshi: done!

aaronenyeshi:
This implementation uses the parameters from LearningToPaint Source code and research paper.

xuzhao9 (approved):
LGTM. Thanks for improving this model!

PR #561, opened 2021-11-12
Set torchtext version to 0.9 for models that use torchtext.legacy

abhinavarora:
As we are deprecating `torchtext.legacy`, we would like to freeze the version of torchtext to 0.9 for models that will use `torchtext.legacy` package.

### DIFF
--- a/torchbenchmark/models/attention_is_all_you_need_pytorch/requirements.txt
+++ b/torchbenchmark/models/attention_is_all_you_need_pytorch/requirements.txt
@@ -2,3 +2,4 @@ dill
 tqdm
 numpy
 spacy==2.3.5
+torchtext==0.9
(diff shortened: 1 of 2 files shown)


xuzhao9:
Unfortunately, fixing torchtext version is undesired because we need to support bisection, which requires re-compiling pytorch/torchvision/torchtext on the recent commits. Instead, I submitted a PR that copies all the legacy code to the models that need them. E.g. https://github.com/pytorch/benchmark/tree/main/torchbenchmark/models/pytorch_struct/legacy.

Ideally we need help from the torchtext team to migrate the two models (pytorch_struct and attention_is_all_you_need_pytorch) to use the new interface, but for now we just use the copied legacy code.

xuzhao9 (changes requested):
Please help migrate the model to use the new torchtext data API.

abhinavarora:
@xuzhao9 Thank you for removing the dependencies and moving them to a local folder. On TorchText we are working with TorchRecipes to write down these recipes. Once done, we can figure out how to write new versions of these models with the latest library. 

Created a tracking issue [here](https://github.com/pytorch/text/issues/1443) for this. For now, I will abandon and close the PR

PR #566, opened 2021-11-16
Fix bug where check_device unit tests are skipped when get_module raises NotImplementedError

aaronenyeshi:
Since we added the check_device() to each of the unit tests: example and
check_device, some unit tests are skipped when get_module() is
NotImplemented. This is because we catch NotImplemented in test.py.
We should catch it in check_device, and not skip the rest of the test.

### DIFF
--- a/torchbenchmark/__init__.py
+++ b/torchbenchmark/__init__.py
@@ -337,7 +337,13 @@ class ModelTask(base_task.TaskBase):
     @staticmethod
     def check_example() -> None:
         model = globals()["model"]
-        module, example_inputs = model.get_module()
+        try:
+            module, example_inputs = model.get_module()
+        except NotImplementedError:
+            # Raise an error for models without a get_module() implementation. Otherwise
+            # all the tests such as train, eval, and example will get skipped in unit testing.
+            raise RuntimeError('Missing get_module() implementation. Required for all models.')
+
         if isinstance(example_inputs, dict):
             # Huggingface models pass **kwargs as arguments, not *args
             module(**example_inputs)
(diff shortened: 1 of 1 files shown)


aaronenyeshi:
For example, when LearningToPaint's get_module() implementation raises NotImplementedError, it will skip all of the tests:
```
~/cluster/work/benchmark$ gpurun python test.py -k LearningToPaint --verbose
test_LearningToPaint_check_device_cpu (__main__.TestBenchmark) ... skipped 'Method check_device is not implemented, skipping...'
test_LearningToPaint_check_device_cuda (__main__.TestBenchmark) ... skipped 'Method check_device is not implemented, skipping...'
test_LearningToPaint_eval_cpu (__main__.TestBenchmark) ... skipped 'Method eval is not implemented, skipping...'
test_LearningToPaint_eval_cuda (__main__.TestBenchmark) ... skipped 'Method eval is not implemented, skipping...'
test_LearningToPaint_example_cpu (__main__.TestBenchmark) ... skipped 'Method get_module is not implemented, skipping...'
test_LearningToPaint_example_cuda (__main__.TestBenchmark) ... skipped 'Method get_module is not implemented, skipping...'
test_LearningToPaint_train_cpu (__main__.TestBenchmark) ... skipped 'Method train is not implemented, skipping...'
test_LearningToPaint_train_cuda (__main__.TestBenchmark) ... skipped 'Method train is not implemented, skipping...'

Ran 8 tests in 70.486s

OK (skipped=8)
```

aaronenyeshi:
Since there are users who require the `get_module()` function, we should raise an error when it is not implemented.

aaronenyeshi:
Also ensure that get_module is implemented in set_train and set_eval paths.

aaronenyeshi:
> Also ensure that get_module is implemented in set_train and set_eval paths.

It looks like adding the try/except block to _set_mode is causing 19 new unit test failures. I think I will pull that one back, and we can submit a separate fix that can address all of those tests together.

aaronenyeshi:
This fix is incomplete and requires implementation of get_module for several models missing it.

PR #572, opened 2021-11-19
Fix the jit code for timm models.

xuzhao9:
We didn't correctly set the jit script for the timm models. This PR makes sure all timm models are correctly jitted.


aaronenyeshi (approved):
LGTM! Thanks for the fix.

PR #574, opened 2021-11-22
Set the correct train batch size for squeezenet.

xuzhao9:
SqueezeNet uses default batch size of 512 for training. As running a 512 batch will usually cause CUDA OOM on a single GPU, we use a combination of `batch_size` and `iter_size` that multiply to 512 as a workaround.

Source: https://github.com/forresti/SqueezeNet

### DIFF
--- a/torchbenchmark/models/squeezenet1_1/__init__.py
+++ b/torchbenchmark/models/squeezenet1_1/__init__.py
@@ -15,12 +15,14 @@ from torchbenchmark.tasks import COMPUTER_VISION
 class Model(BenchmarkModel):
     task = COMPUTER_VISION.CLASSIFICATION
     optimized_for_inference = True
-    # Default train batch size is set to 512, which is batch_size (32) * iter_size (16)
+    # Original train batch size: 512, out of memory on V100 GPU
+    # Use hierarchical batching to scale down: 512 = batch_size (32) * epoch_size (16)
     # Source: https://github.com/forresti/SqueezeNet
     def __init__(self, device=None, jit=False, train_bs=32, eval_bs=16):
         super().__init__()
         self.device = device
         self.jit = jit
+        self.epoch_size = 16
         self.model = models.squeezenet1_1().to(self.device)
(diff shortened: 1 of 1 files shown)


aaronenyeshi:
I think we should discuss this one, I'm not sure if we will get the same quality of profiles without reaching 512 bs.

aaronenyeshi (approved):
Since the paper also uses hierarchical batching, I will give this a stamp! LGTM!

torchbenchmark/models/squeezenet1_1/__init__.py
```diff
@@ -46,7 +48,7 @@ def get_module(self):
     def set_eval(self):
         pass
 
-    def train(self, niter=3):
+    def train(self, niter=16):
```
aaronenyeshi, line 51: May want a small comment here, to avoid confusion.

torchbenchmark/models/squeezenet1_1/__init__.py
```diff
@@ -40,32 +38,22 @@ def __init__(self, device=None, jit=False, train_bs=64, eval_bs=16):
...
-            pred = self.model(*self.example_inputs)
-            y = torch.empty(pred.shape[0], dtype=torch.long, device=self.device).random_(pred.shape[1])
-            loss(pred, y).backward()
-            optimizer.step()
+            for _ in range(self.epoch_size):
+                optimizer.zero_grad()
+                pred = self.model(*self.example_inputs)
+                y = torch.empty(pred.shape[0], dtype=torch.long, device=self.device).random_(pred.shape[1])
+                loss(pred, y).backward()
+                optimizer.step()
```
dzhulgakov, line 53: if you're doing hierarchical batching, `zero_grad` and `.step()` should be outside of the loop. Otherwise this code is just doing the same thing as before :)

xuzhao9: Thanks for the comment! Sorry I misunderstood how to do the hierarchical batching, I will create another PR to fix it.

PR #575, opened 2021-11-22
Fix the train batch size of alexnet.

xuzhao9:
Use the batch size 128 referred in the alexnet paper:
Source: https://arxiv.org/pdf/1404.5997.pdf


aaronenyeshi (approved):
LGTM!

PR #577, opened 2021-11-22
Only load the module specified in run.py argument

xuzhao9:
When running `run.py`, it tries to read and load every model in the `models` directory even though we only need to run one specific model. In practice this often causes problems such as requiring to install `torchvision`/`torchtext` which are not required by the model we are attempting to run.

`run.py` should only load the specific model we are trying to run, instead of loading all the models.

### DIFF
--- a/torchbenchmark/__init__.py
+++ b/torchbenchmark/__init__.py
@@ -464,6 +464,36 @@ def list_models(model_match=None):
                 models.append(Model)
     return models
 
+def find_model_by_name(model):
+    """Find a model by its exact name"""
+    models = filter(lambda x: model.lower() == x.lower(),
+                    map(lambda y: os.path.basename(y), _list_model_paths()))
+    models = list(models)
+    if not models:
+        return None
+    else:
+        assert len(models) == 1, f"Find more than one models {models} with the exact name: {model}"
+        return models[0]
+
+def load_model(model):
+    loaded_models = []
+    models = filter(lambda x: model.lower() == x.lower(),
+                    map(lambda y: os.path.basename(y), _list_model_paths()))
+    models = list(models)
+    assert len(models) == 1, f"Find more than one models {models} with the exact name: {model}"
+    model_name = models[0]
+    try:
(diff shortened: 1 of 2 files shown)


torchbenchmark/__init__.py
```diff
@@ -464,6 +464,26 @@ def list_models(model_match=None):
                 models.append(Model)
     return models
 
+def load_model_by_name(model):
+    models = filter(lambda x: model.lower() == x.lower(),
+                    map(lambda y: os.path.basename(y), _list_model_paths()))
+    models = list(models)
+    if not models:
+        return None
+    assert len(models) == 1, f"Find more than one models {models} with the exact name: {model}"
```
aaronenyeshi, line 473: nit: typo f"Found more than ..."

aaronenyeshi (approved):
Small nit in comment. Also, does this patch prevent us from matching models with partial name match?

This looks like a good speed up for run.py. Might also be useful for gen_summary_metadata.py.

torchbenchmark/__init__.py
xuzhao9: Thanks for point out!

xuzhao9:
> Small nit in comment. Also, does this patch prevent us from matching models with partial name match?
> 
> This looks like a good speed up for run.py. Might also be useful for gen_summary_metadata.py.

Actually I deliberately avoid partial name matching, because I found it is easy to cause human errors. For example, if user runs "Bert" or "maskrcnn", it may implicitly fallback to a model different from user's intention.

aaronenyeshi (approved):
> > Small nit in comment. Also, does this patch prevent us from matching models with partial name match?
> > This looks like a good speed up for run.py. Might also be useful for gen_summary_metadata.py.
> 
> Actually I deliberately avoid partial name matching, because I found it is easy to cause human errors. For example, if user runs "Bert" or "maskrcnn", it may implicitly fallback to a model different from user's intention.

I can see where that is coming from. It sounds good to me!

PR #578, opened 2021-11-23
[WIP] Raise RuntimeError when models are missing get_module() implementation

aaronenyeshi:
Since, we call get_module in each of the unit tests: train, eval,
example, and check_device, the unit tests are skipped when get_module()
is Not Implemented. This is because we catch NotImplemented in test.py.
We should throw error when it is missing in any test.

### DIFF
--- a/torchbenchmark/__init__.py
+++ b/torchbenchmark/__init__.py
@@ -337,7 +337,13 @@ class ModelTask(base_task.TaskBase):
     @staticmethod
     def check_example() -> None:
         model = globals()["model"]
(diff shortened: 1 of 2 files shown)


aaronenyeshi:
This change requires Issue #578 to be fixed.

aaronenyeshi:
@xuzhao9 - I'm wondering if we should skip tests that don't have this implemented, so we can enforce this?

aaronenyeshi:
Closing old prs

PR #580, opened 2021-11-23
Fix the train arch of DLRM

aaronenyeshi:
Use the training arch from the DLRM paper:
Source: https://arxiv.org/pdf/1906.00091.pdf


aaronenyeshi:
Also, fixed some minor linting issues.

xuzhao9 (approved):
LGTM!

PR #581, opened 2021-11-23
Fix the train options of LearningToPaint

aaronenyeshi:
Use the training options from the LearningToPaint paper:
Source: https://arxiv.org/pdf/1903.04411.pdf
Code: https://github.com/megvii-research/ICCV2019-LearningToPaint/blob/master/baseline/train.py

### DIFF
--- a/torchbenchmark/models/LearningToPaint/__init__.py
+++ b/torchbenchmark/models/LearningToPaint/__init__.py
@@ -26,28 +26,33 @@ class Model(BenchmarkModel):
         super(Model, self).__init__()
         self.device = device
         self.jit = jit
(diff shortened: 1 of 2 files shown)


xuzhao9:
Could you please add `train_bs` and `eval_bs` arguments to the `__init__()` function so that they are adjustable at runtime? It could still be useful in, say, in cpu inference we would like to use `eval_bs=1`

aaronenyeshi:
> Could you please add `train_bs` and `eval_bs` arguments to the `__init__()` function so that they are adjustable at runtime? It could still be useful in, say, in cpu inference we would like to use `eval_bs=1`

Done, just added the two options. I also had to separate the model into train and eval to support this.

PR #582, opened 2021-11-24
Add the efficientdet model

xuzhao9:
We are adding the efficientdet model to better cover cases that use dynamic shapes.

Source: https://github.com/rwightman/efficientdet-pytorch
Data: coco2017-minimal (use val for both train and test)
The model only runs on A100 (will CUDA OOM on V100).
It supports AMP, but it is disabled by default.


torchbenchmark/models/detectron2_maskrcnn/__init__.py
```diff
@@ -29,7 +29,6 @@
 class Model(BenchmarkModel):
     task = COMPUTER_VISION.DETECTION
 
-    # This model doesn't support setting batch size for inference
```
xuzhao9, line 32: detectron2 has already added support for batch size inference

xuzhao9:
Hi @erichan1 , can I get a review on this?

erichan1 (approved):

PR #583, opened 2021-11-24
Use the correct batch size for vgg16 training

xuzhao9:
The batch size for train was set to 256. 
Source: https://arxiv.org/pdf/1409.1556.pdf


aaronenyeshi (approved):
LGTM

PR #591, opened 2021-11-29
Fix train batch size for pytorch_struct.

xuzhao9:
Use the original batch size 200 for train.
Source: https://github.com/harvardnlp/pytorch-struct/blob/f4e374e894b94a9411fb3d2dfb44201a18e37b26/notebooks/Unsupervised_CFG.ipynb


aaronenyeshi (approved):
LGTM!

PR #594, opened 2021-11-29
Fix background matting batch size.

xuzhao9:
Original train batch size: 4 (on synthesized data), 8 (on real data)
The hardware environment is unknown.
Original paper link: https://arxiv.org/pdf/2004.00626.pdf


xuzhao9:
Profile:
image

Trace:
[Uploading background_matting.tar.gz…]()

xuzhao9:
The bad GPU utilization is cased by the original model code: https://github.com/senguptaumd/Background-Matting/blob/master/train_real_fixed.py#L212-L241. Should we keep it as-is or comment that part?

aaronenyeshi (approved):
LGTM! I think we should match original implementation for train.

PR #595, opened 2021-11-29
Update the train_bs and eval_bs for super_slomo.

xuzhao9:
Original code link: https://github.com/avinashpaliwal/Super-SloMo/blob/master/train.ipynb
Original paper link: https://arxiv.org/pdf/1712.00080.pdf
Train batch size: 6
Eval batch size: 10

Eval profile:
image

Train profile:
image

### DIFF
--- a/torchbenchmark/models/Super_SloMo/__init__.py
+++ b/torchbenchmark/models/Super_SloMo/__init__.py
@@ -66,11 +66,11 @@ class Model(BenchmarkModel):
             shuffle=False)
 
         trainData, trainFrameIndex = next(iter(trainloader))
-        trainData = _prefetch(trainData)
+        trainData = _prefetch(trainData, self.device)
         self.example_inputs = trainFrameIndex.to(self.device), *trainData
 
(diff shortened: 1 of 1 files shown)


torchbenchmark/models/Super_SloMo/__init__.py
```diff
@@ -30,7 +41,7 @@ def __init__(self, device=None, jit=False):
         root = str(Path(__file__).parent)
         self.args = args = Namespace(**{
             'dataset_root': f'{root}/dataset',
-            'train_batch_size': 6,
+            'train_batch_size': train_bs,
```
aaronenyeshi, line 44: I think you can delete this line since there are no references to `self.args.train_batch_size`. Or modify the references below to use this, and potentially `self.args.eval_batch_size`

aaronenyeshi (approved):
LGTM, small comment/nit.

torchbenchmark/models/Super_SloMo/__init__.py
xuzhao9: Fixed. Thanks for pointing out!

PR #596, opened 2021-11-29
Remove V0 workflow. Use CUDA 11.3 for V1 workflow.

xuzhao9:
CUDA 10.2 is not supported soon and we are using CUDA 11.3 for both V1 and V2 scores.
We are also upgrading to Python 3.8.

This PR also removes V0 workflows, as V0 will not be maintained (e.g. won't backport the BC-breaking torchvision/torchtext fixes to V0).

Upgrading to CUDA 11.3 may cause score change in V1 score, and we will monitor if any regression/optimization is overshadowed by the bump of the CUDA version.


wconstab:
no problem. we are currently based off 'main', but, we should probably consider using v1 for a stable branch.  we don't use v0.

xuzhao9:
@bitfort Can I get a review on this?

bitfort (approved):
Thanks!

PR #599, opened 2021-11-30
quick fix to check_lazy

Krovatkin:


wconstab (approved):
this looks correct; i put the imports inside _check_model to avoid loginfo spam on the main process, IIRC.  You could try putting the imports into the part of the main process that checks `if args.check_model` before constructing the model, if there is still a verbosity issue.  It may not be worth worrying about the verbose messages either, i'm fine merging as is

PR #456, opened 2021-09-10
[LTC][Not for merge] Enabling Lazy Tensor in hf_Bert and resnet50

alanwaketan:
This patch tries to enable Lazy Tensor in hf_Bert/resnet50 and allow overall torchbenchmark to use 'lazy' device.

PR #600, opened 2021-11-30
Added lazy tensor testing CI.

xuzhao9:
Setup initial framework for lazy tensor nightly testing.


alanwaketan:
You probably need some the `bisection.py` part from my PR here: https://github.com/pytorch/benchmark/pull/456/files.

torchbenchmark/models/hf_Bert/__init__.py
```diff
@@ -46,6 +55,8 @@ def train(self, niter=3):
             loss = outputs.loss
             loss.backward()
             self.optimizer.step()
+            if self.device == 'lazy':
```
xuzhao9, line 58: Can we place lazy_tensor related code in a separate python file so that if we need to change it in the future we don't need to change every model's code? E.g. in `torchbenchmark/util/lazy_tensor.py` we have a function `lazy_step()` that does this. @alanwaketan

wconstab: I am not relying on mark_step in this file.  In fact on my own branch I deleted this change.

I wrote another script (lazy_bench.py) which iters over torchbench models and calls the sync functions at the right times.

There will be some small changes needed in main torchbench model code, only for particular models where some hardcoded 'cuda' or 'cpu' assumption is baked in and we need to loosen it.

xuzhao9: Addressed, can you help take another look?

wconstab (approved):
lgtm, thanks!

PR #601, opened 2021-12-01
A hack to run Demucs with lazy 

Krovatkin:


wconstab:
looks reasonable; it should still be possible to jit.script the Demucs nn.module in the torchbench __init__.py, right?  (so, as far as hacks go it's not too bad?)  If so lets get it checked into benchmark main.

wconstab (approved):

PR #607, opened 2021-12-02
Fix the runner temp interface.

xuzhao9:
After updating the infra instances, the thermal file path is changed. This PR updates the new thermal file path so that we can get the CPU temp information.


erichan1 (approved):
LGTM!

PR #616, opened 2021-12-07
[wconstab/ltc] Try running check_lazy.py on the CI

alanwaketan:
This pull request tries to update the wconstab/ltc such that we can run some custom scripts in the TB CI.


bisection.py
```diff
@@ -216,7 +216,7 @@ def _build_lazy_tensor(self, commit: Commit, build_env: Dict[str, str]):
             lazy_tensor_path = os.path.join(self.srcpath, "lazy_tensor_core")
             command = "./scripts/apply_patches.sh"
             subprocess.check_call(command, cwd=lazy_tensor_path, env=build_env, shell=True)
-            command = "python setup.py install"
+            command = "DEBUG=1 USE_CUDA=1 USE_XNNPACK=0 python setup.py install"
```
wconstab, line 219: interesting.. what is the deal with XNNPACK?  I don't remember setting it to zero locally, but I also didn't notice any issues from it.

bisection.py
```diff
@@ -553,7 +553,7 @@ def output_abtest_result(self):
                                     targets=targets,
                                     output_json=args.output,
                                     devbig=args.devbig,
-                                    build_lazy=args.build_lazy,
+                                    build_lazy=True,
```
xuzhao9, line 556: There is no need to hardcode build_lazy here because you can specify it in run-bisection.sh.

.github/scripts/run.sh
```diff
@@ -36,13 +36,15 @@ sudo nvidia-smi -ac ${GPU_FREQUENCY}
...
-    taskset -c "${CORE_LIST}" pytest test_bench.py -k "${BENCHMARK_FILTER}" \
-            --benchmark-min-rounds "${NUM_ROUNDS}" \
-            --benchmark-json ${DATA_DIR}/${DATA_JSON_PREFIX}_${c}.json
-done
+# for c in $(seq 1 $NUM_ITER); do
+#     taskset -c "${CORE_LIST}" pytest test_bench.py -k "${BENCHMARK_FILTER}" \
+#             --benchmark-min-rounds "${NUM_ROUNDS}" \
+#             --benchmark-json ${DATA_DIR}/${DATA_JSON_PREFIX}_${c}.json
+# done
+taskset -c "${CORE_LIST}" python check_lazy.py --output_file ${DATA_DIR}/sweep.out
```
xuzhao9, line 48: You will need to add to `https://github.com/pytorch/benchmark/blob/main/.github/workflows/lazy-tensor-nightly.yml` on main branch to upload the result as github actions artifact.

alanwaketan: Sure, thanks for the suggestion. We will rely on the print first, and then move on to this route.

bisection.py
alanwaketan: I was told by @Krovatkin that this is needed...

bisection.py
alanwaketan: Will fix that in my PyTorch branch.

bisection.py
alanwaketan: Just saw that the default configuration will turn it off. So I removed then. See: setup_build_env in the same file.

bisection.py
```diff
@@ -459,14 +459,15 @@ def run(self):
...
                 targets = right.digest.keys()
-            updated_targets = self.regression(left, right, targets)
-            if len(updated_targets):
-                mid = self.torch_src.get_mid_commit(left, right)
-                if mid == None:
-                    self.result.append((left, right))
-                else:
-                    self.bisectq.append((left, mid, updated_targets))
-                    self.bisectq.append((mid, right, updated_targets))
+            # Commented out for now.
```
wconstab, line 462: is this because bisection.py isn't easy to comment out entirely in some other script?  I guess we won't benefit from bisection at all until we start supporting the regular torchbench timing scripts.

alanwaketan: That's the case. I just want to reuse the building process of bisection.py.

bisection.py
```diff
@@ -181,7 +181,7 @@ def get_mid_commit(self, left: Commit, right: Commit) -> Optional[Commit]:
     def setup_build_env(self, env) -> Dict[str, str]:
         env["USE_CUDA"] = "1"
         env["BUILD_CAFFE2_OPS"] = "0"
-        env["USE_XNNPACK"] = "0"
+        env["USE_XNNPACK"] = "1" # Customized for LTC.
```
wconstab, line 184: it's really confusing 
1) we need XNNPACK to make torchbench happy
   (even though we don't understand why XNNPack is being used, since we're benchmarking cuda)
2) XNNPACK is disabled for normal torchbench CI

???

wconstab (approved):

alanwaketan:
Thanks Will for approving the PR.

PR #618, opened 2021-12-07
Add the yolov5l model

xuzhao9:
Flavour of the model: yolov5lv6
Input data: coco128
train batch size: 16, profile: 
image

eval batch size: 128, profile: 
image


xuzhao9:
Train profiles without pre-fetcher:
image


Eval profiles without pre-fetcher:
image

aaronenyeshi (changes requested):
Do we need all the various yaml settings in `torchbenchmark/models/yolov5/yolov5/models/hub/`?

aaronenyeshi:
Should yolov5s.pt be moved to data?

PR #619, opened 2021-12-08
add timestamps to model launches

Krovatkin:


check_lazy.py
```diff
@@ -61,6 +61,7 @@ def sweep_models(output_filename, tests=['eval', 'train']):
                     print(f"{datetime.strptime(dt,'%Y-%m-%d %H:%M:%S')} : Running launch_command {' '.join(launch_command)}")
                     rc = subprocess.call(launch_command,
                                         env=env,
+                                        timeout = 1800, # 30 minutes, 10 min max per iter
```
wconstab, line 64: great idea.  although.. why so long even, is there any model that should take more than a few min per iter?  I suspect we have some kind of unsolved issue with process startup or something.

wconstab (approved):
yea lets merge this, but, use a 2 min timeout? 
we do have about 100 tests we check (model * [eval, train])
200 min is already over 3 hours!

check_lazy.py
```diff
@@ -61,7 +61,7 @@ def sweep_models(output_filename, tests=['eval', 'train']):
                     print(f"{datetime.strptime(dt,'%Y-%m-%d %H:%M:%S')} : Running launch_command {' '.join(launch_command)}")
                     rc = subprocess.call(launch_command,
                                         env=env,
-                                        timeout = 1800, # 30 minutes, 10 min max per iter
+                                        timeout = 120, # 30 minutes, 10 min max per iter
```
alanwaketan, line 64: You didn't update the comment. Can we tell if the model times out or not in the artifacts?

PR #623, opened 2021-12-08
[wconstab/ltc] Convert the json output from check_lazy.py to csv

alanwaketan:
Convert the json output from check_lazy.py to csv.


bisection.py
```diff
@@ -181,7 +181,7 @@ def get_mid_commit(self, left: Commit, right: Commit) -> Optional[Commit]:
     def setup_build_env(self, env) -> Dict[str, str]:
         env["USE_CUDA"] = "1"
         env["BUILD_CAFFE2_OPS"] = "0"
-        env["USE_XNNPACK"] = "1" # Customized for LTC.
+        env["USE_XNNPACK"] = "0" # Customized for LTC.
```
alanwaketan, line 184: This shouldn't be merged.

.github/scripts/run.sh
```diff
@@ -47,9 +47,16 @@ export GOMP_CPU_AFFINITY="${CORE_LIST}"
 # done
 
 echo "Running check_lazy.py"
-python check_lazy.py --output_file ${DATA_DIR}/sweep.json
-# TODO(alanwaketan): Enable the below format conversion.
-# At the same time, we can workaround it by running the command by hands.
-# python check_lazy.py --json_to_csv ${DATA_DIR}/sweep.json --output_file ${DATA_DIR}/sweep.csv
+# The output is a file full of JSON objects but not legit .JSON.
+python check_lazy.py --output_file ${DATA_DIR}/sweep.out
+# Ignore the error for this command such that we can still have the json file
```
wconstab, line 52: what error are you ignoring? i never understood what was wrong with running the --json_to_csv before. but it looks like you got it working since I found a .csv in the artifacts you sent me.

wconstab (approved):
thanks!

.github/scripts/run.sh
alanwaketan: It's bogus. Let me remove them.

alanwaketan:
Thanks for approving this PR, Will.

PR #628, opened 2021-12-09
Fix several bugs in the bisection script.

xuzhao9:
This PR improves the bisector:
1. Use "main" branch instead of "master"
2. If pytorch build fails, delete the build directory, and retry
3. Enable XNNPACK, disable TEST in build
4. Cleanup all the staled packages before building the new commit.

### DIFF
--- a/bisection.py
+++ b/bisection.py
@@ -12,6 +12,7 @@ Usage:
 
 import os
 import json
+import shutil
 import yaml
 import argparse
 import typing
@@ -144,6 +145,8 @@ class TorchSource:
         if not repo_origin_url == TORCH_GITREPO:
             print(f"WARNING: Unmatched repo origin url: {repo_origin_url} with standard {TORCH_GITREPO}")
         self.update_repos()
+        # Clean up the existing packages
+        self.cleanup()
         return True
 
     # Update pytorch, torchtext, and torchvision repo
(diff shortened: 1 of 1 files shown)


bisection.py
```diff
@@ -237,22 +241,34 @@ def build(self, commit: Commit):
...
 
-    def cleanup(self, commit: Commit):
-        print(f"Cleaning up packages from commit {commit.sha} ...", end="", flush=True)
+    def cleanup(self):
         packages = ["torch", "torchtext", "torchvision"]
-        command = "pip uninstall -y " + " ".join(packages) + " &> /dev/null "
-        subprocess.check_call(command, shell=True)
+        CLEANUP_ROUND = 5
+        # Clean up multiple times to make sure the packages are all uninstalled
+        for _ in range(CLEANUP_ROUND):
```
erichan1, line 264: Curious about this - there's a possibility of uninstall failing?

erichan1 (approved):
LGTM!

alanwaketan (approved):

bisection.py
erichan1: To clarify, my question is why do we need 5 cleanup rounds?

xuzhao9: > To clarify, my question is why do we need 5 cleanup rounds?

This is because there might be multiple versions of staled torch packages in the environment - usually uninstalling twice is sufficient, but it does no harm to do it more times just to be safe.

PR #630, opened 2021-12-10
Fix yolov3 model training to original batch size

xuzhao9:
Train profile:
image


aaronenyeshi (approved):
LGTM! Thank you for fixing this!

PR #631, opened 2021-12-10
Allow install.py to only install deps of specified models.

xuzhao9:
This PR fixes the following problems of the main `install.py`:
- Crash if the model directory doesn't have `install.py` script (should skip instead, because when switching between branches, such as v1.0 and main, some directories will be remained there by git, such as fastNLP in v1.0 will stay as an empty folder after switching to main)
- Doesn't support installing a specific model


aaronenyeshi (approved):
LGTM! This is really useful! Thank you, I was wondering how to install a specific model's deps.

PR #634, opened 2021-12-10
Improve the quality of tts_angular, fix train and eval batch sizes

xuzhao9:
Train (bs=64):
image

Train bs=64 is default for the model, source: https://github.com/mozilla/TTS/blob/master/TTS/speaker_encoder/config.json#L38

Eval (bs=256):
image

256 is the largest BS supported by T4.


aaronenyeshi (approved):
LGTM!

PR #637, opened 2021-12-10
[dlrm] Unconditionally move inputs to the specified device

alanwaketan:
This commit moves the X.to one step above to better support devices other
than CUDA and CPU.

### DIFF
--- a/torchbenchmark/models/dlrm/__init__.py
+++ b/torchbenchmark/models/dlrm/__init__.py
@@ -163,8 +163,8 @@ class Model(BenchmarkModel):
 
         # Preparing data
         X, lS_o, lS_i, self.targets = next(iter(self.train_ld))
+        X = X.to(self.device)
         if self.device == "cuda":
-            X = X.to(self.device)
             lS_i = [S_i.to(self.device) for S_i in lS_i] if isinstance(lS_i, list) \
                 else lS_i.to(self.device)
             lS_o = [S_o.to(self.device) for S_o in lS_o] if isinstance(lS_o, list) \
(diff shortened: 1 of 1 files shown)


torchbenchmark/models/dlrm/__init__.py
```diff
@@ -163,8 +163,8 @@ def __init__(self, device=None, jit=False):
 
         # Preparing data
         X, lS_o, lS_i, self.targets = next(iter(self.train_ld))
+        X = X.to(self.device)
         if self.device == "cuda":
-            X = X.to(self.device)
             lS_i = [S_i.to(self.device) for S_i in lS_i] if isinstance(lS_i, list) \
```
wconstab, line 168: don't we also need to move S_i and S_o to lazy device?

maybe its better to just change the conditional to 
`if self.device != "cpu":`
?

alanwaketan: Fixed.

wconstab (approved):

xuzhao9 (approved):

torchbenchmark/models/dlrm/__init__.py
```diff
@@ -163,7 +163,7 @@ def __init__(self, device=None, jit=False):
 
         # Preparing data
         X, lS_o, lS_i, self.targets = next(iter(self.train_ld))
-        if self.device == "cuda":
+        if self.device != "cpu":
```
xuzhao9, line 166: One question, can we just delete this condition? I guess if `self.device == "cpu"`, `X = X.to(self.device)` is just no-op?

alanwaketan: Yea, seems even better.

PR #639, opened 2021-12-11
Improve the quality of demucs model

xuzhao9:
Original train batch size is 64, which doesn't work on T4. CPU training is too slow, so disable both GPU and CPU training.

Eval batch size 8 looks good:
image

### DIFF
--- a/torchbenchmark/models/demucs/__init__.py
+++ b/torchbenchmark/models/demucs/__init__.py
@@ -19,8 +19,8 @@ from typing import Optional, Tuple
 torch.manual_seed(1337)
 random.seed(1337)
 np.random.seed(1337)
-torch.backends.cudnn.deterministic = True
-torch.backends.cudnn.benchmark = False
+torch.backends.cudnn.deterministic = False
+torch.backends.cudnn.benchmark = True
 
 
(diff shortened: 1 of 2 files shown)


torchbenchmark/models/demucs/__init__.py
```diff
@@ -38,22 +38,25 @@ def forward(self, streams) -> Tuple[Tensor, Tensor]:
...
+        self.model = Demucs(channels=64)
         self.dmodel = self.model
         self.model.to(device)
         self.optimizer = torch.optim.Adam(self.model.parameters(), lr=args.lr)
 
         if 1:
             samples = 80000
-            # TODO: calculate the right shape
-            self.example_inputs = (torch.rand([4, 5, 2, 135576], device=device),)
+            # TODO: enable GPU training after it is supported by infra
```
aaronenyeshi, line 57: Do we want to link a GH Issue as well?

aaronenyeshi (approved):
LGTM!

torchbenchmark/models/demucs/__init__.py
xuzhao9: Yes - linked to GH issue #652.

PR #640, opened 2021-12-11
Improve the quality of quantized_qat models.

xuzhao9:
Improves the quality of quantized model training:
- Added train_bs argument for easier tuning
- Keep train batch size consistent with the paper 

Profile for mobilenet_v2_quantized_qat train (bs=96):
image

Profile for resnet50_quantized_qat train (bs=32):
image


aaronenyeshi (approved):
LGTM! Thank you!

aaronenyeshi:
do we need to add eval_bs as well? Or is that separate?

xuzhao9:
> do we need to add eval_bs as well? Or is that separate?

eval doesn't support GPU so we are not adding the parameter. The planned design is, if `eval_bs` exists in the `__init__()` parameter, the CPU workload will set `eval_bs` to 1.

PR #641, opened 2021-12-13
Avoid instantiating models for the sake of checking their names

Krovatkin:


wconstab (approved):
Nice, glad you thought of this.

PR #643, opened 2021-12-14
[wconstab/ltc] Make nvidia_deeprecommender accept lazy device

alanwaketan:
This pull request makes nvidia_deeprecommender accept lazy device.

It just fixes the issue in the ltc branch instead of main since I don't think it's applicable to make the model robust to unknown device.


wconstab (approved):
lgtm! 

might be better to not do the formatting changes (or keep bulk-formatting changes in separate PRs so its less noisy to review the actual changes).

I would say it's fine to make these changes on main branch, but its not urgent.  eventually we may take the whole diff of wconstab/ltc and merge important bits to main.  Thats another reason that having cosmetic differences won't help us.

alanwaketan:
Thanks, Will.

PR #644, opened 2021-12-14
Improves the model quality of huggingface T5

xuzhao9:
This PR makes the following changes:
- Disable train because it doesn't fit on T4 (default bs=8)
- Use the largest eval bs=1 which fits on T4

Profiling of eval when bs=1:
image


torchbenchmark/models/hf_T5/__init__.py
```diff
@@ -23,22 +21,25 @@ def forward(self, input_ids, decoder_input_ids):
...
 
-        torch.manual_seed(42)
         config = AutoConfig.from_pretrained("t5-small")
         self.model = AutoModelForSeq2SeqLM.from_config(config).to(device)
         self.optimizer = optim.Adam(self.model.parameters(), lr=0.001)
 
-        input_ids = torch.randint(0, config.vocab_size, (2, 1024)).to(device)
-        decoder_ids = torch.randint(0, config.vocab_size, (2, 1024)).to(device)
+        # input_ids = torch.randint(0, config.vocab_size, (train_bs, 1024)).to(device)
+        # decoder_ids = torch.randint(0, config.vocab_size, (train_bs, 1024)).to(device)
```
aaronenyeshi, line 38: Should we delete the commented out parts?

aaronenyeshi (approved):
LGTM!

torchbenchmark/models/hf_T5/__init__.py
xuzhao9: We can keep it there so it is easier to re-enable the train

PR #645, opened 2021-12-14
Improve the quality of soft actor critic model

xuzhao9:
- Fix the train batch size to be 256
- Add comments that eval batch size can't be adjusted
- This model doesn't support prefetching either, because the train data is dynamically generated at runtime according to the environment response

Profile of train, bs=256:
image

Profile of eval, bs=1:
image


aaronenyeshi (approved):
LGTM! thank you

PR #647, opened 2021-12-14
[wconstab/ltc] Add in-memory test expectations

alanwaketan:
Summary:
This commit adds some in-memory test expectations to filter out
tests that don't exist.

Test Plan:
CI.

### DIFF
--- a/check_lazy.py
+++ b/check_lazy.py
@@ -22,6 +22,10 @@ import lazy_tensor_core
 import datetime
 lazy_tensor_core._LAZYC._ltc_init_ts_backend()
 
+# The following models don't have the corresponding tests.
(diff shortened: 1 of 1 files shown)


wconstab (approved):
LGTM! i was going to suggest handling a particular type of exception because I thought we used something like 'NotImplementedError' in torchbench.  But your idea probably works better given how the subprocess is set up.

alanwaketan:
> LGTM! i was going to suggest handling a particular type of exception because I thought we used something like 'NotImplementedError' in torchbench. But your idea probably works better given how the subprocess is set up.

The reason why I didn't do the `NotImplementedError` way is that there are too many such error. And doing a string comparison on the exact reason seems weird. Thanks for approving the change.

PR #648, opened 2021-12-14
Improve quality of pytorch_stargan

xuzhao9:
Default train batch size: 16
Source: https://github.com/yunjey/stargan/blob/94dd002e93a2863d9b987a937b85925b80f7a19f/main.py#L73

Added data prefetch and eval batch size support.

Train profile (bs=16):
image

Eval profile (bs=16):
image

### DIFF
--- a/torchbenchmark/models/pytorch_stargan/__init__.py
+++ b/torchbenchmark/models/pytorch_stargan/__init__.py
@@ -1,10 +1,8 @@
 #!/usr/bin/env python
-import random
-import argparse
-import torch
 import os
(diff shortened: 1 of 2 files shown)


aaronenyeshi (approved):
LGTM, thanks! Why is eval_bs not supported? Could we have two `config` variables, for each train and eval?

xuzhao9:
> LGTM, thanks! Why is eval_bs not supported? Could we have two `config` variables, for each train and eval?

Added eval batch size support and prefetch.

PR #653, opened 2021-12-15
Remove torch requirements in requirements.txt

xuzhao9:
This PR removes all `torch` and its domain packages from `requirements.txt`.
It also adds check in `install.py` to make sure they are installed before installing the benchmark.


wconstab (approved):
awesome, thanks @xuzhao9!

PR #654, opened 2021-12-15
[wconstab/ltc] Double the time out limits

alanwaketan:
Double the time out limits since some legit models time out with the current limit.


wconstab (approved):

alanwaketan:
Thanks, Will.

PR #656, opened 2021-12-16
Improve the quality of speech transformer

xuzhao9:
Use original train batch size 32.
Source: https://github.com/kaituoxu/Speech-Transformer/blob/e6847772d6a786336e117a03c48c62ecbf3016f6/src/bin/train.py#L68

Added code to prefetch the data to GPU.

This model doesn't support changing inference batch size.

Profile of train:
image

Profile of eval: 
image

### DIFF
--- a/torchbenchmark/models/speech_transformer/config.py
+++ b/torchbenchmark/models/speech_transformer/config.py
@@ -102,7 +102,7 @@ class SpeechTransformerTrainConfig:
         self.data_loader = self.tr_loader if not SpeechTransformerTrainConfig.cross_valid else self.cv_loader
         if prefetch:
             result = []
-            for data in zip(range(num_train_bs), self.data_loader):
+            for _batch_num, data in zip(range(num_train_bs), self.data_loader):
                 padded_input, input_lengths, padded_target = data
                 padded_input = padded_input.cuda()
                 input_lengths = input_lengths.cuda()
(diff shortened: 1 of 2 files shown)


torchbenchmark/models/speech_transformer/config.py
```diff
@@ -53,29 +53,30 @@ class SpeechTransformerTrainConfig:
     visdom_lr = 0
     visdom_epoch = 0
     visdom_id = 0
+    cross_valid = False
     # The input files. Their paths are relative to the directory of __file__
     train_json = "input_data/train/data.json"
     valid_json = "input_data/dev/data.json"
     dict_txt = "input_data/lang_1char/train_chars.txt"
-    def __init__(self):
+    def __init__(self, prefetch=True, train_bs=16, num_train_batch=1):
```
aaronenyeshi, line 61: Why is train_bs defaulted to 16 here? Also, I think it should set self.batch_size, otherwise that won't be used.

torchbenchmark/models/speech_transformer/config.py
```diff
@@ -34,7 +34,7 @@ class SpeechTransformerTrainConfig:
     label_smoothing = 0.1
     # minibatch
     shuffle = 1
-    batch_size = 16
+    batch_size = 32
```
aaronenyeshi, line 37: Who is using this value? It looks to be replaced by train_bs instead.

aaronenyeshi (approved):
LGTM, just a few small comments regarding variable batch_size and train_bs.

torchbenchmark/models/speech_transformer/config.py
xuzhao9: No, this value is not used. I will remove this. Thanks for pointing out!

torchbenchmark/models/speech_transformer/config.py
xuzhao9: This is the old value I set. Should remove the default value. Fixed.

PR #659, opened 2021-12-16
Improve the quality of moco train

aaronenyeshi:
Add comments to moco to specify where the batch size comes from.
The moco paper is using batch size of 256 over 8 GPUs, which is 32 per GPU.
Source: https://arxiv.org/pdf/1911.05722.pdf

### DIFF
--- a/torchbenchmark/models/moco/__init__.py
+++ b/torchbenchmark/models/moco/__init__.py
@@ -20,19 +20,24 @@ from torchbenchmark.tasks import OTHER
 
 torch.manual_seed(1058467)
 random.seed(1058467)
-cudnn.deterministic = True
+cudnn.deterministic = False
+cudnn.benchmark = True
 
 
 class Model(BenchmarkModel):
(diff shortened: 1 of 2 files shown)


xuzhao9 (commented):
Does this model support adjusting eval batch sizes? If so, we should increase the batch size for eval to saturate GPU. Also, could you please provide the profiling result of train and eval?

aaronenyeshi:
> Does this model support adjusting eval batch sizes? If so, we should increase the batch size for eval to saturate GPU. Also, could you please provide the profiling result of train and eval?

I will look into adding eval_bs.

Here are the profiles on A100:
Train: 
image

Eval: 
image

aaronenyeshi:
> Does this model support adjusting eval batch sizes? If so, we should increase the batch size for eval to saturate GPU. Also, could you please provide the profiling result of train and eval?

I've added a separate eval_bs to the model.

xuzhao9 (approved):
LGTM. Thanks Aaron!

Issue #657, opened 2021-12-16
Enable inference optimizations for models that support it


dzhulgakov:
Some models code has support for inference specific optimization, like fp16:

https://github.com/pytorch/benchmark/blob/1ef02faa0396d8b1d9ddf56796563e59ba36627b/torchbenchmark/models/tacotron2/waveglow/inference.py#L52
https://github.com/pytorch/benchmark/blob/0954aa616e584a63d60cff6fd56ab40f0ca928a2/torchbenchmark/models/yolov3/__init__.py#L63

We should enable it `eval` settings to be representative.

Broader question is whether it's fair to enable `.half()` for models that don't have it natively. It might be considered a very common optimization so it's reasonable to do so for at least the simple ones (like torchvision ones). More intrusive optimizations aren't "fair" because real users are less likely to do so in their code unguided (and today PyTorch doesn't really guide them)

xuzhao9:
I submitted https://github.com/pytorch/benchmark/pull/660 to enable `half()` or fp16 for inference.
I see some models also support `amp`, is it the same as `half()`?

Also, I see models could enable `amp` or `half()` for training but they are usually not the default. Should we enable them for train by default as well?

xuzhao9:
Added https://github.com/pytorch/benchmark/pull/663, enabling `half()` for all torchvision models.

xuzhao9:
Added https://github.com/pytorch/benchmark/pull/776, support amp mode for all models, but not enabled by default because of the concerns.

xuzhao9:
Resolved by https://github.com/pytorch/benchmark/pull/776. Now all models support `amp` and we provide option to enable `half` for models that support it.

PR #660, opened 2021-12-17
Enable fp16 run for yolov3 and tacotron2.

xuzhao9:
Address part of #657. Enable `half()`/`fp16` for inference on Tacotron2 and yolov3.


xuzhao9:
This will be superceded by https://github.com/pytorch/benchmark/pull/776.

PR #668, opened 2021-12-20
Add TFLOPS measure for resnet50 model

xuzhao9:
Eager result with batch size 128 on V100:
```
$ python run.py resnet50 -t eval -d cuda --flops --bs 128
Running eval method from resnet50 on cuda in eager mode.
Unsupported operator aten::max_pool2d encountered 1 time(s)
Unsupported operator aten::add_ encountered 16 time(s)
GPU Time:             64.208 milliseconds
CPU Dispatch Time:    13.305 milliseconds
CPU Total Wall Time:  64.216 milliseconds
FLOPS:               16.5236 TFLOPs per second
```

Fx2trt result with batch sizer 128 on V100:
```
GPU Time:             17.779 milliseconds
CPU Dispatch Time:     0.465 milliseconds
CPU Total Wall Time:  17.778 milliseconds
FLOPS:               59.6853 TFLOPs per second
```

The hardware roofline for V100 is 120 TFLOPs per second.


run.py
```diff
@@ -83,8 +85,14 @@ def run_one_step(func, nwarmup=WARMUP_ROUNDS):
         t0 = time.time_ns()
         func()
         t1 = time.time_ns()
+        wall_latency = t1 - t0
         print('{:<20} {:>20}'.format("CPU Total Wall Time:", "%.3f milliseconds" % ((t1 - t0) / 1_000_000)), sep='')
 
+    # if model_flops is not None, output the TFLOPs per sec
+    if model_flops:
+        tflops = wall_latency / 1_000_000_000 * model_flops / 1.0e12
```
wushirong, line 93: a bit weird here, the way we count tflops is batch_iter * flops_count/ time_sec/1.0e12

yinghai: Yeah, notice it's TFLOPs per second so we need to divide by latency?

torchbenchmark/models/resnet50/__init__.py
```diff
@@ -43,6 +43,13 @@ def __init__(self, device=None, jit=False, train_bs=32, eval_bs=32, extra_args=[
             self.eval_model.eval()
             self.eval_model = torch.jit.optimize_for_inference(self.eval_model)
 
+    def get_flops(self, test='eval'):
```
yinghai, line 46: We need to time the result by 2 due to how we count the FLOPS differently. In out context, one fused multiplication add (FMA) counts as 2 flops instead of one.

xuzhao9: Thanks. Addressed.

run.py
xuzhao9: Fixed the formula. We always run a single batch in torchbench so `batch_iter` should always be 1.

xuzhao9:
@yinghai  and @wushirong , can I get another review on this PR?

PR #670, opened 2021-12-23
[wconstab/ltc] Skip timm_nfnet training

alanwaketan:
Summary:
This commit skip timm_nfnet training test given it causes OOM on CUDA eager.

Test Plan:
CI.


alanwaketan:
FYI, here is the reference: https://github.com/pytorch/benchmark/blob/main/torchbenchmark/models/timm_nfnet/__init__.py#L90.

Krovatkin (approved):
:shipit:

alanwaketan:
Thanks, Nick.

PR #671, opened 2021-12-23
Enable V2 workflow

xuzhao9:
This PR enables V2 nightly workflow on CI.

How we determine test stability:
1. Run the tests on PyTorch 20211222 for 20 times
2. For each test, check if its max delta `(max - min) / min` is above a threshold (we use 7% as threshold)
3. If yes, we determine the test is unstable

In total, we have 210 stable tests and 74 unstable tests, including 71 unstable cpu tests, and 3 unstable CUDA tests.

Example of the V2 workflow: https://github.com/pytorch/benchmark/actions/runs/1655586619
It takes ~5h to finish the entire suite.


erichan1 (approved):
LGTM!

PR #673, opened 2021-12-25
initial torchtext refactor.

erip:
Updates to newer torchtext APIs for pytorch/text#1443

### DIFF
--- a/torchbenchmark/models/pytorch_struct/__init__.py
+++ b/torchbenchmark/models/pytorch_struct/__init__.py
@@ -2,18 +2,18 @@
 pytorch_struct model, Unsupervised CFG task
 https://github.com/harvardnlp/pytorch-struct/blob/master/notebooks/Unsupervised_CFG.ipynb
 """
-import os
 import pytest
-import torchtext
 import numpy as np
 import torch, random
-import torch_struct
 from torch_struct import SentCFG
 from .networks.NeuralCFG import NeuralCFG
 
-from torchbenchmark.util.torchtext_legacy.field import Field
-from torchbenchmark.util.torchtext_legacy.datasets import UDPOS
-from torchbenchmark.util.torchtext_legacy.iterator import BucketIterator
+from torchtext.datasets import UDPOS
+from torchtext.vocab import build_vocab_from_iterator
+from torchtext.data.utils import (
+    get_tokenizer,
+    ngrams_iterator,
+)
 
 from ...util.model import BenchmarkModel
 from torchbenchmark.tasks import OTHER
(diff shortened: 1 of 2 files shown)


erip:
This is not exactly at feature parity with the legacy API because the `BucketIterator` abstraction no longer exists. I took what seems to be the largest individual batch size as the new batch size, but it's not a 1-for-1 refactor.

torchbenchmark/models/pytorch_struct/__init__.py
```diff
@@ -24,105 +25,99 @@
...
-      torch.nn.utils.clip_grad_norm_(self.model.parameters(), 3.0)
-      self.opt.step()
-
-  def eval(self, niter=1):
-    raise NotImplementedError("Eval is not supported by this model")
+    task = OTHER.OTHER_TASKS
+
+    # Original train batch size: 200
+    # Source: https://github.com/harvardnlp/pytorch-struct/blob/f4e374e894b94a9411fb3d2dfb44201a18e37b26/notebooks/Unsupervised_CFG.ipynb
+    def __init__(self, device=None, jit=False, train_bs=40):
```
xuzhao9, line 53: May I ask why you are decreasing the batch size from 200 (used in the original code) to 40?

torchbenchmark/models/pytorch_struct/__init__.py
```diff
@@ -24,105 +25,99 @@
...
+        self.jit = jit
+
+        train_iter = UDPOS(split='train')
+        vocab = build_vocab_from_iterator(
+            yield_tokens(train_iter), specials=["<unk>", "<bos>", "<eos>"], min_freq=3
+        )
+        vocab.set_default_index(vocab["<unk>"])
+
+        self.train_data = DataLoader(
+          UDPOS(split='train'),
```
xuzhao9, line 65: Can we have a variable, `prefetch`, to control whether we prefetch the data to GPU in `train()`? Since in TorchBench we only run 1 step (1 epoch and 1 batch), if we don't prefetch data, most of the benchmarking time will be spent on dataloading, making it less sensitive to performance regressions/optimizations for the computation code. So we want to have a variable to control whether we want to prefetch the data to device.

xuzhao9: In practice, bottleneck on dataloading usually only happens on the first few steps and in later steps it will bottleneck on computation. We want to have a variable to control whether prefetch the data to device to "simulate" a step in the middle of training/inference process.

torchbenchmark/models/pytorch_struct/__init__.py
erip: Yes, it's a good question. `BucketIterator` was deprecated and removed from torchtext which left a gap of some of the features it supported. One of those features, used here (best I can tell), is max-token batching. This constructs dynamically-sized batches such that the sum of post-pad example lengths don't surpass 200. I took what I believe to be a close approximation (batch by static num of sentences). I can run the profiler in the morning.

PR #651, opened 2021-12-15
Fix size of equation_of_state and isoneutral_mixing, add tke

xuzhao9:
Use the datapoint size=1048576 in https://github.com/dionhaefner/pyhpc-benchmarks/blob/master/README.md

Profiling results are below.

pyhpc_equation_of_state eval:
image

pyhpc_isoneutral_mixing eval:
image

pyhpc_turbulent_kinetic_energy eval:
image

PR #675, opened 2021-12-25
Don't use jit when jit=False for pyhpc benchmarks

jansel:
#651 caused some pretty dramatic changes to my benchmarking results.  It turns out it enabled jit for "eager" mode.  This restores the prior behavior.


xuzhao9 (approved):
LGTM. Sorry for introducing this bug. I didn't check the code when updating the newer version in https://github.com/dionhaefner/pyhpc-benchmarks.

PR #676, opened 2021-12-26
Remove printouts from running benchmarks

jansel:


wconstab (approved):
nice! thanks for doing this

PR #677, opened 2021-12-26
Don't change behavior based on apex package in yolov3

jansel:


wconstab (commented):
I guess this is ok, i'm in favor of consistency, although i'm not sure if it's fair to say apex/amp stuff is banned from our benchmarks.  Are you trying to fix cpu cases?  or gpu cases where AMP isn't available?

jansel:
@wconstab apex/amp is fine with me.  I just don't want an unrelated `pip install ...` to change what the benchmark is measuring.  That is confusing.

I see two options:
1) what this PR does (always disabled)
2) install apex from install.py (it doesn't current) and change this to an error if it is missing

wconstab:
I'd favor #2 in following the spirit of 'benchmark the real user code' which in this case probably wants to use AMP, but if installing it in install.py turns out to cause other issues I'd be ok with #1.

wconstab (approved):

PR #680, opened 2021-12-28
Wconstab/ltc fix attention

wconstab:


torchbenchmark/models/Super_SloMo/requirements.txt
```diff
@@ -1,3 +1 @@
-torch
```
wconstab, line 1: whoops i'll separate these torch/torchvision changes out. it was an accidental part of the commit, i think no longer needed anyway.  (@xuzhao9 did some work on the main installer to help with the issue i was having here)

jansel:
IMO `.item()` is something people use in real models, and we should try to properly support it with Lazy Tensors.   Perhaps we could put Lazy Tensor specific optimizations in a branch, but keep the trunk models as originally written.

jansel:
Closing this as it looks stale.  Feel free to reopen.

PR #686, opened 2022-01-07
Switch to python 3.8 as default in CI and readme.

xuzhao9:
After fixing a few compatibility issues, torchbenchmark can now run on both Python 3.7 and 3.8. We should switch to version 3.8 as default in the CircleCI and README.md.
This PR also moves to CUDA 11.3 as this is the CUDA version used in the nightly CI environment.

### DIFF
--- a/components/_impl/tasks/base.py
+++ b/components/_impl/tasks/base.py
@@ -1,7 +1,6 @@
 """Add Task abstraction to reduce the friction of controlling a remote worker."""
 import abc
 import ast
-import sys
 import functools
 import inspect
 import marshal
@@ -120,28 +119,16 @@ def parse_f(f: typing.Callable) -> typing.Tuple[inspect.Signature, str]:
     src_lines = f_src.splitlines(keepends=False)
 
     node: ast.AST
-    for idx, node in enumerate(f_ast.body[0].body):
+    for node in f_ast.body[0].body:
         # In Python 3.7, there is a bug in `ast` that causes it to incorrectly
         # report the start line of bare multi-line strings:
         #   https://bugs.python.org/issue16806
         # Given that the only use for such strings is a docstring (or multi
         # line comment), we simply elect to skip over them and index on the
         # first node that will give valid indices.
-        # To support both Python 3.7 and 3.8,
-        # We also need to drop bare multi-line strings in 3.8
-        if sys.version_info[:2] == (3, 7):
-            if node.col_offset == -1:
-                assert isinstance(node.value, ast.Str), f"Expected `ast.Str`, got {type(node)}. ({node}) {node.lineno}" 
-                continue
-        else:
-            if idx == 0 and isinstance(node, ast.Expr):
-                text: Option[str] = None
-                if isinstance(node.value, ast.Str):
-                    text = node.value.s
-                elif isinstance(node.value, ast.Constant) and isinstance(node.value.value, str):
-                    text = node.value
-                if text and "\n" in text and ast.get_docstring(f_ast.body[0], clean=False) == text:
-                    continue
+        if node.col_offset == -1:
+            assert isinstance(node.value, ast.Str), f"Expected `ast.Str`, got {type(node)}. ({node}) {node.lineno}"
+            continue
 
         raw_body_lines = src_lines[node.lineno - 1:]
         col_offset = node.col_offset
--- a/components/_impl/tasks/base.py
+++ b/components/_impl/tasks/base.py
@@ -1,6 +1,7 @@
 """Add Task abstraction to reduce the friction of controlling a remote worker."""
 import abc
 import ast
+import sys
 import functools
 import inspect
 import marshal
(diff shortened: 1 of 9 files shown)


components/test/test_subprocess.py
```diff
@@ -156,8 +156,6 @@ def f(
                 raise ValueError
 
             q = 1
-            # Trailing comment isn't part of the execution, so it is actually
```
xuzhao9, line 159: Need to remove the trailing comments because inspect in Python 3.8 fixed the bug of dropping trailing comments.

components/_impl/tasks/base.py
```diff
@@ -119,16 +120,28 @@ def foo() -> None:
...
         # In Python 3.7, there is a bug in `ast` that causes it to incorrectly
         # report the start line of bare multi-line strings:
         #   https://bugs.python.org/issue16806
         # Given that the only use for such strings is a docstring (or multi
         # line comment), we simply elect to skip over them and index on the
         # first node that will give valid indices.
-        if node.col_offset == -1:
-            assert isinstance(node.value, ast.Str), f"Expected `ast.Str`, got {type(node)}. ({node}) {node.lineno}"
-            continue
+        # To support both Python 3.7 and 3.8,
```
robieta, line 130: I don't think we need to change this. Stripping docstrings won't affect the execution in 3.8+.

components/test/test_subprocess.py
robieta: I think this is fine. (It's a very niche case and doesn't affect execution.)

robieta (changes requested):
I think the process isolation code is already good for 3.8

components/test/test_subprocess.py
xuzhao9: Thanks for clarification!

components/_impl/tasks/base.py
xuzhao9: Just to confirm, do you mean we don't strip docstrings for both 3.7 and 3.8? Sounds good to me, and I will go ahead and change the test result in https://github.com/pytorch/benchmark/blob/main/components/test/test_subprocess.py#L88.

robieta: Oh, I see what you mean. I baked the 3.7 behavior into the test. Yeah, you can just remove the docstring check from the test.

xuzhao9: I kinda feel bad if we don't cover docstring in our test case, so I use different test results for different python versions. Please help review, thanks!

robieta: We can also use an f-string predicated on Python version. The only down side is that we don't get the auto-update from expecttest.

robieta: I agree, it is kind of sad to lose test coverage.

xuzhao9: Thanks for the suggestion, improved test using f-string.

xuzhao9:
Can I get another review, @robieta? Thanks!

components/test/test_subprocess.py
```diff
@@ -85,8 +85,14 @@ def f(
             return y
 
         _, body = task_base.parse_f(f)
+        # Python 3.7 removes docstring but 3.8+ doesn't. See `parse_f` for details.
+        docstring = """\
```
robieta, line 89: Nit: You can avoid some escaping by using single quotes. (e.g. `''' """Stuff""" '''`)

components/test/test_subprocess.py
```diff
@@ -85,8 +85,14 @@ def f(
...
+            \"\"\"Docstring
+
+            Note: This will be dropped in Python 3.7. See `parse_f` for details.
+            \"\"\"\n\n""" if not sys.version_info[:2] == (3,7) else ""
         self.assertExpectedInline(
-            self._indent(body), """\
+            self._indent(body), f"""{docstring}\
```
robieta, line 95: Nit: `if sys.version_info[:2] > (3,7)`

components/test/test_subprocess.py
```diff
@@ -121,8 +127,14 @@ def g(x: int) -> int:
         )
 
         _, body = task_base.parse_f(MyClass.g)
+        # Python 3.7 removes docstring but 3.8+ doesn't. See `parse_f` for details.
+        docstring = """\
```
robieta, line 131: Same as above

components/test/test_subprocess.py
```diff
@@ -156,12 +168,16 @@ def f(
                 raise ValueError
 
             q = 1
-            # Trailing comment isn't part of the execution, so it is actually
-            # dropped by `inspect`. (Surprisingly.)
 
         _, body = task_base.parse_f(f)
+        # Python 3.7 removes docstring but 3.8+ doesn't. See `parse_f` for details.
+        docstring = """\
```
robieta, line 174: Same as above

robieta (approved):
LGTM. (Modulo a couple minor nits that you're free to accept or reject.)

PR #692, opened 2022-01-12
Fix assert on tuple

kit1980:
Because of the parentheses assert was running on tuple, and a non-empty tuple is always True, so the assert would not catch empty `json_files`.

An example:

    >>> assert(False, f"Assertion error!") # Assertion error will not be thrown!
    <stdin>:1: SyntaxWarning: assertion is always true, perhaps remove parentheses?


xuzhao9 (approved):
LGTM. Thanks for fixing this bug!

PR #694, opened 2022-01-14
Install the dependencies the first time run the benchmark.

xuzhao9:
Strictly speaking, the bisection workflow shouldn't depend on PyTorch nightly: instead, it should install the TorchBench deps after the first time it compiles PyTorch and its domain packages.

This PR does this optimization and removes PyTorch nightly dependencies for bisection jobs. So the bisection jobs and OnDemand CI jobs can proceed even if the nightly packages are unavailable.

The OnDemand PR CI link is here: https://github.com/pytorch/pytorch/pull/71323. It shows we can run PR CI even without installing nightly packages.


wconstab (approved):
lgtm. thanks!

PR #702, opened 2022-01-19
Cleanup devbig related code.

xuzhao9:
The devbig server is now retired and we can remove related code from the repo.


erichan1 (approved):
LGTM! Nice work cleaning up the codebase.

PR #706, opened 2022-01-21
Enable CUDA graph train on torchvision models

xuzhao9:
Adds CUDA Graph for torchvision models. Since we have used batch sizes that either matches original paper (train), or saturate GPU (device), we won't see much improvement for default batch sizes. However, the performance gain occurs mostly when the batch size is small.

Without CUDA Graph (with batch size 32):
```
$ python run.py resnet50 -t train -d cuda
Running train method from resnet50 on cuda in eager mode.
GPU Time:            1034.030 milliseconds
CPU Dispatch Time:   1026.865 milliseconds
CPU Total Wall Time: 1034.011 milliseconds
```

 With CUDA Graph (with batch size 32):
```
$ python run.py resnet50 -t train -d cuda --train_cudagraph
Running train method from resnet50 on cuda in eager mode.
GPU Time:            1038.927 milliseconds
CPU Dispatch Time:   346.313 milliseconds
CPU Total Wall Time: 1038.941 milliseconds
```

# Latency by batch size (Train, fp32, on V100)

<google-sheets-html-origin>

Batch Size | Eager (ms) | CUDA Graph (ms) | Speedup
-- | -- | -- | --
1 | 89.033 | 50.233 | 43.58%
2 | 92.854 | 56.13 | 39.55%
4 | 93.676 | 71.465 | 23.71%
8 | 105.099 | 105.381 | -0.27%
16 | 167.292 | 167.966 | -0.40%
32 | 297.315 | 297.989 | -0.23%
64 | 561.262 | 562.029 | -0.14%

### DIFF
--- a/torchbenchmark/util/framework/vision/args.py
+++ b/torchbenchmark/util/framework/vision/args.py
@@ -38,7 +38,7 @@ def apply_args(model: BenchmarkModel, args: argparse.Namespace):
     if args.train_cudagraph:
         enable_cudagraph(model, model.example_inputs)
 
-def enable_cudagraph(model: BenchmarkModel, example_input: Tuple[torch.tensor]):
+def enable_cudagraph(model: BenchmarkModel, example_inputs: Tuple[torch.tensor]):
     # setup input and output
     optimizer = optim.Adam(model.model.parameters())
     loss_fn = torch.nn.CrossEntropyLoss()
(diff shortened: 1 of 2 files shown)


torchbenchmark/util/framework/vision/args.py
```diff
@@ -9,6 +9,7 @@ def parse_args(model: BenchmarkModel, extra_args: List[str]) -> argparse.Namespa
     parser.add_argument("--eval-fp16", action='store_false', help="enable eval fp16")
     parser.add_argument("--fx2trt", action='store_true', help="enable fx2trt")
     parser.add_argument("--flops", action='store_true', help="enable flops counting")
+    parser.add_argument("--train_cudagraph", action='store_true', help="enable CUDA Graph for train")
```
eircfb, line 12: Naming nit: why not simply 'cudagraph'? It can also be used for inference.

torchbenchmark/util/framework/vision/model_factory.py
```diff
@@ -52,14 +57,19 @@ def get_module(self):
...
-        loss = torch.nn.CrossEntropyLoss()
+        real_input = [ torch.rand_like(self.example_inputs[0]) ]
+        real_output = [ torch.rand_like(self.example_outputs) ]
         for _ in range(niter):
-            optimizer.zero_grad()
-            pred = self.model(*self.example_inputs)
-            y = torch.empty(pred.shape[0], dtype=torch.long, device=self.device).random_(pred.shape[1])
-            loss(pred, y).backward()
-            optimizer.step()
+            self.optimizer.zero_grad()
```
eircfb, line 63: Is good to move optimizer step into the loop! Should have been like that all along.

ngimel (approved):
This is good for now, later if we use amp with scaler we shouldn't capture optimizer step.

torchbenchmark/util/framework/vision/args.py
xuzhao9: Unfortunately for now we don't have know if the user is attempting to use inference at the point of initialization... I have been hoping to change the API to mark if it is a train or eval test, but haven't got time to do so.

PR #716, opened 2022-01-26
Remove the unused waveglow requirements file

xuzhao9:
These files are not used, but they causes dependbot to complain about using deprecated numpy versions (such as #714  and #713 )


aaronenyeshi (approved):
LGTM, also I don't think our implementation of tacotron2 uses tacotron2/waveglow at all.

PR #718, opened 2022-01-26
Use py38 and CUDA 11.3 as default in the nightly wheel script

xuzhao9:
Use `py38` and `cu113` as the default version to generate nightly configs.
This PR also fixes a bug in the abtest config generation script which uses the `git_version` instead of master commit hash.


xuzhao9:
Can I get a review on this, @erichan1 ?

erichan1 (approved):

PR #731, opened 2022-02-01
changed eval_bs to train_bs

erichan1:
Change to train_bs in training section.


xuzhao9 (approved):
Thanks for fixing this bug!

PR #735, opened 2022-02-04
Add a new "test" argument to models

xuzhao9:
This PR adds a new argument, 'test', which can be 'train' and 'eval', in the model initialization function. It has the following benefits:
- At model initialization, we can choose to only initialize "train" or "eval" model datasets, to avoid wasting device memory
- Separate the space of arguments. It is common that train and eval tests have different experimental features to apply. For example, optimize_for_inference, and tensorrt only work for eval.
- Add an "extra_args" argument to the model, to enable more optional features in the future.


erichan1 (approved):
LGTM! Good work on this migration

PR #736, opened 2022-02-04
Remove the maml models as they have poor quality and should be replaced.

xuzhao9:
This PR removes two models, maml and maml_omiglot. We are working on another round of enhancing models with tensorrt and other features, since these two models are in low quality and bad gpu utilization, and it is not worth to work on these models. So I am removing them for now.


erichan1 (approved):

jansel:
@xuzhao9 can you explain what made these models "low quality" besides GPU utilization?

These are the only example of fine-tuning at inference time in the benchmark suite.

PR #742, opened 2022-02-08
Adds the sweep runner script

xuzhao9:
This PR adds a new `run_sweep.py` script to TorchBench. It runs all specified model tests in subprocess worker, and apply optional arguments to them. Currently, only batch size is supported, we will support `fx2trt` and `torch_tensorrt` in a follow-up PR.


run.py
```diff
@@ -172,26 +172,8 @@ def _validate_devices(devices: str):
...
-        except:
-            print(f"The model {args.model} doesn't support specifying batch size, please remove --bs argument in the commandline.")
-            exit(1)
-    else:
-        if support_extra_args:
-            m = Model(device=args.device, test=args.test, jit=(args.mode == "jit"), extra_args=extra_args)
-        else:
-            m = Model(device=args.device, test=args.test, jit=(args.mode == "jit"))
+
+    m = Model(device=args.device, test=args.test, jit=(args.mode == "jit"), batch_size=args.bs, extra_args=extra_args)
```
xuzhao9, line 176: Now no more hacky condition required.

erichan1 (approved):
LGTM! At first I thought you meant this was sweeping batch size for best batch size, but it seems batch size stays constant and this is running multiple models/tests/devices. Is that correct?

PR #747, opened 2022-02-10
Enabled running torch.LTC benchmarks using XLA devices

miladm:
Enabled running torch.LTC benchmarks using XLA devices.

I've successfully tested the script on `torchvision` Mask RCNN.

### DIFF
--- a/check_lazy.py
+++ b/check_lazy.py
@@ -20,7 +20,6 @@ import importlib
 from torchbenchmark import _list_model_paths
 import lazy_tensor_core
 import datetime
-
 lazy_tensor_core._LAZYC._ltc_init_ts_backend()
 
 # The following models are skipped:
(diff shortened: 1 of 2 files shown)


miladm:
Moved `torch_xla` implementation to a new script. 
@wconstab let me know if this works, or you'd like me to merge it back to `check_lazy.py`

wconstab:
to be honest, we have mostly started using the `lazy_bench.py` script in pytorch repo under lazy_tensor_staging folder.  It has more complexity and focuses on timing rather than just checking simple statistics, but we found over time it's easier to use one script and started adding more statistics/error-message checking into that script.

In short, I am not sure if we care too much about maintaining this script and having a copy is probably fine, up to you.

miladm:
Thanks for the feedback @wconstab!

The idea behind `lazy_bench.py` sounds appealing. Where can I find that script. Didn't find it in this branch at the first glance. Maybe I missed it. lol

wconstab:
@miladm it's here:
https://github.com/pytorch/pytorch/blob/lazy_tensor_staging/lazy_tensor_core/lazy_bench.py

PR #752, opened 2022-02-14
Revert #736 with maml and maml_omiglot

xuzhao9:
This PR is to revert https://github.com/pytorch/benchmark/pull/736 because we prematurely removed `maml` and `maml_omiglot` models without adding proper replacements


jansel (approved):
Thanks!

PR #754, opened 2022-02-15
Enable torch_trt and handle subprocess segfault in subprocess_worker

xuzhao9:
This PR enables [torch_trt](https://github.com/NVIDIA/Torch-TensorRT) module on all the models.
Currently the library will segfault on some models, and the subprocess_worker needs to correctly handle that (otherwise, it will just hang forever because it is blocked by `os.read()` on a pipe whose input process is dead).

We introduce the following mechanism to handle subprocess segfault:
1. The `Pipe` class stores the pid of the child process if the pipe is reading from the child process.
2. When the pipe reads, it always creates a thread that periodically checks the status of the other process at the other end. If the other process dies or is in zombie status, the threads writes a special string, `_DEAD`, into the pipe, together with its exit code.
3. The main thread in the process checks the return message in the pipe, if it finds the `_DEAD` message, throw an exception, which is handled in `subprocess_worker`.

### DIFF
--- a/components/_impl/workers/subprocess_rpc.py
+++ b/components/_impl/workers/subprocess_rpc.py
@@ -194,13 +194,6 @@ class Pipe:
         timeout_callback: typing.Callable[[], typing.NoReturn] = (lambda: None),
     ) -> None:
         self._writer_pid = writer_pid
-        # if _writer_pid is set, check the writer process is the child of current process
-        if self._writer_pid:
-            assert psutil.pid_exists(self._writer_pid), "Writer process must exist. Please report a bug."
-            cur_p = psutil.Process()
-            writer_p = psutil.Process(self._writer_pid)
-            assert writer_p in cur_p.children(), "Writer process must be the child of current process.\
-                                                  Please report a bug."
         self._owns_pipe = read_handle is None and write_handle is None
         if self._owns_pipe:
             self.read_fd, self.write_fd = os.pipe()
--- a/components/_impl/workers/subprocess_rpc.py
+++ b/components/_impl/workers/subprocess_rpc.py
@@ -102,21 +102,19 @@ class _TimeoutPIPE:
     _singleton: typing.Optional["_TimeoutPIPE"] = None
 
     _loop_lock = threading.Lock()
-    _active_reads: typing.Dict[int, typing.Tuple[float, float]]
+    _active_reads: typing.Dict[int, typing.Tuple[float, float, int]]
     _loop_cadence = 1  # second
-    _proc_id: typing.Optional[int] = None
 
     @classmethod
-    def singleton(cls, proc_id) -> "_TimeoutPIPE":
+    def singleton(cls) -> "_TimeoutPIPE":
         # This class will spawn a thread, so we only want one active at a time.
         with cls._singleton_lock:
             if cls._singleton is None:
-                cls._singleton = cls(proc_id)
+                cls._singleton = cls()
             return cls._singleton
 
-    def __init__(self, proc_id: int) -> None:
+    def __init__(self) -> None:
         self._active_reads = {}
-        self._proc_id = proc_id
         self._thread = threading.Thread(target=self._loop)
         self._thread.daemon = True
         self._thread.start()
@@ -129,17 +127,17 @@ class _TimeoutPIPE:
             time.sleep(self._loop_cadence)
             now = time.time()
             with self._loop_lock:
-                for w_fd, (timeout, start_time) in tuple(self._active_reads.items()):
+                for w_fd, (timeout, start_time, writer_pid) in tuple(self._active_reads.items()):
+                    # check if the process is still alive
+                    # it could be in zombie status, or has already been reaped
+                    if not psutil.pid_exists(writer_pid) or \
+                        psutil.Process(writer_pid).status() == psutil.STATUS_ZOMBIE:
+                        os.write(w_fd, _DEAD)
+                        self.pop(w_fd)
+                    # check if process timeout
                     if timeout:
                         if now - start_time >= timeout and w_fd in self._active_reads:
                             os.write(w_fd, _TIMEOUT)
-                            self.pop(w_fd)
-                    else:
-                        # if timeout is None, only check if the process is still alive
-                        # it could be in zombie status, or has already been reaped
-                        if not psutil.pid_exists(self._proc_id) or \
-                            psutil.Process(self._proc_id).status() == psutil.STATUS_ZOMBIE:
-                            os.write(w_fd, _DEAD)
                             self.pop(w_fd)
 
     def pop(self, w_fd: int) -> None:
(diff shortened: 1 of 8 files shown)


components/_impl/workers/subprocess_rpc.py
```diff
@@ -102,17 +104,19 @@ class _TimeoutPIPE:
     _loop_lock = threading.Lock()
     _active_reads: typing.Dict[int, typing.Tuple[float, float]]
     _loop_cadence = 1  # second
+    _proc_id: typing.Optional[int] = None
 
     @classmethod
-    def singleton(cls) -> "_TimeoutPIPE":
+    def singleton(cls, proc_id) -> "_TimeoutPIPE":
```
robieta, line 110: We shouldn't have multiple singletons. (And `cls._singleton = cls(proc_id)` will trample with multiple active workers, which is currently allowed.)

components/_impl/workers/subprocess_rpc.py
```diff
@@ -138,20 +150,15 @@ def pop(self, w_fd: int) -> None:
...
-        # loop thread.
-        if timeout is None:
-            yield
-            return
-
+        # Spawn a loop thread to periodically check the liveness of subprocess
         w_fd = pipe.write_fd
-        assert w_fd is not None, "Cannot timeout without write file descriptor."
-        singleton = cls.singleton()
+        # assert w_fd is not None, "Cannot timeout without write file descriptor."
```
xuzhao9, line 155: TODO: removing this assertion is bad, but still trying to understand why `w_fd` could be `None` here. @robieta any ideas?

components/_impl/workers/subprocess_rpc.py
```diff
@@ -102,17 +104,19 @@ class _TimeoutPIPE:
...
         # This class will spawn a thread, so we only want one active at a time.
         with cls._singleton_lock:
             if cls._singleton is None:
-                cls._singleton = cls()
+                cls._singleton = cls(proc_id)
             return cls._singleton
 
-    def __init__(self) -> None:
+    def __init__(self, proc_id: int) -> None:
         self._active_reads = {}
```
robieta, line 118: If you store the pid as a third element of `_active_reads`, you can associate a pid per outstanding read and not have to specialize the class itself to a particular child.

components/_impl/workers/subprocess_worker.py
```diff
@@ -117,6 +117,11 @@ def __init__(self, timeout: typing.Optional[float] = None) -> None:
             **popen_kwargs,
         )
 
+        # setup the pid of child process for the pipes
```
xuzhao9, line 120: The proc's pid is unknown when we build the pipe, so reset the correct pids here.

components/_impl/workers/subprocess_rpc.py
```diff
@@ -126,9 +130,17 @@ def _loop(self):
             now = time.time()
             with self._loop_lock:
                 for w_fd, (timeout, start_time) in tuple(self._active_reads.items()):
-                    if now - start_time >= timeout and w_fd in self._active_reads:
-                        os.write(w_fd, _TIMEOUT)
-                        self.pop(w_fd)
+                    if timeout:
```
robieta, line 133: I actually think we should unconditionally check zombie status, and then optionally check timeout. (Probably represent no timeout as something like `[None, -1, pid]`?) If we have a long timeout but a process dies immediately, it's a better user experience to fail immediately.

components/_impl/workers/subprocess_rpc.py
```diff
@@ -126,9 +130,17 @@ def _loop(self):
...
+                        if now - start_time >= timeout and w_fd in self._active_reads:
+                            os.write(w_fd, _TIMEOUT)
+                            self.pop(w_fd)
+                    else:
+                        # if timeout is None, only check if the process is still alive
+                        # it could be in zombie status, or has already been reaped
+                        if not psutil.pid_exists(self._proc_id) or \
+                            psutil.Process(self._proc_id).status() == psutil.STATUS_ZOMBIE:
```
robieta, line 141: Random aside: `== psutil.STATUS_ZOMBIE` does such a great job of conveying intent. I love it.

components/_impl/workers/subprocess_rpc.py
```diff
@@ -162,22 +169,25 @@ def maybe_timeout_read(cls, pipe: "Pipe") -> None:
...
+        1) Process liveness checks
+        2) File descriptor lifetimes
+        3) File descriptor inheritance
+        4) Message packing and unpacking
+        5) (Optional) timeouts for reads
     """
 
     def __init__(
         self,
+        proc_id: int = 0,
```
robieta, line 184: nit: Should probably call this `writer_pid` since a pipe connects two processes. Also this should be an `Optional[int] = None` so `get_pid()` can assert that it was set.

robieta (changes requested):
I left some comments. I also don't think it's necessary to have the child check. Spawning an extra thread can add jitter to timing measurements, and the parent should not regularly fail without cleaning up.

components/_impl/workers/subprocess_rpc.py
robieta: It will be None if this pipe can only be used to read. The child only gets the read_fd for the input pipe, and only gets the write_fd for the output and data pipes.

xuzhao9: Thanks! Now the timeout and zombie status check is only for the parent process and `output_pipe`.

xuzhao9:
@robieta I have addressed the comments. I feel it is in a much better shape! Please help re-review.

components/_impl/workers/subprocess_rpc.py
```diff
@@ -138,21 +148,17 @@ def pop(self, w_fd: int) -> None:
...
-        # without calling `cls.singleton()` so we don't spawn an unnecessary
-        # loop thread.
-        if timeout is None:
-            yield
-            return
-
+        # Spawn a loop thread to periodically check the liveness of subprocess
         w_fd = pipe.write_fd
         assert w_fd is not None, "Cannot timeout without write file descriptor."
+        assert pipe.get_writer_pid() is not None, "Cannot check process livenss without pid."
```
robieta, line 154: nit: liveness

components/_impl/workers/subprocess_rpc.py
```diff
@@ -229,6 +250,14 @@ def write(self, msg: bytes) -> None:
 
         os.write(self.write_fd, packed_msg)
 
+    def get_writer_pid(self) -> int:
+        assert self._writer_pid, "Writer pid is not specified. Maybe calling from child process or input pipe.\
```
robieta, line 254: nit: `assert self._writer_pid is not None` is clearer, since zero is falsy.

robieta (approved):
LGTM. (I had a couple of minor comments, but not worth forcing you to go through another round of review.) This is a really good improvement. Thanks!

components/_impl/workers/subprocess_rpc.py
xuzhao9: Aha, thanks for catching that!

PR #760, opened 2022-02-18
Run 10 iterations and get the median latency as result

xuzhao9:
Maybe we should use the maximum number instead of median number?


xuzhao9:
This PR is superseded by https://github.com/pytorch/benchmark/pull/763

PR #761, opened 2022-02-18
Move the jit script code to __post_init__

xuzhao9:
We need to remove current jit code from each model directory and use a unified entry for all the transformations. This is because if we do the jit script first, then change the precision to fp16, the CI test will fail with error: https://app.circleci.com/pipelines/github/pytorch/benchmark/3665/workflows/da928033-03fa-48d0-90a4-788d3ee794ed/jobs/3771

However, I noticed different models are using different `torch.jit` APIs: 1) `torch.jit.script(model)`, 2) `torch.jit.script(model, example_inputs)`, 3) `torch.jit.trace(model, example_inputs)` 4) an extra `torch.jit.optimize_for_inference()` for inference.  Which one should I use if we are sharing the jit scripting code for all the models? @Krovatkin 

The current design is to use `torch.jit.trace(model, example_inputs)` by default. For models that need to call `torch.jit.trace()` (like nvidia_deeprecommender), or models that need to script multiple `torch.nn.Module` instances, they should add a callback function, `jit_callback(self)`, to handle the JIT enablement.


davidberard98:
@xuzhao9 imo it would be best to only move optimize_for_inference to the `__post_init__`, in case the choice of tracing vs scripting is dependent on the model.

I think the issue was that when you do optimize_for_inference().half(), constants in the graph are inlined by ofi() and then half() has no effect on the inlined constants. But if you call half() before optimize_for_inference(), the constants are first converted to half *before* getting inlined, which fixes the issue.

But... if you *were* going to use a single script/trace type for all the models, then AFAIK you should use torch.jit.script(model, example_inputs). Tracing can give wrong results if the model behavior is data dependent. And I think the purpose of example_inputs is to provide type annotations when they aren't given. But it's possible that some of the models aren't scriptable (or you might want to support such models in the future)

xuzhao9:
@davidberard98 Thanks for the comments! In case models are using `trace()` vs. `script()`, we can use `script()` by default, and add a `jit_callback()` function for models that use `trace()`, or need special handling.

davidberard98 (approved):
Sure, that sounds good 👍

Krovatkin (approved):
:shipit:

PR #762, opened 2022-02-19
First PR to add the correctness checking code to eval tests

xuzhao9:
This PR prepares adding the correctness checking code to eval tests:

1. Each `eval()` function now returns `Tuple[torch.Tensor]`, i.e., the inference result
2. Add a test to check 1) is true for every model
3. change `run_sweep.py` to prepare for the correctness checking

A follow-up PR is https://github.com/pytorch/benchmark/pull/763, which adds the actual correctness calculation code.


frank-wei (approved):

frank-wei:
LGTM!

PR #763, opened 2022-02-22
Add the actual TensorRT correctness testing code

xuzhao9:


run.py
```diff
@@ -51,49 +52,55 @@ def run_one_step_with_cudastreams(func, streamcount):
...
-        t0 = time.time_ns()
-        start_event.record()
-        func()
-        t1 = time.time_ns()
-
-        end_event.record()
-        torch.cuda.synchronize()
-        t2 = time.time_ns()
+    result_summary = []
+    for _i in range(num_iter):
```
frank-wei, line 61: @xuzhao9 ,is it better to add the multi run inside of run module. It is different meaning from yours here(your implementation will add more individual tests here). It will help to eliminate the overhead during the first run by averaging it.
   with torch.inference_mode():
        start_event.record()
        for _ in range(iters):
            f(*args)
        end_event.record()
    torch.cuda.synchronize()

xuzhao9: We do observer the overhead of the first run is larger than others. What do you think could be the reason (given that we have done warmup)?

Also, the code is working for both train and inference, so we can't just add one line `with torch.inference_mode()` here. Also, we are measuring the total time including `torch.cuda.synchronize()`, so I don't think multi-run inside the run module is better?

frank-wei: warmup will help as this is NV suggested, otherwise the GPU will be a little bit slow initially. 
The benefit of multi-run before one sync(like I mentioned) is that sync only happen once and kernels are aggregated across runs, the purposes are 1) eliminating some bubbles before 1st kernel and the last kernel from one run's perspective 2) multiple runs have more kernels for GPU to scheduling and hopefully, the bubbles between kernels are reduced. Overall, it is more suitable to test only CUDA side run time.
In our case as you mentioned, we need to consider training and would like to measure the total time. The purpose is different. I am OK with that as long as we are aware of that difference.

frank-wei (approved):

run.py
xuzhao9: We are measuring the total time because we hope to measure both CUDA time, and PyTorch framework overhead time, e.g., we will detect the regression of both sides. I guess the CUDA time can be measured with `start_event.elapsed_time(end_event)`?

PR #765, opened 2022-02-23
Migrate from Ubuntu 16.04 to 20.04

kit1980:


.circleci/config.yml
```diff
@@ -13,9 +13,6 @@ jobs:
       - run:
           name: Install Conda
           command: ./scripts/install_basics.sh
-      - run:
-          name: Install CUDA 11.3
-          command: sudo ./scripts/install_cuda.sh
```
xuzhao9, line 18: Can you also help remove the `scripts/install_cuda.sh`, since it is not used?

xuzhao9 (approved):
LGTM - please also remove `install_cuda.sh`. Also, what cuda and cudnn version does this image use?

.circleci/config.yml
kit1980: removed

kit1980: @xuzhao9 CUDA version on the host is 11.4
CuDNN is not installed on the host, but it's bundled with PyTorch, right?
`torch.backends.cudnn.version()` returns  8302, so CuDNN version 8.3.2.

xuzhao9: @kit1980 Thanks for the information. Yes CUDA 11.4 and CuDNN 8.3.2 looks good (should be compatible with PyTorch cu113 build).
Right, CuDNN is now bundled with PyTorch (https://github.com/pytorch/builder/pull/1005).

PR #766, opened 2022-02-23
Add pygame with gym requirement

kit1980:
Currently the CI fails with "ModuleNotFoundError: No module named 'pygame'".
The reason seems to be a recent update go gym, see https://github.com/openai/gym/issues/2634


kit1980:
The CI for this PR failed after "ModuleNotFoundError: No module named 'pygame'", with "RuntimeError: no valid convolution algorithms available in CuDNN".
So it fixes the pygame issue.

xuzhao9 (approved):
Thanks for fixing this!

PR #769, opened 2022-02-24
Add post train and eval run hook to every model

xuzhao9:
This is needed for adding the dispatch flops counter.

The design is to rename `train()` and `eval()` in each model into `_train()` and `_eval()`, and the actual `train()` and `eval()` function in the base class `BenchmarkModel` will call the inner implementation first, then call the post execution hook method.


xuzhao9:
Maybe we should use another workaround for calculating flops and don't introduce interface change as this PR did.

PR #776, opened 2022-03-01
Enable fp16 amp for all models

xuzhao9:
This PR adds fp16 amp to all models. Basically, it adds an autocast context to all eval tests:
```
with torch.cuda.amp.autocast():
    eval()
```

### DIFF
--- a/run_sweep.py
+++ b/run_sweep.py
@@ -102,7 +102,7 @@ def _run_model_test(model_path: pathlib.Path, test: str, device: str, jit: bool,
         result.results["latency_ms"] = run_one_step(func, device)
         # if the model provides eager eval result, save it for cosine similarity
         correctness = task.get_model_attribute(correctness_name)
-        if correctness:
+        if not correctness == None:
             result.results[correctness_name] = correctness
     except NotImplementedError as e:
         status = "NotImplemented"
--- a/torchbenchmark/models/BERT_pytorch/__init__.py
+++ b/torchbenchmark/models/BERT_pytorch/__init__.py
@@ -159,7 +159,7 @@ class Model(BenchmarkModel):
     def set_module(self, new_model):
         self.model.bert = new_model
 
-    def eval(self, niter=1) -> typing.Tuple[torch.Tensor]:
+    def _eval(self, niter=1) -> typing.Tuple[torch.Tensor]:
         model = self.model
         for _ in range(niter):
             # 1. forward the next_sentence_prediction and masked_lm model
@@ -173,7 +173,7 @@ class Model(BenchmarkModel):
             loss = next_loss + mask_loss
         return (next_sent_output, mask_lm_output)
 
-    def train(self, niter=1):
+    def _train(self, niter=1):
         trainer = self.model
         for _ in range(niter):
             # 1. forward the next_sentence_prediction and masked_lm model
--- a/torchbenchmark/models/speech_transformer/__init__.py
+++ b/torchbenchmark/models/speech_transformer/__init__.py
@@ -59,7 +59,7 @@ class Model(BenchmarkModel):
         elif self.test == "eval":
             self.evalcfg.model = new_model
 
-    def train(self, niter=1):
+    def _train(self, niter=1):
         if self.device == "cpu":
             raise NotImplementedError("CPU is not supported by this model")
         if self.jit:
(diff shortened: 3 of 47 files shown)


run_sweep.py
```diff
@@ -102,7 +102,7 @@ def _run_model_test(model_path: pathlib.Path, test: str, device: str, jit: bool,
         result.results["latency_ms"] = run_one_step(func, device)
         # if the model provides eager eval result, save it for cosine similarity
         correctness = task.get_model_attribute(correctness_name)
-        if correctness:
+        if not correctness == None:
```
xuzhao9, line 105: `correctness` value could be `0.0`

torchbenchmark/models/BERT_pytorch/__init__.py
```diff
@@ -159,7 +159,7 @@ def get_module(self):
     def set_module(self, new_model):
         self.model.bert = new_model
 
-    def eval(self, niter=1) -> typing.Tuple[torch.Tensor]:
+    def _eval(self, niter=1) -> typing.Tuple[torch.Tensor]:
```
xuzhao9, line 162: Rename `eval()` to `_eval()` in order to support nested contexts.

torchbenchmark/models/speech_transformer/__init__.py
```diff
@@ -59,15 +59,15 @@ def set_module(self, new_model):
         elif self.test == "eval":
             self.evalcfg.model = new_model
 
-    def train(self, niter=1):
+    def _train(self, niter=1):
```
dzhulgakov, line 62: nit: maybe there's some way to avoid renaming the method for every single model? You could perhaps do some `invoke_train` wrapper or move the wrapping to a different place

dzhulgakov (commented):
I'm not sure enabling fp16 for every model makes sense. E.g. for pyhpc and similar scientific benchmarks the precision would be terrible. It's mostly applicable to vision models.

Besides for some models, there's already support for amp in the training loop - e.g. for https://github.com/pytorch/benchmark/blob/f56637b2c392873639ee639d88d5df37382204af/torchbenchmark/models/pytorch_unet/__init__.py#L55 (not sure whether we're calling the original training loop or not). I guess it's ok to add a single wrapper, but we should check for those cases that the whole model can be lowered.

Also cc @ngimel

xuzhao9:
> I'm not sure enabling fp16 for every model makes sense. E.g. for pyhpc and similar scientific benchmarks the precision would be terrible. It's mostly applicable to vision models.
> 
> Besides for some models, there's already support for amp in the training loop - e.g. for
> 
> https://github.com/pytorch/benchmark/blob/f56637b2c392873639ee639d88d5df37382204af/torchbenchmark/models/pytorch_unet/__init__.py#L55
> (not sure whether we're calling the original training loop or not). I guess it's ok to add a single wrapper, but we should check for those cases that the whole model can be lowered.
> 
> Also cc @ngimel

Let me clarify: this PR doesn't *enable* fp16 for every model. Instead, it only *adds* fp16 amp support so that every model can enable it if they want, but the default is still fp32. We may later enable fp16 amp based on each model's metadata file, e.g. in https://github.com/pytorch/benchmark/blob/main/torchbenchmark/models/pytorch_unet/metadata.yaml

For the case that some models provide native amp support, we can add an interface, such as `enable_amp()`, as callback. For example:
```
if hasattr(model, 'enable_amp'):
  model.enable_amp()
else:
  # enable with autocast
```

torchbenchmark/models/speech_transformer/__init__.py
xuzhao9: I like the idea! Will make this change.

xuzhao9: Updated: now all tests have the same entry `invoke()`, and we dispatch to either `train()` or `eval()` by the value of `self.test`.

run_sweep.py
xuzhao9: > `if correctness is not None:` to be more pythonic

It looks like `is not` syntax is discouraged in Python 3.8 release note: https://docs.python.org/3.8/whatsnew/3.8.html#changes-in-python-behavior

ngimel: This is for comparisons with strings/ints (that indeed throw warning), `is not None` is still a legit construct, no warning.

xuzhao9: > This is for comparisons with strings/ints (that indeed throw warning), `is not None` is still a legit construct, no warning.

Thanks for pointing out! Mark this as resolved.

xuzhao9:
@dzhulgakov Sorry but can I get another review to make sure all comments are addressed?

ngimel (approved):
LGTM, thanks for adding this

Issue #688, opened 2022-01-08
Gate CUDA failures behind some kind of flag?


Chillee:
Currently, a bunch of models error when you call `get_module()` due to CI reasons. However, if you have an A100 they run perfectly fine. So, it's a bit annoying to need to modify the benchmarks a ton just to get a model that should work on your GPU.

xuzhao9:
Good point, I think we should have an abstract layer to determine what tests (and configurations) are available given the current user platform.

Issue #626, opened 2021-12-09
TODO: Update test_bench.py and test.py to use the same EXCLUDELIST to skip tests


aaronenyeshi:
Create a central EXCLUDELIST that contains all the tests that need to be skipped, and share the same list between test.py and test_bench.py.

At the moment, we only have the list in test.py. This will allow us to have a clear list of skipped tests.

xuzhao9:
Fixed by https://github.com/pytorch/benchmark/pull/781

Issue #598, opened 2021-11-30
Densenet121: Training with batch size 256 will encounter CUDA OOM on CI's Tesla T4


aaronenyeshi:
In the original paper, densenet121 is trained with a batch size of 256 on the ImageNet dataset. They are also training densenet121 on the TitanX GPU with 12GB device memory. However CI's Tesla T4 gets out of memory with 16GB device memory.
Paper: https://arxiv.org/pdf/1608.06993.pdf

In the original repo, DenseNet has an option `-optMemory 4` which significantly reduces the memory footprint and is able to train on TitanX. TorchVision's version of densenet121 has memory_efficient option, that still fails to run on T4 (16GB mem).

Repo: https://github.com/liuzhuang13/DenseNet#memory-efficient-implementation-newly-added-feature-on-june-6-2017
Densenet121 Torchvision: https://github.com/pytorch/vision/blob/d367a01a18a3ae6bee13d8be3b63fd6a581ea46f/torchvision/models/densenet.py#L162

Densenet121 train/example on T4 fails with CUDA OOM:
```
(py37) aaronshi@pytorch-benchmark-dev-0:~/data/as-use/benchmark$ python test.py -k densenet121 --verbose
test_densenet121_check_device_cpu (__main__.TestBenchmark) ... ok
test_densenet121_check_device_cuda (__main__.TestBenchmark) ... ok
test_densenet121_eval_cpu (__main__.TestBenchmark) ... ok
test_densenet121_eval_cuda (__main__.TestBenchmark) ... ok
test_densenet121_example_cpu (__main__.TestBenchmark) ... ok
test_densenet121_example_cuda (__main__.TestBenchmark) ... ERROR
test_densenet121_train_cpu (__main__.TestBenchmark) ... ok
test_densenet121_train_cuda (__main__.TestBenchmark) ... ERROR
```

Torchvision authors may want to implement `-optMemory 4` optimization to allow training on a single device.

aaronenyeshi:
Hi @fmassa, author of memory efficient densenet121 in torchvision. Do you know if anyone would be interested in taking a look at implementing this `-optMemory 4` feature to torchvision's densenet121? 

For reference, Memory Efficient PR: https://github.com/pytorch/vision/pull/1003

xuzhao9:
Fixed by https://github.com/pytorch/benchmark/pull/781

PR #781, opened 2022-03-09
Remove "NotImplemented" flags if the test is limited by hardware capacity

xuzhao9:
When a test is flagged as "NotImplemented", there are actually two cases:
1. The test itself doesn't implement or handle the configs, e.g., unsupervised-learning models like pytorch_struct doesn't have `eval()` tests, or the pyhpc models don't have `train()` tests.
2. The test doesn't support running on our T4 CI GPU machine, but it runs totally fine on other GPUs, such as `V100` or `A100`.

This PR is to eliminate the second case, so that the test can still run through `run.py` or `run_sweep.py` interfaces. Instead, we flag the test to be `not_implemented` in the `metadata.yaml`, and the CI script `test.py` or `test_bench.py` will read from the metadata and determine they are not suitable to run on the CI machine.

This fixes https://github.com/pytorch/benchmark/issues/688, https://github.com/pytorch/benchmark/issues/626, and https://github.com/pytorch/benchmark/issues/598


aaronenyeshi (approved):
LGTM. Thanks, this will keep the model code more clean!

PR #784, opened 2022-03-10
Use fuser contextmanager

xuzhao9:
This PR uses `fuser` context manager to manage run contexts to replace the old "hacky" implementation.
Because the instance generated by `@contextlib.contextmanager` can only be applied once, we need to pass in a lambda and instantiate a new instance every time we run the benchmark


davidberard98 (approved):
Looks good!

Issue #498, opened 2021-10-13
Undocumented LLVM dependency


jamesr66a:
When I try to `install.py`, the build for `tacotron2` fails because it can't find `llvm-config`:

```
Error for /data/users/jamesreed/benchmark/torchbenchmark/models/tacotron2:
---------------------------------------------------------------------------
  ERROR: Command errored out with exit status 1:
   command: /home/jamesreed/local/miniconda3/bin/python -u -c 'import io, os, sys, setuptools, tokenize; sys.argv[0] = '"'"'/tmp/pip-install-gz4dfgi5/llvmlite_c211fe64d78d402eb45e8ea8aa1d1d8e/setup.py'"'"'; __file__='"'"'/tmp/pip-install-gz4dfgi5/llvmlite_c211fe64d78d402eb45e8ea8aa1d1d8e/setup.py'"'"';f = getattr(tokenize, '"'"'open'"'"', open)(__file__) if os.path.exists(__file__) else io.StringIO('"'"'from setuptools import setup; setup()'"'"');code = f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' bdist_wheel -d /tmp/pip-wheel-6s9w112l
       cwd: /tmp/pip-install-gz4dfgi5/llvmlite_c211fe64d78d402eb45e8ea8aa1d1d8e/
  Complete output (26 lines):
  running bdist_wheel
  /home/jamesreed/local/miniconda3/bin/python /tmp/pip-install-gz4dfgi5/llvmlite_c211fe64d78d402eb45e8ea8aa1d1d8e/ffi/build.py
  LLVM version... Traceback (most recent call last):
    File "/tmp/pip-install-gz4dfgi5/llvmlite_c211fe64d78d402eb45e8ea8aa1d1d8e/ffi/build.py", line 105, in main_posix
      out = subprocess.check_output([llvm_config, '--version'])
    File "/home/jamesreed/local/miniconda3/lib/python3.9/subprocess.py", line 424, in check_output
      return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
    File "/home/jamesreed/local/miniconda3/lib/python3.9/subprocess.py", line 505, in run
      with Popen(*popenargs, **kwargs) as process:
    File "/home/jamesreed/local/miniconda3/lib/python3.9/subprocess.py", line 951, in __init__
[...]

PR #787, opened 2022-03-11
Remove model version lock and test on Python 3.9

xuzhao9:
To enable TorchBench on Python 3.9, we need to remove the version locks on the dependencies.
After removing dependencies, installing TorchBench on Python 3.9 no longer requires LLVM.
Fixes https://github.com/pytorch/benchmark/issues/498


jamesr66a (approved):
This seems fine to me, but someone more involved with torchbench should probably give a more authoritative review

PR #794, opened 2022-03-14
attention issue fix for fp16 inference

frank-wei:
The default min value will be overflow when model is under fp16. So I set a min val of fp16 for it.


xuzhao9:
@frank-wei Can you please rebase this PR on the latest main and commit again?

frank-wei:
> @frank-wei Can you please rebase this PR on the latest main and commit again, to fix the CI failure? The failed CI is not caused by this PR, but by the broken main branch this PR is based on, and should be fixed on the latest main.

thanks, @xuzhao9 , rebased.

torchbenchmark/models/BERT_pytorch/bert_pytorch/model/attention/single.py
```diff
@@ -16,7 +16,11 @@ def forward(self, query, key, value, dropout: TensorToTensor, mask: Optional[tor
                  / math.sqrt(query.size(-1))
 
         if mask is not None:
-            scores = scores.masked_fill(mask == 0, -1e9)
+            if scores.dtype == torch.float16:
+                min_mask = -65504.0
```
xuzhao9, line 20: Is there any constant defined like 'torch.float16.MIN_VALUE' or 'torch.float.MIN_VALUE' so that we don't need to hardcode the value?

torchbenchmark/models/BERT_pytorch/bert_pytorch/model/attention/single.py
```diff
@@ -16,7 +16,11 @@ def forward(self, query, key, value, dropout: TensorToTensor, mask: Optional[tor
...
+            else:
+                min_mask = -1e9
+            scores = scores.masked_fill(mask == 0, min_mask)
```
xuzhao9, line 23: This line is filling the `score` variable with `min_mask` for values that are `0`. Does this change the semantics of the code? Looks like it is.

frank-wei: Generally, it won't change the model representation as the min_mask is used as mask to the empty token. So using the minimum float is what the paper suggested.

xuzhao9: > Generally, it won't change the model representation as the min_mask is used as mask to the empty token. So using the minimum float is what the paper suggested.

Can you add comments to the code with a link to the paper so that people understand this change?

xuzhao9 (approved):
LGTM!

PR #802, opened 2022-03-16
Enable fp16 half on timm models

xuzhao9:
This PR enables fp16 (half) by default on `timm` models.

Now all torchvision, huggingface, and timm models support fp16 on CUDA inference by default.

@dzhulgakov please help review and confirm using `fp16` by default makes sense for timm models, which all belong to the CV domain.


dzhulgakov (approved):
Awesome! looks good!

PR #806, opened 2022-03-16
Adding the model input iterators.

xuzhao9:
This PR adds a new interface, `gen_inputs(self) -> Tuple[Generator, Optional[int]]`. The first element is an iterator of the input data, and the second element is the size of the iterator. If the iterator is infinite, the second value is set to None.

Currently, only implement this on `timm` and `torchvision` models. Both of them use randomly generated inputs, therefore the second value is set to None in both cases.

### DIFF
--- a/test.py
+++ b/test.py
@@ -55,7 +55,7 @@ def _create_example_model_instance(task: ModelTask, device: str):
             skip = True
     finally:
         if skip:
-            raise NotImplementedError("Model is not implemented on this device")
+            raise NotImplementedError(f"Model is not implemented on the device {device}")
 
 def _load_test(path, device):
 
--- a/test.py
+++ b/test.py
@@ -44,28 +44,30 @@ class TestBenchmark(unittest.TestCase):
             main(["--repeat=1", "--filter=pytorch_struct", "--device=cpu"])
             self.assertGreaterEqual(mock_save.call_count, 1)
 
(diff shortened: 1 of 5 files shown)


shunting314:
The interface I feel more convenient is:
```
get_inputs(self, n_needed: int) -> List[ExampleBatch]
```

When I test the model for n iterations, I'll call get_inputs(self, n) to get n batches and feed each batch to each run.

But I think the interface you provided is also fine, since I can wrap around it.

shunting314:
Do we know what are the list of models that this interface can be implemented?

shunting314 (approved):

xuzhao9:
> Do we know what are the list of models that this interface can be implemented?

I believe they can be implemented for all models in https://github.com/pytorch/benchmark/tree/main/torchbenchmark/models, but some of them reads data from a mini-dataset, so the length of inputs is limited (for example, `n` will be 10 or something since we don't have more batches).
For others (notably torchvision/timm/huggingface), since the input is randomly generated, we can generate infinite number of inputs

shunting314:
Got it. What's the plan to implement this method for other models in torchbench?

xuzhao9:
> Got it. What's the plan to implement this method for other models in torchbench?

It will need some time to implement, I expect to reach 100% coverage by the end of April

PR #816, opened 2022-03-21
Add TorchDynamo support

xuzhao9:
Add torchdynamo support to TorchBench. Use `torchdynamo.list_backends()` to list available backends and apply them.

Test Plan:
```
$ python run.py resnet18 -d cuda -t eval --torchdynamo ofi
Running eval method from resnet18 on cuda in eager mode.
jit error
  File "/home/xzhao9/torchdynamo/torchdynamo/utils.py", line 192, in torchscript
    return torch.jit.trace(model, example_inputs)
  File "/home/xzhao9/data/miniconda3/envs/py38/lib/python3.8/site-packages/torch/jit/_trace.py", line 741, in trace
    return trace_module(
  File "/home/xzhao9/data/miniconda3/envs/py38/lib/python3.8/site-packages/torch/jit/_trace.py", line 956, in trace_module
    example_inputs = make_tuple(example_inputs)
  File "/home/xzhao9/data/miniconda3/envs/py38/lib/python3.8/site-packages/torch/jit/_trace.py", line 549, in make_tuple
    return tuple(example_inputs)
TypeError: 'NoneType' object is not iterable
GPU Time:              6.079 milliseconds
CPU Dispatch Time:     3.310 milliseconds
CPU Total Wall Time:   6.104 milliseconds
Correctness:            0.999988675117493
```

```
$ python run.py resnet18 -d cuda -t eval --torchdynamo eager
Running eval method from resnet18 on cuda in eager mode.
GPU Time:              6.668 milliseconds
CPU Dispatch Time:     6.023 milliseconds
CPU Total Wall Time:   6.693 milliseconds
Correctness:            0.999996900558472
```

### DIFF
--- a/torchbenchmark/util/backends/torchdynamo.py
+++ b/torchbenchmark/util/backends/torchdynamo.py
@@ -4,65 +4,28 @@ Support TorchDynamo(https://github.com/facebookresearch/torchdynamo) backends
 import argparse
 from typing import List
 
+class NullContext:
+    def __enter__(self):
+        pass
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        pass
+
+def null_experiment(args, model_iter_fn, model, example_inputs):
+    """
+    A no-op experiment useful for making sure TorchBenchmark alone works properly.
+    """
+
+    return []
+
 def parse_torchdynamo_args(model: 'torchbenchmark.util.model.BenchmarkModel', extra_args: List[str]) -> argparse.Namespace:
     parser = argparse.ArgumentParser()
     args = parser.parse_args(extra_args)
     group = parser.add_mutually_exclusive_group()
-    group.add_argument(
-        "--coverage", action="store_true", help="(default) " + help(coverage_experiment)
-    )
-    group.add_argument(
-        "--online-autotune", action="store_true", help=help(speedup_experiment)
-    )
-    group.add_argument(
-        "--offline-autotune", action="store_true", help=help(speedup_experiment)
-    )
-    group.add_argument(
-        "--speedup-fixed1",
-        action="store_true",
-        help="speedup using experimental fixed_strategy backend",
-    )
-    group.add_argument(
-        "--speedup-fixed2",
-        action="store_true",
-        help="speedup using experimental fixed_strategy backend",
-    )
-    group.add_argument(
-        "--speedup-ltc",
-        action="store_true",
-        help="speedup using the ltc backend",
-    )
-    group.add_argument(
-        "--speedup-ltc-trivial",
-        action="store_true",
-        help="speedup using the ltc backend without reusing compiled graph",
-    )
-    group.add_argument(
-        "--overhead", action="store_true", help=help(overhead_experiment)
-    )
-    group.add_argument(
-        "--speedup-ts", action="store_true", help=help(speedup_experiment_ts)
-    )
-    group.add_argument(
-        "--speedup-sr", action="store_true", help=help(speedup_experiment_sr)
-    )
-    group.add_argument(
-        "--speedup-onnx", action="store_true", help=help(speedup_experiment_onnx)
-    )
-    group.add_argument(
-        "--speedup-trt", action="store_true", help=help(speedup_experiment_trt)
-    )
-    group.add_argument(
-        "--accuracy-aot-nop",
-        action="store_true",
-        help="Accuracy testing for AOT vs Eager",
-    )
-    group.add_argument(
-        "--speedup-aot-efficient-fusion",
-        action="store_true",
-        help="speedup using experimental fixed_strategy backend",
-    )
     group.add_argument("--nothing", action="store_true", help=help(null_experiment))
+    parser.add_argument(
+        "--nopython", action="store_true", help="Turn graph breaks into errors"
+    )
     group.add_argument(
         "--nops",
         action="store_true",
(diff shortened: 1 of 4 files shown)


torchbenchmark/util/backends/torchdynamo.py
```diff
@@ -0,0 +1,49 @@
...
+    import torchdynamo
+    optimize_ctx = NullContext()
+    if args.nothing:
+        pass
+    elif args.nops:
+        optimize_ctx = torchdynamo.eval_frame._optimize_catch_errors(
+            torchdynamo.testing.debug_insert_nops, nopython=args.nopython
+        )
+    model.add_context(lambda: optimize_ctx)
+    torchdynamo.reset()
```
xuzhao9, line 49: @jansel  Borrowed from https://github.com/facebookresearch/torchdynamo/blob/main/torchbench.py#L840, not really sure what is in this `reset()` call. For example, if we run a test for 10 times, should we run `reset()` once, or 10 times?

jansel: You should just call `reset()` once per benchmark, it clears code caches.

torchbenchmark/util/backends/torchdynamo.py
```diff
@@ -0,0 +1,49 @@
...
+    def __enter__(self):
+        pass
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        pass
+
+def help(fn):
+    return fn.__doc__
+
+def null_experiment(args, model_iter_fn, model, example_inputs):
```
jansel, line 17: I think you can just cut this.

torchbenchmark/util/backends/torchdynamo.py
```diff
@@ -0,0 +1,49 @@
+"""
+Support TorchDynamo(https://github.com/facebookresearch/torchdynamo) backends
+"""
+import argparse
+from typing import List
+
+class NullContext:
```
jansel, line 7: Is this needed?  It is not actually using torchdynamo.

torchbenchmark/util/backends/torchdynamo.py
```diff
@@ -0,0 +1,49 @@
...
+    def __enter__(self):
+        pass
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        pass
+
+def help(fn):
+    return fn.__doc__
```
jansel, line 15: This likely isn't needed.  Just put the docs in `help="blah blah"`

torchbenchmark/util/backends/torchdynamo.py
```diff
@@ -0,0 +1,49 @@
...
+def apply_torchdynamo_args(model: 'torchbenchmark.util.model.BenchmarkModel', args: argparse.Namespace):
+    import torchdynamo
+    optimize_ctx = NullContext()
+    if args.nothing:
+        pass
+    elif args.nops:
+        optimize_ctx = torchdynamo.eval_frame._optimize_catch_errors(
+            torchdynamo.testing.debug_insert_nops, nopython=args.nopython
+        )
+    model.add_context(lambda: optimize_ctx)
```
jansel, line 48: I am curious if you just replace this with:
`lambda: torchdynamo.optimize("ofi")`
do you see a speedup on any model for inference?

I'd expect that one line is all you need.

There are currently 27 choices you could replace "ofi" with, see: https://github.com/facebookresearch/torchdynamo/issues/58

torchbenchmark/util/backends/torchdynamo.py
```diff
@@ -0,0 +1,49 @@
...
+    parser = argparse.ArgumentParser()
+    group = parser.add_mutually_exclusive_group()
+    group.add_argument("--nothing", action="store_true", help=help(null_experiment))
+    group.add_argument(
+        "--nops",
+        action="store_true",
+        help="Test that bytecode rewriting works properly.",
+    )
+    parser.add_argument(
+        "--nopython", action="store_true", help="Turn graph breaks into errors"
```
jansel, line 34: I don't think this makes sense in torchbench, it is more for debugging graph breaks.

torchbenchmark/util/backends/torchdynamo.py
```diff
@@ -0,0 +1,49 @@
...
+
+def null_experiment(args, model_iter_fn, model, example_inputs):
+    """
+    A no-op experiment useful for making sure TorchBenchmark alone works properly.
+    """
+
+    return []
+
+def parse_torchdynamo_args(model: 'torchbenchmark.util.model.BenchmarkModel', dyamo_args: List[str]) -> argparse.Namespace:
+    parser = argparse.ArgumentParser()
```
jansel, line 25: Perhaps simplify this to just have one option:
`--backend="ofi"`
or 
`--torchdynamo="ofi"`

You can test with the "eager" backend.


Which would just call:
```
model.add_context(functools.partial(torchdynamo.optimize, args.backend))
```

jansel (changes requested):

torchbenchmark/util/backends/torchdynamo.py
xuzhao9: Thanks! I have addressed this comment.

torchbenchmark/util/backends/torchdynamo.py
xuzhao9: Makes sense. Removed.

torchbenchmark/util/backends/torchdynamo.py
xuzhao9: Removed.

torchbenchmark/util/backends/torchdynamo.py
xuzhao9: @jansel I see a speedup when using ofi as the backend, but I also see a strange error `jit error` because `example_inputs` is `None`, which it shouldn't be.

jansel: Hrm, I'll look into that error.

jansel: @xuzhao9 can you retry with the latest TorchDynamo.  This patch should fix it: https://github.com/facebookresearch/torchdynamo/commit/86a2bea62cebedb51d5a1f2230e7817ff2f786fb

xuzhao9: @jansel Thanks! I confirm the latest torchdynamo fixes this issue.

jansel (approved):

PR #822, opened 2022-03-24
fix lower_to_trt api

frank-wei:
as titled.

### DIFF
--- a/torchbenchmark/util/backends/fx2trt.py
+++ b/torchbenchmark/util/backends/fx2trt.py
@@ -1,5 +1,6 @@
 import torch
 from typing import Tuple, Optional
+from fx2trt_oss.fx.lower import lower_to_trt
 
 def enable_fx2trt(max_batch_size: int, fp16: bool, model: torch.nn.Module, example_inputs: Tuple[torch.tensor],
                   is_hf_model: bool=False, hf_max_length: Optional[int]=None) -> torch.nn.Module:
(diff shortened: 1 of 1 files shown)


xuzhao9 (approved):
LGTM!

torchbenchmark/util/backends/fx2trt.py
```diff
@@ -1,5 +1,6 @@
 import torch
 from typing import Tuple, Optional
+from fx2trt_oss.fx.lower import lower_to_trt
```
xuzhao9, line 3: Actually, please move this line in `enable_fx2trt()`, since our CI system only checks default settings and don't have fx2trt access

torchbenchmark/util/backends/fx2trt.py
```diff
@@ -21,57 +22,3 @@ def enable_fx2trt(max_batch_size: int, fp16: bool, model: torch.nn.Module, examp
         )
     return lower_to_trt(module=model, input=example_inputs, \
```
xuzhao9, line 23:
```suggestion
    from fx2trt_oss.fx.lower import lower_to_trt
    return lower_to_trt(module=model, input=example_inputs, \
```

xuzhao9 (changes requested):
Sorry please move the `import` statement into the function to fix the CI.

torchbenchmark/util/backends/fx2trt.py
frank-wei: I see. It is updated.

xuzhao9 (approved):

PR #835, opened 2022-03-29
Add FAMBench xlmr model

xuzhao9:
This PR adds the xlmr model from [FAMBench](https://github.com/facebookresearch/FAMBench/blob/main/benchmarks/xlmr/ootb/xlmr.py) to the TorchBench suite.
It uses the same configuration as the original model, but runs only 1 batch.

### DIFF
--- a/torchbenchmark/models/fambench_xlmr/__init__.py
+++ b/torchbenchmark/models/fambench_xlmr/__init__.py
@@ -29,20 +29,27 @@ import torch.nn.functional as F
 class Model(BenchmarkModel):
     task = NLP.LANGUAGE_MODELING
     FAMBENCH_MODEL = True
-    # original parameters: FAMBench/benchmarks/run_xlmr_ootb.sh
-    DEFAULT_TRAIN_BSIZE = 16
-    DEFAULT_EVAL_BSIZE = 16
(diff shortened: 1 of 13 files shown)


xuzhao9:
@erichan1 can I get a review?

torchbenchmark/models/fambench_xlmr/__init__.py
```diff
@@ -0,0 +1,83 @@
...
+    from xlmr_parser import init_argparse
+
+from torchbenchmark.util.model import BenchmarkModel
+from torchbenchmark.tasks import NLP
+import torch.nn.functional as F
+
+class Model(BenchmarkModel):
+    task = NLP.LANGUAGE_MODELING
+    FAMBENCH_MODEL = True
+    # original parameters: FAMBench/benchmarks/run_xlmr_ootb.sh
```
erichan1, line 32: run_xlmr_ootb.sh is a dummy default config that is too small. Use this one. 

./run_xlmr_ootb.sh -c "--inference-only --famconfig=fb-1dev-A --num-batches=100 --batch-size=96 --sequence-length=64 --vocab-size=250000 --half-model --use-gpu --warmup-batches=20"

erichan1 (approved):
LGTM! Just small change to model configs.

Issue #839, opened 2022-03-31
timm_efficientdet get_module() is broken


jansel:
```
$ ./torchbench.py --no-skip -k timm_efficientdet --nothing
loading annotations into memory...
Done (t=0.01s)
creating index...
index created!
loading annotations into memory...
Done (t=0.01s)
creating index...
index created!
Traceback (most recent call last):
  File "./torchbench.py", line 1024, in <module>
    main()
  File "./torchbench.py", line 915, in main
    for device, name, model, example_inputs in iter_models(args):
  File "./torchbench.py", line 115, in iter_models
    yield load_model(device, model_name, args.training, args.check_accuracy)
  File "./torchbench.py", line 144, in load_model
    model, example_inputs = benchmark.get_module()
  File "/home/jansel/torchbenchmark/torchbenchmark/models/timm_efficientdet/__init__.py", line 126, in get_module
    for _, (input, target) in zip(range(self.num_batches), self.loader_eval):
  File "/home/jansel/conda/envs/torchdynamo/lib/python3.8/site-packages/effdet/data/loader.py", line 125, in __iter__
    for next_input, next_target in self.loader:
  File "/home/jansel/pytorch/torch/utils/data/dataloader.py", line 530, in __next__
    data = self._next_data()
  File "/home/jansel/pytorch/torch/utils/data/dataloader.py", line 570, in _next_data
    data = self._dataset_fetcher.fetch(index)  # may raise StopIteration
  File "/home/jansel/pytorch/torch/utils/data/_utils/fetch.py", line 52, in fetch
    return self.collate_fn(data)
  File "/home/jansel/conda/envs/torchdynamo/lib/python3.8/site-packages/effdet/data/loader.py", line 83, in __call__
    cls_targets, box_targets, num_positives = self.anchor_labeler.label_anchors(
  File "/home/jansel/conda/envs/torchdynamo/lib/python3.8/site-packages/effdet/anchors.py", line 357, in label_anchors
    box_targets_out.append(box_targets[count:count + steps].view([feat_size[0], feat_size[1], -1]))
RuntimeError: view size is not compatible with input tensor's size and stride (at least one dimension spans across two contiguous subspaces). Use .reshape(...) instead.
```

xuzhao9:
After https://github.com/pytorch/benchmark/pull/842, I run the same command and get the following error:

```
  File "./torchbench.py", line 1023, in <module>
    main()
  File "./torchbench.py", line 917, in main
    run_one_model(
[...]

PR #841, opened 2022-03-31
Disable timm_efficientdet get_module()

jansel:
See #839

Do we have tests for the get_module() API?

### DIFF
--- a/torchbenchmark/models/timm_efficientdet/__init__.py
+++ b/torchbenchmark/models/timm_efficientdet/__init__.py
@@ -123,6 +123,7 @@ class Model(BenchmarkModel):
         self.num_epochs = 1
 
     def get_module(self):
+        raise NotImplementedError("https://github.com/pytorch/benchmark/issues/839")
         for _, (input, target) in zip(range(self.num_batches), self.loader_eval):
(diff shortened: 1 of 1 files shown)


xuzhao9:
We do check `get_module()` API in https://github.com/pytorch/benchmark/blob/main/test.py#L67, it is a bit strange why the test didn't capture this. I have fixed this issue in https://github.com/pytorch/benchmark/pull/842.

jansel:
Can we switch the CI to use an A100?

xuzhao9:
> Can we switch the CI to use an A100?

There is no stable A100 runner available on Circle CI or even Github self-hosted runners because:
- AWS simply doesn't support any 1xA100 VM instance
- AWS does support p4d.24xlarge, which has 8xA100 GPUs. But it doesn't fit for CI, because AWS almost never guarantee the capacity to be able to allocate a p4d.24xlarge instance anytime on-demand. It is possible to create an instance, keep it always running, but 1) it may not worth to have an always-on 8xA100 instance just for PR CI 2) having only 1 instance will limit the parallelization of CI jobs (we can have at most 1 CI workflow running if we have only 1 p4d.24xlarge instance).

PR #844, opened 2022-03-31
Improve the `run_sweep.py` script to support precision attribute

xuzhao9:
Rename the `fp16` option to `precision`, providing 3 options: `fp32`, `fp16`, `amp`. It also adds the `precision` attribute to the result of `run_sweep.py`.

Renaming the option to `precision` makes it easier to support other precisions in the future, such as `bfloat16`.


frank-wei (approved):
LGTM to me.

PR #845, opened 2022-03-31
Pin cmake version because recent cmake upgrade breaks pytorch

xuzhao9:
See details at: https://github.com/pytorch/pytorch/issues/74985


erichan1 (approved):
LGTM! Good fix.

PR #846, opened 2022-04-01
adding lazy_bench tests

Krovatkin:


wconstab:
I'm happy about the idea of landing this onto main finally, but does it actually work? I think there are still some model-specific changes on wconstab/ltc branch?  It might be worth seeing if the remainder of those diffs are easily landable to main, or if there are any issues that we might need to address differently.  (worst case we can continue to have the ltc branch of benchmark..)

Krovatkin:
>  I think there are still some model-specific changes on wconstab/ltc branch

@wconstab  Oh, I was under assumption you already ported everything. I only tested a few models. Let do a full sweep then.

lazy_bench.py
```diff
@@ -0,0 +1,723 @@
...
+    t1 = time.perf_counter()
+    return results, t1 - t0
+
+def to_device(tensors, device):
+    """Handles moving tensor or tensors (in various containers) to a new device.
+        Used for various purposes (either correctness checking, or even as an impromptu
+        means of synchronization.) Note: this method doesn't apply a cuda sync, do that outside.
+    """
+
+    try:
```
wconstab, line 307: you know, at some point i hope we can delete all this. @xuzhao9 made some changes or at least proposed to make the output of eval() always tensors or tuples of tensors.  If that is done already, we might be able to just delete this.  But i think we should still land it as-is first.  one thing at a time..

wconstab (approved):
lgtm

PR #873, opened 2022-04-19
Fix PyTorch unet train, only use mixed precision in forward pass.

xuzhao9:
We should only enable amp context on the forward pass in training, not backward pass.


erichan1 (approved):
Makes sense! Thanks for the fix.

PR #880, opened 2022-04-26
Do not report CPU dispatching time in run.py

xuzhao9:
We should not report CPU dispatch time in run.py because it's inaccurate for many models, such as timm.
This is because in third-party model libraries, it is common to have cuda sync in the middle of model execution, in which case we can't really calculate CPU dispatch time by substracting `t1-t0`.


erichan1 (approved):

aaronenyeshi (approved):
LGTM!

PR #881, opened 2022-04-27
fix lazy_bench.py precision option

Krovatkin:


lazy_bench.py
```diff
@@ -606,7 +606,7 @@ def save_error(name, test, error, dir):
                         help="which model run in subprocess. This will ignore filter and exclude")
     parser.add_argument("--allclose_atol", type=float, default=1e-4,
                         help="Absolute tolerance to check lazy result again the correct result")
-    parser.add_argument("--fp16", choices=["no", "half", "amp"], default="no", help="enable fp16 modes from: no fp16, half, or amp")
+    parser.add_argument("--precision", choices=["fp32", "fp16", "amp"], default="fp32", help="enable fp16 modes from: fp32, fp16/half, or amp")
```
wconstab, line 609: woah i forgot we had the fp16 option. what was wrong with it though? buggy, or just a little oddly set up ("no")?

wconstab (approved):

lazy_bench.py
Krovatkin: it's just different values that TB uses. Other than that seems to work.

PR #882, opened 2022-04-27
run some lazy test on CI

Krovatkin:
Adding a very simply sanity check to make sure other changes won't break `bench_lazy.py`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -35,6 +35,10 @@ jobs:
           name: Validate pytest-benchmark invocation of training suite
           command: ./scripts/run_bench_and_upload.sh
           no_output_timeout: "60m"
+      - run:
+          name: Sanity check for lazy_bench.py training suite
+          command: python lazy_bench.py --just_run_once  --run_in_subprocess alexnet --test=eval
+          no_output_timeout: "60m"
 
 workflows:
   version: 2


.circleci/config.yml
```diff
@@ -35,6 +35,10 @@ jobs:
           name: Validate pytest-benchmark invocation of training suite
           command: ./scripts/run_bench_and_upload.sh
           no_output_timeout: "60m"
+      - run:
+          name: Sanity check for lazy_bench.py training suite
+          command: python lazy_bench.py --just_run_once  --run_in_subprocess alexnet --test=eval
```
wconstab, line 40: is this mainly to see if lazy bench throws?  e.g. still need another test in order to ensure correctly 'compiling' via looking at 
metrics

Krovatkin: this is literally to make sure lazy_bench.py runs to a completion. this isn't to test that LT works correctly or even runs)

.circleci/config.yml
```diff
@@ -35,6 +35,10 @@ jobs:
...
+          no_output_timeout: "60m"
```
xuzhao9, line 41: How long will this command normally take? Does it require GPU?

xuzhao9 (approved):
LGTM!

.circleci/config.yml
Krovatkin: it runs one inference of AlexNet on GPU which only takes a few seconds. I could switch it to running on CPU if  it's more cost effective for us.

PR #886, opened 2022-04-28
Improve correctness algorithm

xuzhao9:
This PR improves the correctness algorithm by:
1. run 10 iterations and only pass the test when all of them pass
2. add a stableness test, if the model doesn't output stable results, do not run the correctness test
3. use the `torch.allclose()` to perform the correctness test, except for fx2trt+fp16, which keeps using cosine similarity

### DIFF
--- a/run.py
+++ b/run.py
@@ -183,4 +183,4 @@ if __name__ == "__main__":
     else:
         run_one_step(test, model_flops=model_flops)
     if hasattr(m, 'correctness'):
-        print('{:<20} {:>20}'.format("Correctness:", "%.15f" % m.correctness), sep='')
+        print('{:<20} {:>20}'.format("Correctness: ", m.correctness), sep='')
--- a/torchbenchmark/util/env_check.py
+++ b/torchbenchmark/util/env_check.py
@@ -68,7 +68,7 @@ def torch_allclose(eager_output: Tuple['torch.Tensor'], output: Tuple['torch.Ten
     # sanity checks
     assert len(eager_output) == len(output), "Correctness check requires two inputs have the same length"
     for i in range(len(eager_output)):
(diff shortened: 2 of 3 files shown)


jansel (approved):
Thanks for fixing this @xuzhao9!  Glad to see improving our correctness checking algorithms.

I'd like to find a way to check the correctness of the unstable benchmarks too.  Do we know what the root cause of the instability is?  Is there a problem with how we are running the benchmarks?  Or are they inherently unstable by design?

frank-wei (approved):

xuzhao9:
> I'd like to find a way to check the correctness of the unstable benchmarks too. Do we know what the root cause of the instability is? Is there a problem with how we are running the benchmarks? Or are they inherently unstable by design?
Currently we don't know, we have already set the random seed and other stuff to minimize the randomness. One possibility is we could set `cudnn.deterministic=True` to see if helps, but this harms performance; otherwise, we need to see if the model code resets randomness somewhere inside its code.
In either case, I will rerun the workflow after this is landed, and create an issue to track all unstable models.

jansel:
> > I'd like to find a way to check the correctness of the unstable benchmarks too. Do we know what the root cause of the instability is? Is there a problem with how we are running the benchmarks? Or are they inherently unstable by design?
> 
> Currently we don't know, we have already set the random seed and other stuff to minimize the randomness. One possibility is we could set `cudnn.deterministic=True` to see if helps, but this harms performance; otherwise, we need to see if the model code resets randomness somewhere inside its code. In either case, I will rerun the workflow after this is landed, and create an issue to track all unstable models.

Torchbench.py is able to check correctness for all but 1 benchmark, so perhaps there are improvements you could take from that.

Issue #898, opened 2022-05-04
pplbench_beanmachine benchmark doesn't use pytorch


jansel:
I ran it with the pytorch profiler, and it has zero ops in it.

Looking at the source code.  
```
    def get_module(self):
        return self.model, self.example_inputs
```

Leads to
```
        self.model = RobustRegression()
```

Which comes from:
```
from pplbench.models.robust_regression import RobustRegression
```

Looking for the string "torch" in the `pplbench.models.*` folder:
```
$ grep "torch" .../site-packages/pplbench/models/*.py
$
```
Has no matches.

Looking at robust_regression.py, it seems like it primarily uses `xarray` not PyTorch.

I'd suggest deleting this benchmark.  Or finding an alternate version based on PyTorch.

jansel:
Other parts of BeamMachine might use PyTorch, just not `pplbench.models.robust_regression.RobustRegression()`.

Chillee:
I'm pretty sure the benchmark isn't correctly implemented - will ping the relevant folk.

jansel:
@horizon-blue we are using the `get_module()` API here: https://github.com/pytorch/benchmark/blob/97efd8b0dbec4aaae5f95a76dfee97f6859556e6/torchbenchmark/models/pplbench_beanmachine/__init__.py#L76 -- Which doesn't use any of the code you linked.

Perhaps we should pre-compute the `robust_regression` stuff and exclude it from benchmark timing.  There is not much PyTorch can do to speed up code that doesn't use PyTorch.

xuzhao9:
> @horizon-blue we are using the `get_module()` API here:
> 
> https://github.com/pytorch/benchmark/blob/97efd8b0dbec4aaae5f95a76dfee97f6859556e6/torchbenchmark/models/pplbench_beanmachine/__init__.py#L76
> 
> -- Which doesn't use any of the code you linked.
> Perhaps we should pre-compute the `robust_regression` stuff and exclude it from benchmark timing. There is not much PyTorch can do to speed up code that doesn't use PyTorch.

My understanding is: if `self.test` is set to "train", then the model returned by `get_module()` API uses PyTorch. If `self.test` is set to "eval", then no PyTorch code is used.

Therefore, I submitted https://github.com/pytorch/benchmark/pull/909 to disable the eval test, but keeps the train test.

jansel:
I belive the `get_module()` API doesn't use pytorch in either mode.  It always just returns the xarray robust_regression.   

[...]

PR #909, opened 2022-05-11
Disable eval test in pplbench_beanmachine

xuzhao9:
Disable eval test in pplbench_beanmachine, but keeps the train test.
The eval code is shared among various frameworks, so it doesn't use PyTorch.
Fixes https://github.com/pytorch/benchmark/issues/898

### DIFF
--- a/torchbenchmark/models/pplbench_beanmachine/__init__.py
+++ b/torchbenchmark/models/pplbench_beanmachine/__init__.py
@@ -20,6 +20,9 @@ class Pplbench(torch.nn.Module):
 
         # Get data for inference & evaluating model
         if test == "eval":
+            # The evaluation step doesn't use PyTorch code because it needs to handle
+            # various probabilistic programming languages. See GitHub issue #898
+            raise NotImplementedError("PPLBench Beanmachine eval test doesn't use PyTorch and is disabled - see GH issue #898.")
             # Increased number of samples (n) and number of features(k) to increase computation for eval
             # Default values Reference: https://github.com/facebookresearch/pplbench/blob/main/examples/robust_regression.json
             self.train_data, self.test_data = self.model.generate_data(seed=int(time.time()), n=500000, k=500)
--- a/torchbenchmark/models/pplbench_beanmachine/__init__.py
+++ b/torchbenchmark/models/pplbench_beanmachine/__init__.py
@@ -17,9 +17,10 @@ class Pplbench(torch.nn.Module):
 
         # Instantiate model
         self.model = RobustRegression()
+        self.test = test
 
         # Get data for inference & evaluating model
-        if test == "eval":
+        if self.test == "eval":
             # Increased number of samples (n) and number of features(k) to increase computation for eval
             # Default values Reference: https://github.com/facebookresearch/pplbench/blob/main/examples/robust_regression.json
             self.train_data, self.test_data = self.model.generate_data(seed=int(time.time()), n=500000, k=500)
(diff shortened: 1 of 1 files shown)


torchbenchmark/models/pplbench_beanmachine/__init__.py
```diff
@@ -20,6 +20,9 @@ def __init__(self, test):
 
         # Get data for inference & evaluating model
         if test == "eval":
+            # The evaluation step doesn't use PyTorch code because it needs to handle
+            # various probabilistic programming languages. See GitHub issue #898
+            raise NotImplementedError("PPLBench Beanmachine eval test doesn't use PyTorch and is disabled - see GH issue #898.")
```
jansel, line 25: The `get_module()` API just returns self.model which is set on line 19.

That will never use PyTorch.

jansel (changes requested):

torchbenchmark/models/pplbench_beanmachine/__init__.py
jansel: We should refactor the benchmark not to measure the non-PyTorch regression.

xuzhao9: I updated the code, sorry about the bug. I want to clarify that the `get_module()` API returns the model set on line [76](https://github.com/pytorch/benchmark/blob/01d9cd9209eee91ca9f3c9c9eb90b8dfabf8f109/torchbenchmark/models/pplbench_beanmachine/__init__.py#L76), not line 19.

jansel: We are still passing the `robust_regression.RobustRegression` to the model on line 30.

How much of the time of this mode is spent in non-PyTorch code?

horizon-blue: > We are still passing the `robust_regression.RobustRegression` to the model on line 30.

The naming is getting confusing :)... The `robust_regression.RobustRegression` class on line 30 is not the same as xarray data generation class:

In `pplbench.models`, there's a [robust regression class](https://github.com/facebookresearch/pplbench/blob/main/pplbench/models/robust_regression.py) that's used for data generation and evaluation.

In `pplbench.ppls.beanmachine`, there's also a [robust regression class](https://github.com/facebookresearch/pplbench/blob/main/pplbench/ppls/beanmachine/robust_regression.py#L18), which defines the model in Bean Machine (with pytorch distributions). We are passing this class to MCMC inference method so it knows which model to run when we call `infer`.

jansel: @horizon-blue so to confirm, xarray is *not* used and 100% of the time is spent in pytorch?

horizon-blue: *Most* of the operations in  [`forward`](https://github.com/pytorch/benchmark/blob/97efd8b0dbec4aaae5f95a76dfee97f6859556e6/torchbenchmark/models/pplbench_beanmachine/__init__.py#L46-L47) is done in pytorch. xarray is still used to parse and convert the data.

You mentioned in the other thread that TorchDynamo only calls `get_module()`, in which case `forward` wouldn't be invoked anyways, which is probably why you aren't seeing torch operator in your profiler.

:) I begin to feel like it might be easier if we rework this benchmark to use Bean Machine inference library directly instead of via PPL Bench. It doesn't make sense to introduce an extra layer of complexity when we only want to measure the core computation.

jansel: > _Most_ of the operations in [`forward`](https://github.com/pytorch/benchmark/blob/97efd8b0dbec4aaae5f95a76dfee97f6859556e6/torchbenchmark/models/pplbench_beanmachine/__init__.py#L46-L47) is done in pytorch. xarray is still used to parse and convert the data.

Can we quantify how much time is spent in PyTorch, versus other libraries?  Perhaps using the PyTorch profiler.

PR #913, opened 2022-05-12
Fix fp16 precision and cos sim correctness

xuzhao9:
Currently we are using fp32 as the baseline for all correctness tests.
When using `torch.allclose()` to compare the result between fp32 and torchdynamo eager+fp16, the test returns "incorrect"; on the other hand, using cosine similarity, it returns "correct".

Here is an overview of the results:

|               | torch.allclose() | cosine similarity |
|---------------|------------------|-------------------|
| fp32 baseline vs. torchdynamo-eager-fp16 | *incorrect*        | correct           |
| fp16 baseline vs. torchdynamo-eager-fp16 | correct          | correct           |

Therefore, we are using cosine similarity for all fp16 correctness tests.

Note that this is not a problem of torchdynamo - using `torch.allclose()` to compare fp32 baseline and fp16 result without torchdynamo, the result is also `incorrect`.

### DIFF
--- a/torchbenchmark/util/env_check.py
+++ b/torchbenchmark/util/env_check.py
@@ -102,7 +102,9 @@ def cos_similarity(eager_output: Tuple['torch.Tensor'], output: Tuple['torch.Ten
         t1 = eager_output[i]
         t2 = output[i]
         cos = torch.nn.CosineSimilarity(dim=0, eps=1e-4)
-        # need to call float() because fp16 tensor may overflow when calculating cosine similarity
-        result *= cos(t1.flatten().float(), t2.flatten().float())
+        # deal with special case: both t1 and t2 are zeros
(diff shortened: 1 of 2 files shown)


xuzhao9:
@frank-wei Please let me know if my algorithm makes sense (always using fp32 as the ground-truth for correctness, and use `torch.close()` for fp32 tests, and use cosine similarity for fp16 tests correctness check)

frank-wei:
> @frank-wei Please let me know if my algorithm makes sense (always using fp32 as the ground-truth for correctness, and use `torch.close()` for fp32 tests, and use cosine similarity for fp16 tests correctness check)

Hi, @xuzhao9  that is what I used in torchdynamo tests.

frank-wei (approved):

jansel:
IMO we should use fp64 as the ground-truth.  We have seen issues with fp32 variance as well.

PR #914, opened 2022-05-13
Promote microbenchmarks to userbenchmark

xuzhao9:
We would like to "promote" the `torchbenchmark/microbenchmarks` directory to the `userbenchmark` in the root directory, because:
-  more teams will be adding their benchmarks into this directory, and it is not necessarily microbenchmarks (e.g. distributed/multi-gpu)
- we would like to have a unified way to collect all benchmarks available, run them and get their results. It may also involve comparisons with non-pytorch frameworks

### DIFF
--- a/run_benchmark.py
+++ b/run_benchmark.py
@@ -15,10 +15,10 @@ def run():
     args, bm_args = parser.parse_known_args()
 
     try:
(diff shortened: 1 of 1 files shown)


davidberard98 (approved):
LGMT!

Q: is this effectively just renaming the microbenchmarks? Or is there some other effect?

PR #929, opened 2022-05-24
Add support for NVIDIA DCGM based FLOPs/sec calculation

FindHao:
Add choices for `--flops`:

- `--flops model`: this option will use an estimation method to calculate the flops.
- `--flops dcgm`: this option will use NVIDIA DCGM API to collect hardware counters for FP32 computations, and calculate the flops.
## Dependency
[NVIDIA DCGM](https://developer.nvidia.com/dcgm) is required by this function and could be easily installed following the [official installation guide](https://docs.nvidia.com/datacenter/dcgm/latest/dcgm-user-guide/getting-started.html#installation).

`numba` is the required dependent package which could be installed by `pip install numba`. 

## Run

For example, you can run the following command to get the flops of resnet50.

```
python run.py -d cuda --flops dcgm resnet50
```
The last part of the output is supposed to be like the following. 
```
GPU Time:             12.097 milliseconds
CPU Total Wall Time:  12.137 milliseconds
FLOPS:               1.9684 TFLOPs per second
Correctness:                      Correct
```


components/model_analyzer/TorchBenchAnalyzer.py
```diff
@@ -0,0 +1,120 @@
...
+
+class ModelAnalyzer:
+    def __init__(self):
+        self.gpu_factory = GPUDeviceFactory()
+        self.gpus = self.gpu_factory.verify_requested_gpus(['all', ])
+        # the metrics to be collected
+        # self.gpu_metrics = [GPUUtilization, GPUPowerUsage,
+        #                     GPUFreeMemory, GPUUsedMemory, GPUFP32Active, GPUTensorActive]
+        self.gpu_metrics = [GPUFP32Active]
+        # the final metric results. Its format is {GPU_UUID: {GPUUtilization: }}
```
xuzhao9, line 28: What is `GPUUtilization` here? Can you give a complete example of the final metric result?

components/model_analyzer/TorchBenchAnalyzer.py
```diff
@@ -0,0 +1,120 @@
...
+        records_groupby_gpu = {}
+        records_groupby_gpu = gpu_record_aggregator.groupby(
+            self.gpu_metrics, lambda record: record.device_uuid())
+        
+        for gpu in self.gpus:
+            self.gpu_metric_value[gpu.device_uuid()] = {}
+        for metric_type, metric in records_groupby_gpu.items():
+            for gpu_uuid, metric_value in metric.items():
+                self.gpu_metric_value[gpu_uuid][metric_type] = metric_value
+        # return self.gpu_metric_value
```
xuzhao9, line 69: Maybe this line is not needed?

xuzhao9, line 69: Maybe this line is not needed?

components/model_analyzer/TorchBenchAnalyzer.py
```diff
@@ -0,0 +1,120 @@
...
+    
+    def set_monitoring_interval(self, attempted_interval):
+        """
+        The default monitoring internval is 10ms.
+        """
+        if attempted_interval < 0.001:
+            logger.warning("The attempted interval is too short, would cause untruested profiling results.")
```
xuzhao9, line 76: should be "untrusted"?

xuzhao9, line 76: should be "untrusted"?

components/model_analyzer/TorchBenchAnalyzer.py
```diff
@@ -0,0 +1,120 @@
...
+
+    def print_flops(self):
+        print("==========Summary==========")
+        for gpu_uuid in self.gpu_metric_value:
+            gpu = self.gpu_factory.get_device_by_uuid(gpu_uuid)
+            print(self.gpu_metric_value[gpu_uuid][GPUFP32Active].value())
+            # TFLOPs/second = Device_SM_Count x Device_FMAs_Per_Cycle_Per_SM x 2 x Running_Frequency_KHz x DCGM_Activity / 1e+9
+            print("GPU : TFLOPs/Second %.4f" % (gpu._sm_count * gpu._fma_count * 2 *
+                gpu._frequency * self.gpu_metric_value[gpu_uuid][GPUFP32Active].value() / 1e+9))
+        # @Yueming Hao: print all collected gpu records, for debug only
```
xuzhao9, line 87: It will be better to print full values when user set this to debug mode, say, `print_flops(self, debug=False)` and print full information when `debug=True`.

xuzhao9, line 87: It will be better to print full values when user set this to debug mode, say, `print_flops(self, debug=False)` and print full information when `debug=True`.

components/model_analyzer/TorchBenchAnalyzer.py
```diff
@@ -0,0 +1,120 @@
...
+            else:
+                raise TorchBenchAnalyzerException("No available GPU with uuid ", gpu_uuid, " found!")
+        else:
+            if len(self.gpu_metric_value) > 1:
+                logger.warning("There are multiple available GPUs and will only return the first one's flops.")
+            gpu_uuid = next(iter(self.gpu_metric_value))
+            gpu = self.gpu_factory.get_device_by_uuid(gpu_uuid)
+            return gpu._sm_count * gpu._fma_count * 2 * gpu._frequency * self.gpu_metric_value[gpu_uuid][GPUFP32Active].value() / 1e+9
+
+def test_dcgm_installation():
```
xuzhao9, line 111: The prefix `test_` is for unit tests, which this is a runtime check, so better rename it to `check_dcgm()`.

xuzhao9, line 111: The prefix `test_` is for unit tests, which this is a runtime check, so better rename it to `check_dcgm()`.

components/model_analyzer/TorchBenchAnalyzer.py
```diff
@@ -0,0 +1,120 @@
+
+from components.model_analyzer.dcgm.dcgm_monitor import DCGMMonitor
+from components.model_analyzer.tb_dcgm_types.da_exceptions import TorchBenchAnalyzerException
```
xuzhao9, line 3: Why this directory is called `tb_dcgm_types`? If they are from Nvidia, let's keep its original name

xuzhao9, line 3: Why this directory is called `tb_dcgm_types`? If they are from Nvidia, let's keep its original name

run.py
```diff
@@ -92,10 +107,14 @@ def run_one_step(func, nwarmup=WARMUP_ROUNDS, model_flops=None, num_iter=10):
 
     # if model_flops is not None, output the TFLOPs per sec
     if model_flops:
-        flops, batch_size = model_flops
-        tflops = flops * batch_size / (cpu_walltime / 1.0e9) / 1.0e12
+        if dcgm_enabled:
+            tflops = np.median(flops_records)
```
xuzhao9, line 111: Are we still using median flops here, or are we measuring all 10 iterations and using the average?

xuzhao9, line 111: Are we still using median flops here, or are we measuring all 10 iterations and using the average?

run.py
```diff
@@ -164,18 +183,25 @@ def _validate_devices(devices: str):
...
     test = m.invoke
     model_flops = None
+
     if args.flops:
-        assert hasattr(m, "get_flops"), f"The model {args.model} does not support calculating flops."
-        model_flops = m.get_flops()
+        if args.flops == 'model':
+            assert hasattr(m, "get_flops"), f"The model {args.model} does not support calculating flops."
+            model_flops = m.get_flops()
+        else:
```
xuzhao9, line 200:
```suggestion
        elif args.flops == 'dcgm':
        # ...
        else:
          print(f"unsupported flops counter: {args.flops}")
          sys.exit(1)        
```

xuzhao9, line 200:
```suggestion
        else:
```
```suggestion
        elif args.flops == 'dcgm':
        # ...
        else:
          print(f"unsupported flops counter: {args.flops}")
          sys.exit(1)        
```

run.py
```diff
@@ -164,18 +183,25 @@ def _validate_devices(devices: str):
...
+            from components.model_analyzer.TorchBenchAnalyzer import test_dcgm_installation
+            if test_dcgm_installation():
```
xuzhao9, line 202: Here, if user specifies `args.flops == 'dcgm'` and `test_dcgm_installation()` returns False, we should print an error exit immediately.

xuzhao9, line 202: Here, if user specifies `args.flops == 'dcgm'` and `test_dcgm_installation()` returns False, we should print an error exit immediately.

xuzhao9 (changes requested):

components/model_analyzer/TorchBenchAnalyzer.py
FindHao: `GPUUtilization` is a class imported `from components.model_analyzer.tb_dcgm_types.gpu_utilization`. One example of this `self.gpu_metrics` is 
```
{'GPU-4177e846-1274-84e3-dcde': 
    {<class 'components.model_analyzer.tb_dcgm_types.gpu_fp32active.GPUFP32Active'>: 
            <components.model_analyzer.tb_dcgm_types.gpu_fp32active.GPUFP32Active object at 0x7f14bbae2280>
     }
}
```
The first key is a string indicating the GPU_UUID. The second key is the class of GPUFP32Active. The inside value is an instance of class GPUFP32Active.

components/model_analyzer/TorchBenchAnalyzer.py
FindHao: The last commented line? OK, I'll remove it.

components/model_analyzer/TorchBenchAnalyzer.py
FindHao: Yes. It is a typo.

components/model_analyzer/TorchBenchAnalyzer.py
FindHao: Good suggestion! will fix it in next commit.

components/model_analyzer/TorchBenchAnalyzer.py
FindHao: will fix it in the next commit.

components/model_analyzer/TorchBenchAnalyzer.py
FindHao: The files in this folder comes from difference folders, and I also add some new files. Maybe I can change it to `record_types`.

run.py
FindHao: Forget to submit the change for this part. will fix it in next submit.

run.py
FindHao: Since `argparse` has done this choice limitation for us, do we still need write the else branch? I have tested for arguments other than the choices we given.

run.py
FindHao: I have print error inside this [`test_dcgm_installation()`](https://github.com/pytorch/benchmark/blob/52c604be3d8d532a78b68038614a142cb45fcf8b/components/model_analyzer/TorchBenchAnalyzer.py#L117). Do you think it should be moved to the lines you marked?

run.py
FindHao: fixed by c8e1094007c80e0516e1172ac93f3b332d5be882

components/model_analyzer/TorchBenchAnalyzer.py
FindHao: fixed by ba025da3dd97bee0118495d82c2cff6de5b6459a

components/model_analyzer/TorchBenchAnalyzer.py
FindHao: fixed by ba025da3dd97bee0118495d82c2cff6de5b6459a

components/model_analyzer/TorchBenchAnalyzer.py
FindHao: fixed by ba025da3dd97bee0118495d82c2cff6de5b6459a

components/model_analyzer/TorchBenchAnalyzer.py
```diff
@@ -0,0 +1,120 @@
...
+            self.gpu_metric_value[gpu.device_uuid()] = {}
+        for metric_type, metric in records_groupby_gpu.items():
+            for gpu_uuid, metric_value in metric.items():
+                self.gpu_metric_value[gpu_uuid][metric_type] = metric_value
+    
+    def set_monitoring_interval(self, attempted_interval):
+        """
+        The default monitoring internval is 10ms.
+        """
+        if attempted_interval < 0.001:
```
adnanaziz, line 74: i thought 100 ms is finest sampling rate? 

from https://docs.nvidia.com/datacenter/dcgm/latest/dcgm-user-guide/feature-overview.html

By default, DCGM provides the metrics at a sample rate of 1Hz (every 1000ms). Users can query the metrics at any configurable frequency (minimum is 100ms) from DCGM

components/model_analyzer/TorchBenchAnalyzer.py
```diff
@@ -0,0 +1,120 @@
...
+    def print_flops(self, debug=True):
+        print("==========Summary==========")
+        for gpu_uuid in self.gpu_metric_value:
+            gpu = self.gpu_factory.get_device_by_uuid(gpu_uuid)
+            print(self.gpu_metric_value[gpu_uuid][GPUFP32Active].value())
+            # TFLOPs/second = Device_SM_Count x Device_FMAs_Per_Cycle_Per_SM x 2 x Running_Frequency_KHz x DCGM_Activity / 1e+9
+            print("GPU : TFLOPs/Second %.4f" % (gpu._sm_count * gpu._fma_count * 2 *
+                gpu._frequency * self.gpu_metric_value[gpu_uuid][GPUFP32Active].value() / 1e+9))
+        # @Yueming Hao: print all collected gpu records, for debug only
+        if debug:
```
adnanaziz, line 87: there should be a more elegant way to log this, e.g., with logger levels (and use a more idiomatic way to print gpu_records, assume it's a hash? json with pretty print option is often a good way. see https://stackoverflow.com/questions/15785719/how-to-print-a-dictionary-line-by-line-in-python

components/model_analyzer/TorchBenchAnalyzer.py
```diff
@@ -0,0 +1,120 @@
...
+            gpu_uuid = next(iter(self.gpu_metric_value))
+            gpu = self.gpu_factory.get_device_by_uuid(gpu_uuid)
+            return gpu._sm_count * gpu._fma_count * 2 * gpu._frequency * self.gpu_metric_value[gpu_uuid][GPUFP32Active].value() / 1e+9
+
+def check_dcgm():
+    try: 
+        temp_model_analyzer = ModelAnalyzer()
+        temp_model_analyzer.start_monitor()
+        temp_model_analyzer.stop_monitor()
+    except Exception as e:
```
adnanaziz, line 116: should try to be more fine grained, don't just catch everything. similarly i would expect to see exception rethrown here?

components/model_analyzer/dcgm/dcgm_value.py
```diff
@@ -0,0 +1,151 @@
...
+DCGM_INT32_NOT_PERMISSIONED = (DCGM_INT32_BLANK + 3)
+DCGM_INT64_NOT_PERMISSIONED = (DCGM_INT64_BLANK + 3)
+DCGM_FP64_NOT_PERMISSIONED = (DCGM_FP64_BLANK + 3.0)
+DCGM_STR_NOT_PERMISSIONED = "<<<NOT_PERM>>>"
+
+
+###############################################################################
+# Functions to check if a value is blank or not
+def DCGM_INT32_IS_BLANK(val):
+    if val >= DCGM_INT32_BLANK:
```
adnanaziz, line 47: highly nonidiomatic - should be return val >= DCGM_INT32_BLANK

components/model_analyzer/dcgm/dcgm_value.py
```diff
@@ -0,0 +1,151 @@
...
+def DCGM_FP64_IS_BLANK(val):
+    if val >= DCGM_FP64_BLANK:
+        return True
+    else:
+        return False
+
+
+# Looks for <<< at first position and >>> inside string
+def DCGM_STR_IS_BLANK(val):
+    if 0 != val.find("<<<"):
```
adnanaziz, line 69: feels awkward, not really sure what it's doing

adnanaziz (changes requested):
nice to see dcgm pulled into perf, thanks for doing this.

i am curious to know what the perf overhead of dcgm is, can we run torchbench with / without it and with different sampling rates?

components/model_analyzer/TorchBenchAnalyzer.py
FindHao: OK. Will fix this in next submit.

components/model_analyzer/TorchBenchAnalyzer.py
FindHao: rethrown? you mean throw exceptions to upper level? https://github.com/pytorch/benchmark/pull/929#discussion_r881056874 @xuzhao9 also mentioned this. So, do you think I should handle those exceptions in the upper level?

components/model_analyzer/dcgm/dcgm_value.py
FindHao: Those flies start with `# Copyright (c) 2020, NVIDIA CORPORATION. All rights reserved.` come from [official NVIDIA DCGM python bindings](https://github.com/NVIDIA/DCGM/tree/master/testing/python3) or [Triton model analyzer](https://github.com/triton-inference-server/model_analyzer/tree/main/model_analyzer). I barely changed those kinds of files, maybe only several lines.

components/model_analyzer/dcgm/dcgm_value.py
FindHao: please check the above comment https://github.com/pytorch/benchmark/pull/929#discussion_r881901268

FindHao:
Reply to this comment https://github.com/pytorch/benchmark/pull/929#pullrequestreview-984113425 , yes, I'll do some experiments after fix all those bugs.

components/model_analyzer/TorchBenchAnalyzer.py
FindHao: fixed by https://github.com/pytorch/benchmark/pull/929/commits/717396e4cb1b418bdf1b6d66e34fd2ba4f42242e

components/model_analyzer/TorchBenchAnalyzer.py
FindHao: fixed by https://github.com/pytorch/benchmark/pull/929/commits/fae1a1d38a54d14442a11f05127540e001fe4911

run.py
FindHao: fixed by https://github.com/pytorch/benchmark/pull/929/commits/fae1a1d38a54d14442a11f05127540e001fe4911

components/model_analyzer/TorchBenchAnalyzer.py
xuzhao9: Looks good to me now

xuzhao9 (approved):
LGTM now, thanks!

PR #947, opened 2022-06-06
Add more GPU metrics to DCGM monitor

FindHao:
## New GPU metrics 
add support for the following GPU metrics, 

- GPUDRAMActive:   The ratio of cycles the device memory interface is active sending or receiving data.
- GPUPCIETX: The number of bytes of active PCIe tx (transmit) data including both header and payload. It is supposed to be device memory write traffic.
- GPUPCIERX: The number of bytes of active PCIe rx (read) data including both header and payload. It is supposed to be device memory read traffic.

## Export all records to csv file ordered by timestamp

Add a new function `export_all_records_to_csv()` which could output all records to a csv file.


xuzhao9:
Can you update the PR title to something like "Add more GPU metrics to DCGM monitor"?

FindHao:
For commit https://github.com/pytorch/benchmark/pull/947/commits/737d19479052694832cfba80b1b16244666412ab

- add an argument `--export-all-records file_path` to `run.py`. 

This option could export current default gpu metrics' DCGM records to a csv file specified by this option.

The final csv file could be be like the following.


timestamp(ms) | gpu_fp32active(%) | gpu_picerx(bytes) | gpu_picetx(bytes) | duration(ms) | read_throughput(GB/s) | write_throughput(GB/s)
-- | -- | -- | -- | -- | -- | --
0 | 0 | 17241379 | 155172413 | 0 |   |  
0.23 | 0 | 3164139 | 9492419 | 0.23 | 12.81 | 38.44
2.6 | 0 | 1131301 | 2036343 | 2.37 | 0.44 | 0.8
3.82 | 0 | 4206098 | 4588471 | 1.22 | 3.22 | 3.51

- `timestamp(ms)` is the timestamp for a record.
- `gpu_fp32active(%)` is the ratio of FP32 unit active cycles during this record
- `gpu_pcierx(bytes)` is how many bytes read from device memory
- `gpu_pcietx(bytes)` is how many bytes write to device memory
-  `duration(ms)` is how long this record monitors
- `read_throughput(GB/s)` is derived by `gpu_pcierx(bytes)` / `duration` *1000/1024/1024/1024
- `write_throughput(GB/s)` is derived by `gpu_pcietx(bytes)` / `duration` *1000/1024/1024/1024

We could easily generate a line chart by open this file with google sheet or excel.

run.py
```diff
@@ -177,7 +182,9 @@ def _validate_devices(devices: str):
     parser.add_argument("--cudastreams", action="store_true",
                         help="Utilization test using increasing number of cuda streams.")
     parser.add_argument("--bs", type=int, help="Specify batch size to the test.")
-    parser.add_argument("--flops", choices=["model", "dcgm"], help="Return the flops result")
+    parser.add_argument("--flops", choices=["model", "dcgm"], help="Return the flops result.")
+    parser.add_argument("--export-all-records", type=str, action="store",
```
xuzhao9, line 186: let's remove the `type=str` part and use `action='store_true'`

xuzhao9 (changes requested):
Overall looks good. Please address the comments and I will review it again. Thanks!

run.py
FindHao: solved by https://github.com/pytorch/benchmark/pull/947/commits/af83645cb679ca3b18ad318c86346a2f882ebe77

xuzhao9 (approved):
LGTM!

FindHao:
messed up commits

PR #957, opened 2022-06-14
Add arg for running dist benchmarks locally

wconstab:


torchbenchmark/util/distributed/submit.py
```diff
@@ -48,6 +48,12 @@ def parse_args(args: List[str]=None):
         help="The Slurm partition to submit to"
     )
 
+    parser.add_argument(
+        "--cluster",
+        default=None,
+        help="Which slurm cluster to target. Use 'local' to run jobs locally, 'debug' to run jobs in process"
```
xuzhao9, line 54: Just want to confirm: if `--cluster` is `None`, slurm will use the partition specified in `--partition`? What does the "debug" option mean for this argument?

xuzhao9 (approved):

torchbenchmark/util/distributed/submit.py
wconstab: debug runs everything in the local process, whereas local runs everything on the local machine in multiple processes.  neither one actually uses the cluster.

If 'None', it uses the default cluster, i guess?  The cluster itself has multiple partitions.  (partitions are probably an invalid concept for the 'local' or 'debug' cluster).

actually, it would be good if you could test the script still works with the default --cluster=None, since i wasn't able to verify this fully with my setup.

xuzhao9: > 

Sorry I am still trying to understand this option. If I am specifying `--cluster debug --ngpus 8`, am I using single gpu in the local process, or all 8 gpus? What about multiple nodes?
I have validated that the default `--cluster` is `None` still works.

xuzhao9:
@wconstab maybe we can close this PR?

wconstab:
Yes, this was landed as part of #959

PR #958, opened 2022-06-14
Add unittest-xml-reporting deps to fix the broken CI

xuzhao9:
The upstream PR https://github.com/pytorch/pytorch/pull/79229 requires CircleCI workflow to install the new dependency `unittest-xml-reporting`


erichan1 (approved):
Good fix!

PR #963, opened 2022-06-15
Fix the te fuser setup

### LINKED ISSUE #956: NVFuser is the default fuser for Nvidia GPU now
Codes for switching fuser need to be updated.

xuzhao9:
Correctly enable `te` fuser environment when fuser is specified as `te`.
Fixes #956

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -19,9 +19,11 @@ def pytest_addoption(parser):
 
 def set_fuser(fuser):
     if fuser == "te":
-        # use default fuser (TE) and
-        # profiling executor settings
-        pass
+        torch._C._jit_set_profiling_executor(True)
+        torch._C._jit_set_profiling_mode(True)
+        torch._C._jit_override_can_fuse_on_cpu(False)
+        torch._C._jit_override_can_fuse_on_gpu(True)
+        torch._C._jit_set_texpr_fuser_enabled(True)
     elif fuser == "old":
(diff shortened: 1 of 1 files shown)


LSC527:
I don't know much about these `torch._C.` variables. One thing to be mention is that set `torch._C._jit_set_nvfuser_enabled(False)` can turn off nvfuser.

conftest.py
```diff
@@ -19,9 +19,11 @@ def pytest_addoption(parser):
 
 def set_fuser(fuser):
     if fuser == "te":
-        # use default fuser (TE) and
-        # profiling executor settings
-        pass
+        torch._C._jit_set_profiling_executor(True)
+        torch._C._jit_set_profiling_mode(True)
+        torch._C._jit_override_can_fuse_on_cpu(False)
```
davidberard98, line 24: this can probably be changed to True to enable nnc on cpu (torch.jit.fuser behavior changed in https://github.com/pytorch/pytorch/pull/74078)

davidberard98 (approved):
LGTM - I would just suggest turning on NNC cpu fusion by default as well.

conftest.py
xuzhao9: Updated the code. I copied the config from timm: https://github.com/rwightman/pytorch-image-models/blob/2f2b22d8c7889174dbf11b92c2d72d8587f9164b/timm/utils/jit.py#L28, so probably good to update there too.

PR #965, opened 2022-06-16
Migrate the workflow to use the new torch-tensorrt package

xuzhao9:
fx2trt has been merged to pytorch/TensorRT, and we are updating the workflow file to use pytorch/TensorRT.
We need to patch pytorch/TensorRT to make sure it builds against PyTorch nightly release.


frank-wei (approved):
LGTM!

PR #986, opened 2022-06-22
Add Dockerfile for building torchbench docker

xuzhao9:
@xwang233 
Build torchbench docker based on pytorch nightly docker.

Environment:
- python 3.8
- cuda 11.3
- pytorch latest nightly

This docker supports nvidia gpu passthrough, but users need to make sure their docker daemon is running with nvidia runtime.
It also bundles torchdynamo and functorch. We will add bundling to Torch-TensorRT and TensorRT later.

The command to build:
```
$ docker pull ghcr.io/pytorch/pytorch-nightly:latest
$ pushd docker
$ docker build . -t pytorch/benchmark:latest
```

The command to run:
```
$ docker run -it  --gpus all pytorch/benchmark:latest /bin/bash
(torchbench) root@5e76c95cb424:/workspace/benchmark#  cd /workspace/benchmark
(torchbench) root@5e76c95cb424:/workspace/benchmark# python run.py resnet18 -d cuda
Running eval method from resnet18 on cuda in eager mode.
GPU Time:              4.578 milliseconds
CPU Total Wall Time:   4.613 milliseconds
```
 
Fixes https://github.com/pytorch/benchmark/issues/984

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -11,11 +11,15 @@ RUN ln -s /opt/conda/etc/profile.d/conda.sh /etc/profile.d/conda.sh && \
 RUN apt-get update
 RUN apt-get install -y git git-lfs jq
 
+RUN git clone https://github.com/pytorch/functorch /workspace/functorch
+RUN git clone https://github.com/pytorch/torchdynamo /workspace/torchdynamo
 RUN git clone https://github.com/pytorch/benchmark /workspace/benchmark
+
 # Clone conda env
 RUN conda create --name torchbench --clone base && \
     echo "conda activate torchbench" >> ${HOME}/.bashrc
 
 # Run the setup
+RUN cd /workspace/functorch; python setup.py develop
+RUN cd /workspace/torchdynamo; python setup.py develop
 RUN cd /workspace/benchmark; python install.py
-
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -20,6 +20,6 @@ RUN conda create --name torchbench --clone base && \
     echo "conda activate torchbench" >> ${HOME}/.bashrc
 
 # Run the setup
+RUN cd /workspace/benchmark; python install.py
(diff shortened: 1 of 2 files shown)


docker/Dockerfile
```diff
@@ -0,0 +1,25 @@
+# base Dockerfile: PyTorch nightly docker
+# ghcr.io/pytorch:pytorch-nightly
+FROM ghcr.io/pytorch/pytorch-nightly:latest
```
xwang233, line 3: We can use build-args like
```dockerfile
ARG FROM_IMAGE=ghcr.io/pytorch/pytorch-nightly:latest
FROM ${FROM_IMAGE}
```
so it's easy to change base image from `docker build` command line.

docker/Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+RUN git clone https://github.com/pytorch/torchdynamo /workspace/torchdynamo
+RUN git clone https://github.com/pytorch/benchmark /workspace/benchmark
+
+# Clone conda env
+RUN conda create --name torchbench --clone base && \
+    echo "conda activate torchbench" >> ${HOME}/.bashrc
+
+# Run the setup
+RUN cd /workspace/benchmark; python install.py
+RUN cd /workspace/functorch; python setup.py develop
```
xwang233, line 24: To keep dependency versions the same, we can add a `pip uninstall functorch -y` here to remove the existing installations.

Same for `torchdynamo`

xwang233 (approved):
Thank you for adding this docker build functionality!

docker/Dockerfile
xuzhao9: Adding this should be fine; just wondering in what case there will be existing installations?

xwang233: Hmm, our current pytorch build installs all those libraries for benchmark. I think it's ok to leave it like this for now. We'll modify our docker build to not install them when we use a commonly shared benchmark infrastructure.

PR #999, opened 2022-06-28
Use sync-free cuda event timing in benchmark

xwang233:
This PR uses sync-free cuda event timing as suggested in https://github.com/pytorch/torchdynamo/issues/425#issuecomment-1163378442

See also https://github.com/pytorch/torchdynamo/issues/425

### DIFF
--- a/run_sweep.py
+++ b/run_sweep.py
@@ -38,7 +38,7 @@ def run_one_step(func, device: str, nwarmup=WARMUP_ROUNDS, num_iter=20) -> Tuple
             t0 = time.time_ns()
             func()
             t1 = time.time_ns()
-        result_summary.append((t1 - t0) / NANOSECONDS_PER_MILLISECONDS)
+            result_summary.append((t1 - t0) / NANOSECONDS_PER_MILLISECONDS)
     
     if device != 'cuda':
         # return wall_latency
--- a/run_sweep.py
+++ b/run_sweep.py
@@ -17,33 +17,49 @@ import torch
 from typing import List, Optional, Dict, Any, Tuple
 from torchbenchmark import ModelTask
 
-WARMUP_ROUNDS = 3
+WARMUP_ROUNDS = 11
 WORKER_TIMEOUT = 600 # seconds
 MODEL_DIR = ['torchbenchmark', 'models']
 NANOSECONDS_PER_MILLISECONDS = 1_000_000.0
 
-def run_one_step(func, device: str, nwarmup=WARMUP_ROUNDS, num_iter=10) -> Tuple[float, Optional[Tuple[torch.Tensor]]]:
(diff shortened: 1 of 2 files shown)


run_sweep.py
```diff
@@ -17,33 +17,49 @@
 from typing import List, Optional, Dict, Any, Tuple
 from torchbenchmark import ModelTask
 
-WARMUP_ROUNDS = 3
+WARMUP_ROUNDS = 11
```
xuzhao9, line 20: Let's make this a tune-able argument of `run_sweep.py`.

run_sweep.py
```diff
@@ -17,33 +17,49 @@
...
 WORKER_TIMEOUT = 600 # seconds
 MODEL_DIR = ['torchbenchmark', 'models']
 NANOSECONDS_PER_MILLISECONDS = 1_000_000.0
 
-def run_one_step(func, device: str, nwarmup=WARMUP_ROUNDS, num_iter=10) -> Tuple[float, Optional[Tuple[torch.Tensor]]]:
+def run_one_step(func, device: str, nwarmup=WARMUP_ROUNDS, num_iter=20) -> Tuple[float, Optional[Tuple[torch.Tensor]]]:
```
xuzhao9, line 25: And this one as well, add it to the command line option.

run_sweep.py
```diff
@@ -17,33 +17,49 @@
...
+        for event in events:
+            times.append(event.elapsed_time(end_event))
+
+        linreg = torch.linalg.lstsq(
+            torch.vstack([torch.arange(num_iter), torch.ones(num_iter)]).T,
+            torch.tensor(times))
+        # print(linreg, flush=True)
+        step_time_ms = -linreg[0][0].item()
+
+        return step_time_ms
```
xuzhao9, line 61: Does this measure the function wall time or the gpu events time? If it is not wall time, we should create another metrics, "gpu_time", in the "results" dict.

torchbenchmark/__init__.py
```diff
@@ -336,7 +336,6 @@ def set_train(self) -> None:
     def invoke(self) -> None:
         self.worker.run("""
             model.invoke()
-            maybe_sync()
```
xuzhao9, line 339: Sorry I don't understand why you are removing this sync function?

xwang233: The idea is to remove cuda sync at every step.

xwang233: cc @ngimel @ptrblck @csarofeen

xuzhao9: Sorry I am asking why remove cuda sync at every step? Currently running CUDA sync at the end of every test in the subprocess is our assumption.

Issue #904, opened 2022-05-06
Add a `--stress` option to keep running the benchmark model


xuzhao9:
Sometimes we want to stress the machine/GPU with a single model. Add a `--stress [time-limit-in-min]` option to keep running the model until the time limit.
We also need to provide QPS/Throughput metrics in this case.

xuzhao9:
Fixed by #1002.

PR #1002, opened 2022-06-28
Add `--stress` option to `run.py`

FindHao:
Add `--stress <seconds>` option to `run.py`. Users can specify a number of minutes to stress devices. 

Related to Issue https://github.com/pytorch/benchmark/issues/904

### DIFF
--- a/run.py
+++ b/run.py
@@ -54,7 +54,7 @@ def run_one_step_with_cudastreams(func, streamcount):
         print('{:<20} {:>20}'.format("GPU Time:", "%.3f milliseconds" % start_event.elapsed_time(end_event)), sep='')
 
 
-def run_one_step(func, nwarmup=WARMUP_ROUNDS, model_flops=None, num_iter=10, model=None, export_dcgm_metrics_file=False):
+def run_one_step(func, nwarmup=WARMUP_ROUNDS, model_flops=None, num_iter=10, model=None, export_dcgm_metrics_file=False, stress=0):
     # Warm-up `nwarmup` rounds
     for _i in range(nwarmup):
         func()
@@ -69,8 +69,14 @@ def run_one_step(func, nwarmup=WARMUP_ROUNDS, model_flops=None, num_iter=10, mod
             model_analyzer.set_export_csv_name(export_dcgm_metrics_file)
             model_analyzer.add_mem_throughput_metrics()
         model_analyzer.start_monitor()
-
-    for _i in range(num_iter):
+    if stress != 0:
+        cur_time = time.time_ns()
+        start_time = cur_time
+        target_time = stress * 60 * 1e9 + start_time
+        num_iter = -1
+    else:
+        _i = 0
+    while ( num_iter != -1 and _i < num_iter ) or ( stress != 0 and cur_time < target_time ) :
         if args.device == "cuda":
             torch.cuda.synchronize()
             start_event = torch.cuda.Event(enable_timing=True)
@@ -98,6 +104,11 @@ def run_one_step(func, nwarmup=WARMUP_ROUNDS, model_flops=None, num_iter=10, mod
             func()
             t1 = time.time_ns()
             result_summary.append([(t1 - t0) / 1_000_000])
+        if stress != 0:
+            cur_time = time.time_ns()
+            print(cur_time)
+        else:
+            _i += 1
     if dcgm_enabled:
             model_analyzer.stop_monitor()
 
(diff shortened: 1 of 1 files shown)


run.py
```diff
@@ -98,6 +104,11 @@ def run_one_step(func, nwarmup=WARMUP_ROUNDS, model_flops=None, num_iter=10, mod
             func()
             t1 = time.time_ns()
             result_summary.append([(t1 - t0) / 1_000_000])
+        if stress != 0:
+            cur_time = time.time_ns()
+            print(cur_time)
```
xuzhao9, line 109: Can we have a more informative output? E.g., `f"Iteration {num_iter}  |  {latency of this iter} ms | {estimated_rest_time}" `?

run.py
```diff
@@ -69,8 +69,14 @@ def run_one_step(func, nwarmup=WARMUP_ROUNDS, model_flops=None, num_iter=10, mod
...
-
-    for _i in range(num_iter):
+    if stress != 0:
+        cur_time = time.time_ns()
+        start_time = cur_time
+        target_time = stress * 60 * 1e9 + start_time
+        num_iter = -1
+    else:
+        _i = 0
+    while ( num_iter != -1 and _i < num_iter ) or ( stress != 0 and cur_time < target_time ) :
```
xuzhao9, line 79: If `stress` is not `0`, what will be the `_i` value here?

xuzhao9 (changes requested):

run.py
FindHao: If the `stress` is not 0, the `_i` will never be touched because `_i` follows `num_iter` and `num_iter == -1`. So we don't need to give it a value.

run.py
FindHao: fixed by https://github.com/pytorch/benchmark/pull/1002/commits/40efc6d4630c3098648f122e8abf03fdac60e111

run.py
FindHao: fixed by https://github.com/pytorch/benchmark/pull/1002/commits/40efc6d4630c3098648f122e8abf03fdac60e111

run.py
xuzhao9: @FindHao sounds good, we can print number of iterations every 10 seconds (actually, let's change the unit of stress test to seconds instead of minutes)

FindHao: ok. 
I'll add the qps thing in next PR to avoid making this PR too messy.

FindHao: fixed in https://github.com/pytorch/benchmark/pull/1002/commits/e4d3fe423d9106cd9144a3700dd76cb698d1321f

run.py
```diff
@@ -69,8 +69,14 @@ def run_one_step(func, nwarmup=WARMUP_ROUNDS, model_flops=None, num_iter=10, mod
             model_analyzer.set_export_csv_name(export_dcgm_metrics_file)
             model_analyzer.add_mem_throughput_metrics()
         model_analyzer.start_monitor()
-
-    for _i in range(num_iter):
+    if stress != 0:
```
xuzhao9, line 72: Let's be Pythonic and use `if stress:`

run.py
xuzhao9: Let's make this predicate clearer:
```suggestion
    while (not stress and _i < num_iter ) or (stress and cur_time < target_time ) :
```

run.py
```diff
@@ -98,6 +104,15 @@ def run_one_step(func, nwarmup=WARMUP_ROUNDS, model_flops=None, num_iter=10, mod
             func()
             t1 = time.time_ns()
             result_summary.append([(t1 - t0) / 1_000_000])
+        if stress != 0:
```
xuzhao9, line 107:
```suggestion
        if stress:
```

run.py
```diff
@@ -98,6 +104,15 @@ def run_one_step(func, nwarmup=WARMUP_ROUNDS, model_flops=None, num_iter=10, mod
...
+            cur_time = time.time_ns()
+            # print out the status every 10s.
+            if (cur_time - last_time) >= 10 * 1e9:
+                est = (target_time - cur_time) / 1e9
+                print('{:<20} {:>20}'.format("Passed Iterations:", "%d" % _i))
```
xuzhao9, line 112: Let's make the two prints on the same line, and organize it like 
```suggestion
                | iter 20 | time per iter | rest time |
```

run.py
FindHao:
```suggestion
    if stress:
```

run.py
FindHao: fixed by https://github.com/pytorch/benchmark/pull/1002/commits/7ebb628fa0baa5333c53df6df5eca7d7b3bf11ac

Issue #895, opened 2022-05-03
Slow CPU performance results in frustrating user experience


jansel:
I've noticed that in the past few months the user experience for using TorchBench for correctness testing [TorchDynamo](https://github.com/facebookresearch/torchdynamo) has gotten much worse.

TorchBench used to run quickly, and I could check all models in minutes.  Now doing the same thing seems to take hours.

Doing some bisecting, it seems what has happened to make it so much slower is increasing batch sizes.  The new batch sizes seem tuned for GPU and reasonable for GPU, but result in a painful experience on CPU.

I'd propose we use different batch sizes on CPU and GPU, especially for inference where batch size is rather arbitrary.  We should target execution times <10 seconds, and ideally <1 second.

We should check if these smaller batch sizes result in different performance characteristics.  For example, turning a compute-bound model into an overhead bound one.  If not, giving TorchBench users faster time-to-signal would be a huge benefit.

Perhaps we could also have a non-default "large batch size" mode, for extra testing.

cc @xuzhao9

xuzhao9:
Correct, I agree that we should use small batch sizes for CPU inference. In addition, we should have different scales of for GPU inference, because some users use small GPU, which default test batch size will be CUDA OOM.

However, for train usually we just use the batch size in the example repository or the original paper, so the batch size can't be arbitrarily adjusted.

jansel:
I'm also noticing very low CPU utilization.  Usually only 1 thread active.  Did something change in how we setup threading?

xuzhao9:
> I'm also noticing very low CPU utilization. Usually only 1 thread active. Did something change in how we setup threading?

Can I get a command to reproduce this problem?

jansel:
```
./torchbench.py --nothing -n100 -k alexnet
```

xuzhao9:
> ```
> ./torchbench.py --nothing -n100 -k alexnet
> ```
For the low cpu utilization, I tried to stress testing alexnet using the newly added `--stress` option (https://github.com/pytorch/benchmark/pull/1002):
`$ python run.py alexnet --stress 200`

From htop, it looks like all threads are being used:
<img width="1883" alt="image" src="https://user-images.githubusercontent.com/502017/176758726-d0c70b02-c9ac-42fb-8180-73211b80a195.png">

so I think it is environment setup issues.

xuzhao9:
Closed by https://github.com/pytorch/benchmark/pull/1003

PR #1003, opened 2022-06-29
Fix slow tests on CPU

xuzhao9:
Partly address https://github.com/pytorch/benchmark/issues/895. Fix the slow cpu models https://github.com/pytorch/torchdynamo/blob/main/benchmarks/torchbench.py#L110. 

For some slow models, like fastNLP_Bert, we are already using batch size 1. In this case:
- if the input data is from real data set (e.g. fastNLP_Bert input is from CMRC2018, vision_maskrcnn is from COCO 2017), we do nothing
- if the input data is randomly generated, we reduce the size of inputs (e.g. hf_* models)

We will continue address this problem in a follow-up PR, where we will reduce CPU batch size to around 1/8 of GPU batch size.


jansel (approved):
Looks good, how fast are these after this change?  Do we still have 1+hour benchmarks?

PR #1012, opened 2022-07-05
Remove optimized_for_inference

xuzhao9:
Now we are enabling ofi for jit inference by default (https://github.com/pytorch/benchmark/blob/main/torchbenchmark/util/backends/jit.py#L4), so remove optimized_for_inference in metadata.


aaronenyeshi (approved):
LGTM!

PR #133, opened 2020-09-25
Include instructions for benchmark initial setup

wconstab:

PR #1021, opened 2022-07-09
fix help message of test_bench.py

seastar105:
README.md was edited by #133 but help message of test_bench.py is still saying `python test.py --setup_only`. 

Fixed it.


xuzhao9 (approved):
Thanks!

PR #1027, opened 2022-07-12
Support probing the best batch size for inference tests

FindHao:
Add proper_bs to find the best batch_size for current devices.

Users can specify `--proper_bs` for `run_sweep.py` to enable this feature. 

The final output file will contains details and optimal batch_size in the `results` section like the following.

```json
        "results": {
            "details": [
                {
                    "batch_size": 1,
                    "latency_ms": 49.924044499999994,
                    "tflops": 0.726485308717901
                },
                {
                    "batch_size": 2,
                    "latency_ms": 26.455478749999997,
                    "tflops": 1.4638535345863946
                },
                {
                    "batch_size": 3,
                    "latency_ms": 19.817013499999998,
                    "tflops": 1.8383686961789516
                }
            ],
            "optimal_latency_bs": 3,
            "optimal_tflops_bs": 3
        }
```

### DIFF
--- a/scripts/proper_bs.py
+++ b/scripts/proper_bs.py
@@ -19,6 +19,7 @@ def run_one_step_flops(func, device: str, nwarmup=WARMUP_ROUNDS, num_iter=10, fl
     if flops:
         model_analyzer = ModelAnalyzer()
         model_analyzer.start_monitor()
+        model_analyzer.set_monitoring_interval(0.01)
     for _i in range(num_iter):
         if device == "cuda":
             torch.cuda.synchronize()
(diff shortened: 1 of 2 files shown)


scripts/proper_bs.py
```diff
@@ -0,0 +1,100 @@
...
+            print(f"Batch Size {batch_size} ", end='')
+            latency_ms_cur = 0
+            if not task.model_details.exists:
+                status = "NotExist"
+                return
+            task.make_model_instance(test=test, device=device, jit=jit, batch_size=batch_size, extra_args=extra_args)
+            result.precision = task.get_model_attribute("dargs", "precision")
+            # Check the batch size in the model matches the specified value
+            if batch_size and (not task.get_model_attribute(bs_name) == batch_size):
+                raise ValueError(f"User specify batch size {batch_size}, but model {result.name} runs with batch size {task.get_model_attribute(bs_name)}. Please report a bug.")
```
xuzhao9, line 68: We need to update this message: some models don't support tuning batch size. For those models, we will set the attribute `ALLOW_CUSTOMIZE_BSIZE = False` (https://github.com/pytorch/benchmark/blob/main/torchbenchmark/models/drq/__init__.py#L90), so you should first check if the model has this attribute and it is set to False. If so, raise a ValueError exception with message "Model does not support tuning batch size", and don't need to run the test.

xuzhao9 (changes requested):
Please improve the exception handling for models don't support variable batch sizes, thanks!

xuzhao9 (changes requested):
Let's make sure we still output just one file for one run. Can we revert this part back?

xuzhao9 (approved):
Thanks!

PR #1032, opened 2022-07-14
Add extra_info to test_bench.py reports

xwang233:
This makes the test_bench.py reports more consistent with those from run_sweep.py


xuzhao9 (approved):
LGTM, thanks!

PR #1045, opened 2022-07-20
fix batch size problem

zzpmiracle:
When bs not specified, this repeat will leads to infinite loop in https://github.com/pytorch/benchmark/blob/main/.github/scripts/run-config.py#L91

### DIFF
--- a/.github/scripts/run-config.py
+++ b/.github/scripts/run-config.py
@@ -69,7 +69,7 @@ def get_devices(config):
 
 def get_batch_sizes(config):
     if not "batch_size" in config:
-        return itertools.repeat(None)
+        return [None]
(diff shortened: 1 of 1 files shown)


xuzhao9:
Can you post the command that causes the infinite loop?

zzpmiracle:
```
import itertools
a = ["a", itertools.repeat(None)]
for b,c in itertools.product(*a): 
   print(b)
```
this demo code could not print anything, but cause MemoryError

xuzhao9:
Ah I see, yes it will cause memory error in this case. Thanks for fixing this bug!

xuzhao9 (approved):

PR #1049, opened 2022-07-20
fix a printout bug; remove default memory monitoring

FindHao:
When adding more metrics rather than FP32Active and PCIERX/TX, the final output head will be duplicated `gpu_fp32active`. It is caused by missing else-branch to catch the other metrics.

Since we get more clear understanding of PCIERX/TX metrics, I think they are not so attractive and it's better to remove them from default configurations.

### DIFF
--- a/run.py
+++ b/run.py
@@ -67,7 +67,6 @@ def run_one_step(func, nwarmup=WARMUP_ROUNDS, model_flops=None, num_iter=10, mod
         model_analyzer = ModelAnalyzer()
         if export_dcgm_metrics_file:
             model_analyzer.set_export_csv_name(export_dcgm_metrics_file)
-            model_analyzer.add_mem_throughput_metrics()
         model_analyzer.start_monitor()
     if stress:
         cur_time = time.time_ns()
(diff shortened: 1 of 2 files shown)


run.py
```diff
@@ -67,7 +67,6 @@ def run_one_step(func, nwarmup=WARMUP_ROUNDS, model_flops=None, num_iter=10, mod
         model_analyzer = ModelAnalyzer()
         if export_dcgm_metrics_file:
             model_analyzer.set_export_csv_name(export_dcgm_metrics_file)
-            model_analyzer.add_mem_throughput_metrics()
```
xuzhao9, line 70: If we are removing this function call, should we remove this function as well?

FindHao: Yes, we can. I'm thinking about adding an API to let users add other metrics more easily without change TorchBenchAnalyzer.py. 
For now, let's remove this function.

FindHao: solved by https://github.com/pytorch/benchmark/pull/1049/commits/7580a44a4cd51edb3623889513328b0047b4a003

xuzhao9 (approved):
LGTM

PR #1050, opened 2022-07-21
Enable 'binned' dynamic shapes for hf models

wconstab:
Looking for feedback overall on dynamic shapes approaches.

See https://github.com/pytorch/torchdynamo/pull/631 for corresponding support in torchdynamo's runner

### DIFF
--- a/torchbenchmark/util/framework/huggingface/model_factory.py
+++ b/torchbenchmark/util/framework/huggingface/model_factory.py
@@ -2,7 +2,6 @@ import math
 import random
 import torch
 from torch import optim
-from torch.jit.frontend import NotSupportedError
 from torchbenchmark.util.model import BenchmarkModel
 import transformers
 from transformers import AutoConfig, ReformerConfig, BertConfig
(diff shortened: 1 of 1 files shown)


torchbenchmark/util/framework/huggingface/model_factory.py
```diff
@@ -59,6 +61,18 @@ def __init__(self, name, test, device, jit=False, batch_size=None, extra_args=[]
             input_ids = torch.randint(0, config.vocab_size, (self.batch_size, self.max_length)).to(device)
             decoder_ids = torch.randint(0, config.vocab_size, (self.batch_size, self.max_length)).to(device)
             self.example_inputs = {'input_ids': input_ids, 'labels': decoder_ids}
+
+            nbuckets = 8
```
xuzhao9, line 65: Can we place this in `get_dynamic_shapes_module()`? My concern is that when we testing non-dynamic shapes with large batch size, it will allocate large chunk of input data which occupies GPU memory.

We would like to only allocate GPU memory when it is needed.

wconstab: yea, that seems fair. I think i wanted to do it in parallel to the original way, and one concern is keeping the tensors the same if called 2x, but i could still do it inside get_dynamic_shapes_module and store/cache it on the self obj when it is created.

torchbenchmark/util/framework/huggingface/model_factory.py
```diff
@@ -1,5 +1,8 @@
+import math
+import random
 import torch
 from torch import optim
+from torch.jit.frontend import NotSupportedError
```
xuzhao9, line 5: Can we import this in `get_dynamic_shaps_module` as well, since it is not used in eager mode?

xuzhao9 (approved):
LGTM, with small comments

PR #1054, opened 2022-07-23
Remove niter for models

xuzhao9:
We should remove niter and use the runner to control number of iterations.
Fixes https://github.com/pytorch/benchmark/issues/1023

### DIFF
--- a/torchbenchmark/models/BERT_pytorch/__init__.py
+++ b/torchbenchmark/models/BERT_pytorch/__init__.py
@@ -161,34 +161,32 @@ class Model(BenchmarkModel):
     def set_module(self, new_model):
         self.model.bert = new_model
 
-    def eval(self, niter=1) -> typing.Tuple[torch.Tensor]:
+    def eval(self) -> typing.Tuple[torch.Tensor]:
         model = self.model
-        for _ in range(niter):
-            # 1. forward the next_sentence_prediction and masked_lm model
-            next_sent_output, mask_lm_output = model.model.forward(*self.example_inputs)
+        # 1. forward the next_sentence_prediction and masked_lm model
+        next_sent_output, mask_lm_output = model.model.forward(*self.example_inputs)
 
-            # 2-1. NLL(negative log likelihood) loss of is_next classification result
-            # 2-2. NLLLoss of predicting masked token word
-            # 2-3. Adding next_loss and mask_loss : 3.4 Pre-training Procedure
-            next_loss = model.criterion(next_sent_output, self.is_next)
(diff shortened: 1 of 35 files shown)


wconstab (approved):
Stamping as the code looks correct to me and I've also wanted to make this change previously.

however, please don't merge until you talk to @jansel and @Krovatkin, anyone else who might be using torchbench and depending on niter arg.  I don't know of anyone actually using it, but it'd be bad to just make it disappear under someone.

jansel:
> Stamping as the code looks correct to me and I've also wanted to make this change previously.
> 
> however, please don't merge until you talk to @jansel and @Krovatkin, anyone else who might be using torchbench and depending on niter arg. I don't know of anyone actually using it, but it'd be bad to just make it disappear under someone.

This is fine.  TorchDynamo only uses the `get_module()` API since it is more flexible. 

Could we rewrite `train()`/`eval()` to just use the `get_module()` API?  That might make torchbecnh easier to maintain.

xuzhao9:
@jansel I would still prefer to keep `train()` and `eval()` as-is, because model code can be very diverse, and I feel `get_module()` is over-simplifying the actual train/eval loops. For example, some models we would like to run more than 1 batch of data (like [fastNLP]()), some requires more than one `torch.nn.Module` to run an entire train/eval loop (like [soft_actor_critic](https://github.com/pytorch/benchmark/blob/main/torchbenchmark/models/soft_actor_critic/__init__.py#L199=), [dcgan](https://github.com/pytorch/benchmark/blob/main/torchbenchmark/models/dcgan/__init__.py#L223=))

jansel:
> @jansel I would still prefer to keep `train()` and `eval()` as-is, because model code can be very diverse, and I feel `get_module()` is over-simplifying the actual train/eval loops. For example, some models we would like to run more than 1 batch of data (like [fastNLP](https://github.com/pytorch/torchdynamo/issues/165#issuecomment-1113496749)), some requires more than one `torch.nn.Module` to run an entire train/eval loop (like [soft_actor_critic](https://github.com/pytorch/benchmark/blob/main/torchbenchmark/models/soft_actor_critic/__init__.py#L199=), [dcgan](https://github.com/pytorch/benchmark/blob/main/torchbenchmark/models/dcgan/__init__.py#L223=))

Could we just do something custom for the few models that do weird stuff, then use a generic loop for most other models?

xuzhao9:
@jansel  Sure, I can implement a general version of `train()` and  `eval()` using `get_module()` in the base model.

Issue #1046, opened 2022-07-20
How to add an new backend?


zzpmiracle:
Hello, I want to add an new backend to run benchmark **without** modify this repo's code. In torchdynamo repo, I use @create_backend decorator to finish this, but I can't find suitable interface in this repo.

xuzhao9:
Can you specify what backend you want to add? We automatically support all backends available in `torchdynamo.list_backends()` (https://github.com/pytorch/benchmark/blob/main/torchbenchmark/util/backends/torchdynamo.py#L16)

xuzhao9:
We already support torchdynamo frontend, so if you have already added the BladeDISC backend for torchdynamo, you don't need to modify this repo, you can just use something like `--torchdynamo BladeDISC`, if you have already registered it in torchdynamo (i.e., BladeDISC is available in `torchdynamo.list_backends()`).

For your request `add an interface to register backends`, I still need some context. Do you mean you would like to run BladeDISC directly on eager mode, without torchscript or torchdynamo frontend, or do you mean you would like an option for torchscript to add backend?

zzpmiracle:
Yes, we'd like to add backends just for torchscript now. 
> We can consider some torchdynamo-like interface here, e.g., use decorator to register new eager or torchscript backends.

Yeah, this would be fascinating for the ecology of compiler. And I think it's necessary when more backends want to use this benchmark repo.

xuzhao9:
@zzpmiracle I created https://github.com/pytorch/benchmark/pull/1059 to make it easier to add more backends.

zzpmiracle:
@xuzhao9 Thanks! I think it's very convinient now to add more backends. And maybe I can change other backends in this repo in this way later.
Thanks again!

xuzhao9:
Closed by #1059

PR #1059, opened 2022-07-26
Make it easier to add more backends

xuzhao9:
This PR makes it easier to add more backends. Fixes https://github.com/pytorch/benchmark/issues/1046
To add a new backend:
 
1. create a file, say `jit.py`, in https://github.com/pytorch/benchmark/tree/main/torchbenchmark/util/backends.
2. In the file, create a function, say `def torchsript(model, backend_args)`, to transform the model using your backend, or add run contexts to it.
3. In https://github.com/pytorch/benchmark/blob/5a857e97721019eb0aa622976a8ac378cd4f2fbb/torchbenchmark/util/backends/__init__.py#L35, register your backend with import, e.g., `from .jit import torchscript`.

You can now try your own backend with your own backend:
`python run.py <model-name> --backend <backend-name> <backend-options`, for example:
```
$ python run.py resnet50 --backend torchscript
CPU Total Wall Time: 667.329 milliseconds
$ python run.py resnet50 --backend torchscript --no-ofi
CPU Total Wall Time: 1633.694 milliseconds
```

### DIFF
--- a/torchbenchmark/util/backends/__init__.py
+++ b/torchbenchmark/util/backends/__init__.py
@@ -1,0 +1,37 @@
+"""
+Utils for managing backends
+"""
+import functools
+
+BACKENDS = dict()
+
+def create_backend(fn):
+    @functools.wraps(fn)
+    def inner(model: 'torchbenchmark.util.model.BenchmarkModel', **kwargs):
+        if model is None:
+            return None
+
+        try:
+            fn(model, **kwargs)
+        except KeyboardInterrupt:
+            raise
+        except Exception as e:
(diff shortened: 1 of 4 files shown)


erichan1 (approved):
Great that we have a common place to put model transformation/compilation backends. Any other backends planned?

xuzhao9:
> Great that we have a common place to put model transformation/compilation backends. Any other backends planned?

@erichan1 Alibaba developers want to add their backend of `torch_blade`: https://github.com/pytorch/benchmark/issues/1046

PR #1072, opened 2022-08-01
seperate atol and rtol in correct check and read tol from os.env

zzpmiracle:
Some backends may not have enough precision for some tests, so can we add these environment variables("TORCHBENCH_ATOL" and "TORCHBENCH_RTOL") to dynamicly contol tolerance of correctness_check？


xuzhao9 (approved):
LGTM!

PR #1073, opened 2022-08-01
shorten output format

zzpmiracle:
Currently, the output results are too long when three or more backends run together. So we can shorten this output format.

### DIFF
--- a/.github/scripts/bmutils/summarize.py
+++ b/.github/scripts/bmutils/summarize.py
@@ -81,12 +81,12 @@ def find_result_by_header(r, header, base_arg):
             return "N/A"
     elif tp == "latency":
         if is_ok(r[args]):
-            return r[args]["results"]["latency_ms"]
+            return round(r[args]["results"]["latency_ms"], 3)
         else:
             return r[args]["status"]
     elif tp == "speedup":
         if is_ok(r[base_arg]) and is_ok(r[args]):
-            return round(r[base_arg]["results"]["latency_ms"] / r[args]["results"]["latency_ms"], 2)
+            return round(r[base_arg]["results"]["latency_ms"] / r[args]["results"]["latency_ms"], 3)
         else:
             return "N/A"
     else:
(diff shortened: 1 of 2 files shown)


.github/scripts/bmutils/summarize.py
```diff
@@ -86,7 +86,7 @@ def find_result_by_header(r, header, base_arg):
             return r[args]["status"]
     elif tp == "speedup":
         if is_ok(r[base_arg]) and is_ok(r[args]):
-            return r[base_arg]["results"]["latency_ms"] / r[args]["results"]["latency_ms"]
+            return round(r[base_arg]["results"]["latency_ms"] / r[args]["results"]["latency_ms"], 2)
```
xuzhao9, line 89: Let's keep 3 decimals here as well.

run_sweep.py
```diff
@@ -106,6 +106,7 @@ def _run_model_test(model_path: pathlib.Path, test: str, device: str, jit: bool,
         num_batches = task.get_model_attribute("NUM_BATCHES")
         if num_batches:
             result.results["latency_ms"] = result.results["latency_ms"] / num_batches
+        result.results["latency_ms"] = round(result.results["latency_ms"], 3)
```
xuzhao9, line 109: I think it is still good to have precise latency numbers here, what do you think?

.github/scripts/bmutils/summarize.py
zzpmiracle: OK.

run_sweep.py
zzpmiracle: It's rational to keep the original data. How about changeing it in the summarize.py, so that we don't change the original data, and maker the output shorter.

fortianyou: I think we can add a parameter such as pretty mode or max-precision when printing out the summary, and left the original data as it is.

xuzhao9 (approved):
LGTM!

PR #1077, opened 2022-08-01
add support for Execution Graph

FindHao:
Add two options in `run.py` to export execution graph.

 - `--profile-eg` will collect execution graph by [PARAM](https://github.com/facebookresearch/param)
 - `--profile-eg-folder [folder_path]` specifies the folder to save generated execution graph.


xuzhao9 (approved):
Thanks!

PR #1078, opened 2022-08-02
Fix eval batch size on Nvidia A100 40GB

xuzhao9:
Use `torch.cuda.device_name()` to specify the eval batch size on GPU devices.

### DIFF
--- a/torchbenchmark/models/pyhpc_equation_of_state/metadata.yaml
+++ b/torchbenchmark/models/pyhpc_equation_of_state/metadata.yaml
@@ -1,3 +1,6 @@
+devices:
+  NVIDIA A100-SXM4-40GB:
+    eval_batch_size: 16777216
 eval_benchmark: false
 eval_deterministic: false
 eval_nograd: true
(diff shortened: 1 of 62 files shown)


torchbenchmark/models/pyhpc_isoneutral_mixing/metadata.yaml
```diff
@@ -1,3 +1,6 @@
+devices:
+  NVIDIA A100-SXM4-40GB:
+    eval_batch_size: 67108864
```
FindHao, line 3:
```suggestion
    eval_batch_size: 1048576
```
In our sheet, its TFLOPS is only 0.01 and I think we can choose the first 0.01 to save execution time.

torchbenchmark/models/pyhpc_equation_of_state/metadata.yaml
```diff
@@ -1,3 +1,6 @@
+devices:
+  NVIDIA A100-SXM4-40GB:
+    eval_batch_size: 16777216
```
FindHao, line 3:
```suggestion
    eval_batch_size: 1
```
In our sheet, its TFLOPS is always 0.

FindHao (changes requested):

FindHao (approved):
LGTM

PR #1101, opened 2022-08-11
Enable CPU profiling when device='cuda' and profiling turned on

davidberard98:
* __->__ #1101

When profiling CUDA models, I think we should collect both CUDA and CPU activity. Without CPU activity profiling turned on, we can't see which CPU events are responsible for launching the cuda kernels.

Differential Revision: [D38638497](https://our.internmc.facebook.com/intern/diff/D38638497)


davidberard98:
@davidberard98 has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D38638497).

aaronenyeshi (approved):
LGTM

PR #1110, opened 2022-08-15
[correctness check] added use_cosine_similarity as an extra argument

snadampal:
The default allclose() correctness check doesn't work for bfloat16
kernels due to precision differences between fp32 and bfloat16. For trt
kernels this is addressed by switching to cosine_similarity. Instead
of adding flags for every such runtime, this patch adds "use_consine_similarity"
as an extra argument for benchmark execution to explicitly use this correctness
check.


xuzhao9:
Just curious, how are you testing the bfloat16 precision? Currently we only support float32 and float16 officially.

snadampal:
Hi @xuzhao9 I'm testing it with OneDNN + ArmComputeLibrary. OneDNN supports fast fpmath mode to enable Bfloat16 kernels.
`DNNL_DEFAULT_FPMATH_MODE=BF16`

xuzhao9 (approved):
LGTM!

PR #1121, opened 2022-08-17
Add ability to set distributed to none for hf_bert e2e benchmark

erichan1:
Currently, setting --distributed none when running hf_bert from run_e2e.py is unsupported. fix that.


xuzhao9 (approved):
Good fix!

PR #1141, opened 2022-08-30
Setup CUDA environment when running bisection and abtest

xuzhao9:
CUDA 11.3 is being deprecated, and we are using environment variables to control the CUDA version instead of changing the link target of `/usr/local/cuda`.

This has many benefits:
- switching cuda version no longer needs sudo permission
- support switching cuda versions at both runtime and compile time

This PR also creates a `cuda_utils.py` to place all CUDA version related code.


davidberard98 (approved):
Thanks!

PR #1158, opened 2022-09-06
Checkout file from S3 instead of git lfs

xuzhao9:
git-lfs seems to be unstable when working cross-repository (showing auth error). Since our demand is very simple, let's use S3 instead to manage the input data.


scripts/install_basics.sh
```diff
@@ -18,6 +18,15 @@ conda install -y python=3.8
 # install unittest-xml-reporting
 pip install unittest-xml-reporting
 
+# install CUDA 11.6
```
desertfire, line 21: Why do we need CUDA 11.6 for this reading from S3 change?

xuzhao9: This is to fix the Circle CI test error, seems 11.3 is not supported anymore.

desertfire (approved):

PR #1165, opened 2022-09-07
Rewrite the torchdynamo optimizer to use decorator

xuzhao9:
TorchDynamo is deprecating context manager and is using decorator instead. We need to update the dynamo related code.

### DIFF
--- a/torchbenchmark/util/backends/torchdynamo.py
+++ b/torchbenchmark/util/backends/torchdynamo.py
@@ -2,7 +2,6 @@
 Support TorchDynamo(https://github.com/facebookresearch/torchdynamo) backends
 """
 import argparse
-import functools
 from typing import List
 import torchdynamo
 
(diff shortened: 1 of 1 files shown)


torchbenchmark/util/backends/torchdynamo.py
```diff
@@ -17,7 +16,11 @@ def parse_torchdynamo_args(model: 'torchbenchmark.util.model.BenchmarkModel', dy
 
 def apply_torchdynamo_args(model: 'torchbenchmark.util.model.BenchmarkModel', args: argparse.Namespace, precision: str):
     if args.torchdynamo == "fx2trt" and precision == "fp16":
-        model.add_context(functools.partial(torchdynamo.optimize, torchdynamo.optimizations.backends.fx2trt_compiler_fp16))
+        dynamo_optimizer = torchdynamo.optimize(torchdynamo.optimizations.backends.fx2trt_compiler_fp16)
     else:
-        model.add_context(functools.partial(torchdynamo.optimize, args.torchdynamo))
+        dynamo_optimizer = torchdynamo.optimize(args.torchdynamo)
+    if model.test == "train":
+        model.train = dynamo_optimizer(model.train)
```
wconstab, line 23: the part about using dynamo_optimizer over a function looks right to me.

the dynamo harness applies the optimizer over the 'model_iter_fn' which is the equivalent to torchbench's `model.train` and `model.eval`
https://github.com/pytorch/torchdynamo/blob/main/benchmarks/common.py#L936
https://github.com/pytorch/torchdynamo/blob/main/benchmarks/common.py#L1640

This is the torchbench 'model_iter_fn' (separate ones for train/infr)
https://github.com/pytorch/torchdynamo/blob/main/benchmarks/torchbench.py#L338

as you can see, there are some differences which lead to different things being measured and sometimes different failure cases.  It'd be my preference that the two suites used the same iter function to avoid this divergence.

cc @anijain2305

wconstab (approved):
although on second thought... go ahead and land this at least, it is a step in the right direction

anijain2305 (approved):

PR #1167, opened 2022-09-07
Check type of example inputs from model.get_module()

nmacchioni:
The example test checker unpacks example inputs either as **kwargs, if the example inputs are a dict instance, or as *args by default. We should only unpack as *args if the example inputs are a tuple instance, otherwise raise an assert error by default. Unpacking non-tuple inputs as *args can lead to tricky dimension mismatch errors.


xuzhao9 (commented):
I think we also allow example_inputs to be a list (sometimes the input is a list of objects and it is accepted by the * operator)

xuzhao9 (approved):
LGTM

PR #1171, opened 2022-09-08
Remove the use of niter in the model code

xuzhao9:
To get more consistent results with torchdynamo's benchmark script, let's completely deprecate the niter argument.


wconstab (approved):
💯

PR #1173, opened 2022-09-09
Use capturable=True for huggingface & vision models

davidberard98:
* __->__ #1173

Copied from this branch:
https://github.com/pytorch/benchmark/tree/wconstab/ddp_experiments

This is necessary to make the optimizer capturable by dynamo.


davidberard98:
This was implemented using `ADAM_CAPTURABLE=1` environment variable instead.

PR #1179, opened 2022-09-14
maml_omniglot model with functorch

zou3519:
Very similar to the maml_omniglot model in torchbench (which is what we originally based this model off of). We also load some sample inputs from there because the model takes the same inputs.

Test Plan:
- python test.py -k TestBenchmark.test_functorch_maml_omniglot_train_cuda

### DIFF
--- a/torchbenchmark/models/functorch_maml_omniglot/__init__.py
+++ b/torchbenchmark/models/functorch_maml_omniglot/__init__.py
@@ -1,0 +1,105 @@
+import torch
+import torch.optim as optim
+import torch.nn as nn
+import torch.nn.functional as F
+from functorch import make_functional_with_buffers, vmap, grad
+import functools
+from pathlib import Path
+from typing import Tuple
+
+from ...util.model import BenchmarkModel
+from torchbenchmark.tasks import OTHER
+
+
+def loss_for_task(net, n_inner_iter, x_spt, y_spt, x_qry, y_qry):
+    params, buffers, fnet = net
+    querysz = x_qry.size(0)
+
+    def compute_loss(new_params, buffers, x, y):
+        logits = fnet(new_params, buffers, x)
+        loss = F.cross_entropy(logits, y)
+        return loss
+
+    new_params = params
+    for _ in range(n_inner_iter):
+        grads = grad(compute_loss)(new_params, buffers, x_spt, y_spt)
+        new_params = [p - g * 1e-1 for p, g, in zip(new_params, grads)]
+
+    # The final set of adapted parameters will induce some
+    # final loss and accuracy on the query dataset.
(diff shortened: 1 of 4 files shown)


torchbenchmark/models/functorch_maml_omniglot/__init__.py
```diff
@@ -0,0 +1,105 @@
...
+    qry_loss = F.cross_entropy(qry_logits, y_qry)
+    qry_acc = (qry_logits.argmax(
+        dim=1) == y_qry).sum() / querysz
+
+    return qry_loss, qry_acc
+
+
+class Model(BenchmarkModel):
+    task = OTHER.OTHER_TASKS
+    DEFAULT_TRAIN_BSIZE = 1
```
xuzhao9, line 41: Is this the batch size used in the original training code? I am working on another version of maml/protonet originated from fewshot(https://github.com/pytorch/benchmark/pull/1184/files#diff-9c209c92cea7ccad363e7d477e210a3120028678829b11086cd9772af97181eaR9) and it looks like the batch size should be larger than this.

xuzhao9 (approved):
Overall LGTM. See inline comments about the batch size we should use.

torchbenchmark/models/functorch_maml_omniglot/__init__.py
zou3519: I copied this from https://github.com/pytorch/benchmark/blob/8045bb7d126c1cad4aed212b48a35478db8ec12b/torchbenchmark/models/maml_omniglot/__init__.py#L33-L34.

"batch_size" is a term that doesn't apply to the maml model. The original maml omniglot (https://arxiv.org/pdf/1703.03400.pdf) was trained under 32 tasks (also referred to as meta batch-size) and done using both 1-shot and 5-shot.

The code provided in this PR does 32 tasks and 5-shot. I can change `DEFAULT_TRAIN_BSIZE` to 32 instead and add a comment if that sounds reasonable to you? (The first dimension of the inputs is the task dimension and has size 32)

xuzhao9: I see,  `maml_omniglot` and `maml` models were marked as "incorrect" in https://github.com/pytorch/benchmark/issues/328, so we created https://github.com/pytorch/benchmark/pull/1184/ trying to use an actual correct implementation, but I haven't go time to get it work. I am okay with landing this PR for now, but I think we should investigate if https://github.com/pytorch/benchmark/pull/1184 is a better implementation, then rebase all maml models to fewshot.

zou3519: I will merge this for now. Happy to rebase this model if we do want to replace it with the one in the fewshot repo.

PR #1180, opened 2022-09-14
Add lennard-jones model

zou3519:
The main thing we want to test here is that using functorch's vmap x jacrev on some fully connected layers works. This is a common sequence of operations for functorch and this PR is an example of one such model that exercises it.

Test Plan:
- `python test.py -k "test_lennard_jones_eval_cuda"`
- `python test.py -k "test_lennard_jones_train_cuda"`

### DIFF
--- a/torchbenchmark/models/lennard_jones/__init__.py
+++ b/torchbenchmark/models/lennard_jones/__init__.py
@@ -60,9 +60,7 @@ class Model(BenchmarkModel):
         )
         self.model = self.model.to(device)
 
-        if batch_size is None:
-            batch_size = Model.DEFAULT_TRAIN_BSIZE
-        r = torch.linspace(0.5, 2 * sigma, steps=batch_size, requires_grad=True)
+        r = torch.linspace(0.5, 2 * sigma, steps=self.batch_size, requires_grad=True)
 
         # Create a bunch of vectors that point along positive-x.
         # These are the dummy inputs to the model.
--- a/torchbenchmark/models/lennard_jones/__init__.py
+++ b/torchbenchmark/models/lennard_jones/__init__.py
@@ -70,6 +70,7 @@ class Model(BenchmarkModel):
 
         # Generate some dummy targets based off of some interpretation of the lennard_jones force.
         norms = torch.norm(self.drs, dim=1).reshape(-1, 1)
+        self.norms = norms
         # Create training energies
         self.training_energies = torch.stack(list(map(lennard_jones, norms))).reshape(-1, 1)
         # Create forces with random direction vectors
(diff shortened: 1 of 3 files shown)


torchbenchmark/models/lennard_jones/metadata.yaml
```diff
@@ -0,0 +1,10 @@
+devices:
+  NVIDIA A100-SXM4-40GB:
+    eval_batch_size: 1000
```
zou3519, line 3: @xuzhao9 I'm not sure how this is supposed to be determined, I just put the default batch size that the above script runs with.

xuzhao9: Yeah the default batch size is fine. We have a script to generate the optimal batch size for different devices based on FLOPS result, but we can save that work later.

torchbenchmark/models/lennard_jones/__init__.py
```diff
@@ -0,0 +1,101 @@
...
+            nn.Linear(16, 16),
+            nn.Tanh(),
+            nn.Linear(16, 16),
+            nn.Tanh(),
+            nn.Linear(16, 1)
+        )
+        self.model = self.model.to(device)
+
+        if batch_size is None:
+            batch_size = Model.DEFAULT_TRAIN_BSIZE
```
xuzhao9, line 64: `super().__init__()` should handle the correct setting of batch size so this is not needed.

torchbenchmark/models/lennard_jones/__init__.py
```diff
@@ -0,0 +1,101 @@
...
+        optimizer.zero_grad()
+        energies, forces = make_prediction(model, self.drs)
+        loss = loss_fn(self.training_energies, self.training_forces, energies, forces)
+        loss.backward()
+        optimizer.step()
+
+    def eval(self) -> Tuple[torch.Tensor]:
+        model = self.model
+        model.eval()
+        with torch.no_grad():
```
xuzhao9, line 99: Maybe we should use `torch.inference_mode()` api instead?

xuzhao9 (approved):
Overall LGTM. See minor inline comments.

torchbenchmark/models/lennard_jones/__init__.py
zou3519: Inference mode doesn't work with the forward pass, because the forward pass computes jacobians and inference mode turns autograd off.

(torch.no_grad still works though -- functorch transforms still work in a no_grad block as opposed to inference mode, which really disables autograd).

torchbenchmark/models/lennard_jones/__init__.py
zou3519: Done, thanks for pointing it out.

Issue #405, opened 2021-06-15
Errors in running with latest version of pytorch


t-naveenkumar:
I am getting errors in running some networks with latest stable version of PyTorch for CPU

## commands for reproducing the issue
`conda create -n stable python=3.7 -y`
`conda activate stable`
`conda install -y pytorch torchtext torchvision cpuonly -c pytorch`
`git clone  https://github.com/pytorch/benchmark.git`
`cd benchmark`
`python install.py`
`pytest test_bench.py::TestBenchNetwork::test_eval[resnet50-freeze-cpu-jit] --fuser old --benchmark-json old.json --ignore_machine_config`

I am getting following error. 
[error_log.txt](https://github.com/pytorch/benchmark/files/6653651/error_log.txt)

I am not getting the same error for nightly build though.

Please let me know if there is need for more information.

PR #1190, opened 2022-09-17
add infra and models for ddp experiments

wconstab:


torchbenchmark/util/framework/vision/model_factory.py
```diff
@@ -25,7 +25,7 @@ def __init__(self, model_name, test, device, jit=False, batch_size=None, extra_a
             self.example_outputs = torch.rand_like(self.model(*self.example_inputs))
             self.model.train()
             # setup optimizer and loss_fn
-            self.optimizer = optim.Adam(self.model.parameters())
+            self.optimizer = optim.Adam(self.model.parameters(), capturable=True)
```
davidberard98, line 28: I'm guessing we probably shouldn't land this because it breaks on non-cuda:

https://github.com/pytorch/pytorch/blob/84c4b079328c2a97d78d95c47b841f8dca6036bb/torch/optim/adam.py#L262

Or maybe we can use `capturable=(device == 'cuda')`. Although maybe this is also bad for non-dynamo use cases (based on Natalia's comment that perf is better when `capturable=False`)

torchbenchmark/util/framework/huggingface/model_factory.py
```diff
@@ -58,7 +61,7 @@ def __init__(self, name, test, device, jit=False, batch_size=None, extra_args=[]
             config.num_buckets = 128
         class_ctor = getattr(transformers, class_models[name][3])
         self.model = class_ctor.from_config(config).to(device)
-        self.optimizer = optim.Adam(self.model.parameters(), lr=0.001)
+        self.optimizer = optim.Adam(self.model.parameters(), lr=0.001, capturable=True)
```
davidberard98, line 64: same here

ddp_experiments.py
```diff
@@ -0,0 +1,282 @@
+import argparse
```
xuzhao9, line 1: Instead of creating many experiment scripts in the root directory, can we wrap them into userbenchmarks and place them in the `userbenchmark/` directory? For example, can we place them in the `userbenchmark/ddp_experiment`, and then users can run the benchmark using `python run_benchmark.py ddp_experiment`?

wconstab: umm, i can move the script under that folder.  but since the script is really just a launcher for submitting many slurm jobs i'm not sure if it's good to try to further wrap it underneath another 'run_benchmark' layer.

torchbenchmark/util/framework/vision/model_factory.py
wconstab: yea i'll remove this for now

i guess we'll have to just keep a dev branch while we work out how to properly fix this.

or maybe i'll make this an ENV flag so we can just stick on main branch and workaround till fixed.

ddp_experiments.py
xuzhao9: Wrapping it under `run_benchmark` layer gives many benefits. It automatically gives us ability to run it through Github Actions (such as https://github.com/pytorch/benchmark/actions/workflows/userbenchmark-t4-metal.yml), so that users can use our CI infra to get benchmark results, and it will also connect to internal perf dashboard to track the metrics values over time.

And it keeps the root directory cleaner and more organized as well.

userbenchmark/distributed/ddp_experiments.py
```diff
@@ -0,0 +1,282 @@
+import argparse
```
xuzhao9, line 1: Since this experiment is owned by compiler team, not distributed, I suggest to move this to `userbenchmark/ddp_experiment/__init__.py`

torchbenchmark/util/framework/vision/model_factory.py
davidberard98: Will add `ADAM_CAPTURABLE` env flag for this.

ddp_experiments.py
davidberard98: @xuzhao9 I moved the scripts from `userbenchmark/distributed` to `userbenchmark/ddp_experiments`. However I have not integrated the experiments with userbenchmark layer. Currently the ddp experiment benchmarks aren't runnable with pytorch as-is, due to some dynamo / pytorch bugs (we need to patch some changes in pytorch to workaround the issue). So we wouldn't be able to benefit from the CI or internal perf dashboard anyway.

Would it be okay to leave the scripts as they are (standalone runnable scripts) for ease of development, until we've fixed those issues in pytorch & dynamo?

torchbenchmark/models/hf_GPT2_large/metadata.yaml
```diff
@@ -0,0 +1,8 @@
+eval_benchmark: false
+eval_deterministic: false
+eval_nograd: true
+train_benchmark: false
+train_deterministic: false
+not_implemented:
```
xuzhao9, line 6: Can you add `device: cuda` here to skip the CUDA unit test? Looks like it caused CUDA OOM.

userbenchmark/ddp_experiments/__init__.py
```diff
@@ -0,0 +1,283 @@
...
+        os.environ["NCCL_DEBUG_SUBSYS"] = 'INIT,ENV,NET'
+        os.environ['NCCL_SOCKET_IFNAME'] = 'ens'
+        # os.environ["NCCL_ALGO"] = 'ring'
+        os.environ["FI_PROVIDER"] = 'efa'
+        os.environ["FI_EFA_USE_DEVICE_RDMA"]= str(1)
+        os.environ["NET_TYPE"] = 'efa'
+
+
+def main():
+    args = parse_args()
```
xuzhao9, line 183: Can we have an example command to run this userbenchmark and an output json text? Would like to know how it looks like.

torchbenchmark/util/framework/vision/model_factory.py
```diff
@@ -25,7 +26,11 @@ def __init__(self, model_name, test, device, jit=False, batch_size=None, extra_a
             self.example_outputs = torch.rand_like(self.model(*self.example_inputs))
             self.model.train()
             # setup optimizer and loss_fn
-            self.optimizer = optim.Adam(self.model.parameters())
+            self.optimizer = optim.Adam(
```
xuzhao9, line 29: Will this break the benchmarks when running with older pytorch builds? When was the `capturable=` argument introduced?

torchbenchmark/util/distributed/core_model/apply_trainer.py
```diff
@@ -1,18 +1,29 @@
...
+            # If buffer broadcast is necessary, specific optimizations might be
+            # necessary to optimize performance. Disable it by default.
+            broadcast_buffers=False,
+            # Set gradient as bucket view to avoid unnecessary copies
+            gradient_as_bucket_view=True,
+            # TODO: tune bucket_cap_mb
+            static_graph=True,
+        )
+        return ddp_model
+    elif trainer == "fsdp":
```
xuzhao9, line 23: Thanks for adding the FSDP support!

xuzhao9 (changes requested):
Looks nice, but let's fix the broken CI test and make sure the inline comments are addressed. Thanks!

torchbenchmark/util/framework/vision/model_factory.py
davidberard98: @xuzhao9 it was added in june 2022 (https://github.com/pytorch/pytorch/pull/77862), is that old enough? Otherwise I think it's okay to remove it

davidberard98: Also could you explain the reason to be concerned about running with older pytorch builds, and what the guidelines are for how long the feature should be in pytorch before it's safe to include in a benchmark here? Since these tests are targeted for ddp testing, we'll probably eventually start adding some features that depend on new dynamo features.

userbenchmark/ddp_experiments/__init__.py
davidberard98: Added a README.md (modeled after the distributed userbenchmark) - does that work?

torchbenchmark/models/hf_GPT2_large/metadata.yaml
davidberard98: Thanks, I added this and the test is now passing.

userbenchmark/ddp_experiments/__init__.py
xuzhao9: Yes, adding `README.md` sounds good to me.

torchbenchmark/util/framework/vision/model_factory.py
xuzhao9: I  am asking this because we sometimes receive GitHub issues of users complaining the benchmark doesn't run on the latest stable release (such as https://github.com/pytorch/benchmark/issues/405) and we have to add bc-compatible layers to handle it.

ddp_experiments.py
xuzhao9: Yeah, it is okay to me if you prefer not to use userbenchmark/CI feature and only from command line right now.

xuzhao9 (approved):
LGTM. Thanks for improving this PR!

PR #1194, opened 2022-09-20
[Dont Merge] For TorchDynamo DDP collaboration

anijain2305:
cc @davidberard98 @wconstab


torchbenchmark/util/model.py
```diff
@@ -130,6 +124,13 @@ def __post__init__(self):
             else:
                 module, _inputs = self.get_module()
             self.set_module(apply_trainer(module, self.dargs.distributed))
+
+        # apply optimization args
```
wconstab, line 128: oh, this changes the ordering so DDP is applied first? i think i was assuming this to be the case and did it this way in my local test scripts.  maybe this was one of the major reasons for seeing issues on torchbench.  if ddp goes on after dynamo wraps the model, what can we do?

PR #1198, opened 2022-09-21
Remove incorrect MAML implementation

zou3519:
Fixes https://github.com/pytorch/benchmark/issues/328

Here's some context:
- We discovered that the implementation doesn't actually use maml (https://github.com/dragen1860/MAML-Pytorch/issues/59)
- We filed https://github.com/pytorch/benchmark/issues/328
- We added maml_omniglot https://github.com/pytorch/benchmark/pull/349 as the correct version of of the maml model.
- We didn't delete the maml model (because I was worried that it was doing a "different" type of maml that I hadn't seen before that is still valid).

The last step to resolve this issue is to delete the incorrect MAML example, unless we have reasons to keep it around.


xuzhao9:
It makes sense to me to remove maml and keep maml_omniglot as the correct impl.
Another stakeholder of the maml model is the torchdynamo team, @jansel are you okay with this PR?

jansel:
`maml` is one of the hardest models for dynamo to handle because it involves a `copy.deepcopy()` and some dynamic autograd during inference.

I think it is pretty useful and would like to keep it, though perhaps more of a challenging test case and less as a benchmark.

xuzhao9:
I see, let's keep this model then. I will close this PR, and create a separate PR to mark maml as correctness test only and will not be used for performance testing.

Thanks @jansel and @zou3519 !

PR #1204, opened 2022-09-22
Add stable diffusion model benchmark

xuzhao9:
Add stable diffusion model from huggingface diffuser repository.

Running eval method from stable_diffusion on cuda in eager mode with input batch size 1.
GPU Time:            2921.900 milliseconds
CPU Total Wall Time: 2921.957 milliseconds
GPU 0 Peak Memory:              8.4949 GB
CPU Peak Memory:                3.9766 GB

Fixes https://github.com/pytorch/benchmark/issues/1374

To adopt it to the core model set, we still need to do the following:
1. Disable loading model from checkpoint, if checkpointing doesn't affect data flow and performance
2. Run only 1 iteration in the eval
3. Disable noisy output


torchbenchmark/canary_models/stable_diffusion/install.py
```diff
@@ -0,0 +1,12 @@
...
+import torch
+
+MODEL_NAME = "stabilityai/stable-diffusion-2"
+
+def load_model_checkpoint():
+    StableDiffusionPipeline.from_pretrained(MODEL_NAME, torch_dtype=torch.float16, safety_checker=None)
+
+if __name__ == '__main__':
+    install_diffusers()
+    load_model_checkpoint()
```
msaroufim, line 12: Should we make this an optional step?

msaroufim (changes requested):

torchbenchmark/canary_models/stable_diffusion/install.py
xuzhao9: I am not sure how to make diffuser models not loading checkpoint - I will address that in another PR.

PR #1222, opened 2022-09-29
Update ddp+dynamo benchmark arguments

davidberard98:
* __->__ #1222

* Reduce default timeout to 2 hours; afaik, all of these pass well
within 2 hours unless there are configuration issues (e.g. bad nodes,
etc.). Reduced timeout means that we avoid blocking the cluster while
waiting for a dead job, which also means that _our_ experiment will also
probably finish faster because it's not blocked on dead jobs.
* Add `--exclude` argument which can be used to avoid bad nodes.

Differential Revision: [D39952268](https://our.internmc.facebook.com/intern/diff/D39952268)

### DIFF
--- a/userbenchmark/ddp_experiments/__init__.py
+++ b/userbenchmark/ddp_experiments/__init__.py
@@ -48,7 +48,7 @@ def parse_args(args: List[str]=None):
 
     parser.add_argument(
         "--timeout",
-        default=1440,
+        default=120,
         type=int,
         help="Duration of the job"
     )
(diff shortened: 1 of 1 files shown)


davidberard98:
@davidberard98 has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D39952268).

wconstab (approved):
lgtm! 💯

xuzhao9 (approved):

PR #1226, opened 2022-09-30
Required changes after cuBLAS workspace allocation change

eqy:
The cuBLAS workspace PR https://github.com/pytorch/pytorch/pull/85447 moves allocations that were previously managed by cuBLAS to the `CUDACachingAllocator`. While the lifetime of this memory has not changed, it is now visible to `torch.cuda.memory_allocated()` which can throw false-positives in testing results. A workaround to get an accurate memory measurement is to clear all allocated cuBLAS workspaces via `torch._C._cuda_clearCublasWorkspaces()` which was also added in this PR.

Do not merge this for now as the cuBLAS workspace PR has been reverted.

CC @xuzhao9 @seemethere

### DIFF
--- a/torchbenchmark/__init__.py
+++ b/torchbenchmark/__init__.py
@@ -520,10 +520,11 @@ class ModelTask(base_task.TaskBase):
         if skip or os.getenv('PYTORCH_TEST_SKIP_CUDA_MEM_LEAK_CHECK', '0') == '1':
             yield
             return
-
+        torch._C._cuda_clearCublasWorkspaces()
         self.gc_collect()
         memory_before = self.worker.load_stmt("torch.cuda.memory_allocated()")
         yield
+        torch._C._cuda_clearCublasWorkspaces()
         self.gc_collect()
         assert_equal(
             memory_before,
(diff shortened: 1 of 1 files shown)


xuzhao9:
In `watch_cuda_memory()`, we isolate the test in a subprocess, so you can't directly run `torch._C._cuda_clearCublasWorkspaces()`, can you use `self.worker.load_stmt("torch._C._cuda_clearCublasWorkspaces()")` instead?

eqy:
@xuzhao9 This PR has been updated and the corresponding change to PyTorch was merged in https://github.com/pytorch/pytorch/commit/352d9264822b8064b0c0792bc00492e69e569a37. What would be the best way to insure that this change to `benchmark` is added at the right time? Could we use some kind of check that `torch._C` has the `_cuda_clearCublasWorkspaces` function?

xuzhao9:
> @xuzhao9 This PR has been updated and the corresponding change to PyTorch was merged in [pytorch/pytorch@352d926](https://github.com/pytorch/pytorch/commit/352d9264822b8064b0c0792bc00492e69e569a37). What would be the best way to insure that this change to `benchmark` is added at the right time? Could we use some kind of check that `torch._C` has the `_cuda_clearCublasWorkspaces` function?

Maybe something like `torch._C._cuda_clearCublasWorkspaces if hasattr(torch._C, '_cuda_clearCublasWorkspaces') else None` ?

eqy:
OK, I wasn't sure if adding `torch` as an import would be correct, checking now in CI

xuzhao9:
Can you rebase this PR on the lastest main? The CI failed with a pre-existing error that has been fixed in trunk.

But from the output it looks like the problem is solved.

PR #1229, opened 2022-10-03
Run a100 workflow with docker

xuzhao9:


xuzhao9:
Deprecated, we only need 1 A100 runner workflow.

PR #1245, opened 2022-10-14
ddp experiments: run all measurements on the same allocation

davidberard98:
Previously, we would launch a different slurm job for measurement, e.g.:
1) resnet50 w/ inductor + graph breaks, 2 node
2) resnet50 w/ inductor + NO graph breaks, 2 node
3) ...

But there's a of variation between different nodes - possibly due to network topology etc. These slurm jobs would often each be submitted to a different set of nodes, which would add a lot of noise to the data.

So instead, with this PR, we do the following:
* allocate enough nodes to run all the measurements
* launch (8 * max_nodes) jobs, and provide a list of measurements to each of the jobs
* in each job, we iterate through the list of measurements. For each measurement, we spawn a new process that runs the measurement.
* Once the new process exits, we synchronize via a barrier that's implemented via torch.distributed's FileStore.
* Note that if we have, say, 8 nodes, and one measurement only requires 4 nodes, then the other 4 nodes won't launch any jobs and will just sit idle waiting on the barrier.

Typical usage:
```
python userbenchmark/ddp_experiments/__init__.py --job_dir /full/path/to/shared/directory
```
then the results will be dumped in `/full/path/to/shared/directory`.  From there you can use `userbenchmark/ddp_experiments/parse_ddp.py` to view the results (follow up PR coming soon). Other options include:
```
--repeat [n] : repeats the experiments n times each so you can reduce the effect of any transient load on the machines or network
--profile True : turn on the profiler (note that for inductor, you need to turn off cuda graphs for the profiler to work)
--exclude node1,node2 : exclude a comma-separated list of nodes from the slurm allocation.
--timeout 600 : timeout in minutes  
```

### DIFF
--- a/userbenchmark/ddp_experiments/__init__.py
+++ b/userbenchmark/ddp_experiments/__init__.py
@@ -3,6 +3,7 @@ import importlib
 import os
 import copy
 import csv
+import dataclasses
 import io
 import json
 import multiprocessing
@@ -114,6 +115,12 @@ def parse_args(args: List[str]=None):
         default="",
         help="comma-separated list of nodes to exclude from the slurm allocation",
     )
+    parser.add_argument(
+        "--repeat",
+        type=int,
+        default=1,
+        help="number of times to repeat the experiments",
+    )
 
 
     try:
@@ -136,6 +143,9 @@ def get_init_file(args):
     return init_file
 
 
+# This implements a barrier function, where all processes wait until they all
+# reach the barrier() call.
+# rank: there should be one
 class FileBarrier:
     def __init__(self, rank, world_size, sync_file, timeout: Optional[timedelta] = None):
         self.rank = rank
@@ -146,6 +156,7 @@ class FileBarrier:
             timeout = timedelta(minutes=30)
         self.store.set_timeout(timeout)
         self.call_idx = 0
+        self.barrier()
 
     def barrier(self):
         self.call_idx += 1
@@ -329,6 +340,7 @@ def main():
         return "eager"
 
     experiments = []
+    for i in range(args.repeat):
         for nodes in node_list:
             for model_name in models:
                 for model_args in model_args_configs:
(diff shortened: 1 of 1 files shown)


userbenchmark/ddp_experiments/__init__.py
```diff
@@ -134,26 +143,95 @@ def get_init_file(args):
...
+                proc = multiprocessing.Process(target=self.run_once, args=(args, model_args, q))
+                proc.start()
+                proc.join()
+                print(f"exit code: {proc.exitcode} and result: {result_dict}")
+                if proc.exitcode == 0:
+                    result_dict['result'] = q.get()
+                else:
+                    result_dict['result'] = None
+                assert 'result' in result_dict
+                # wrap in <RESULT></RESULT> so we can parse partial results in the stdout logs
```
wconstab, line 228: this is fine, but is it not possible to just monitor some path on /fsx and see partial results show up as individual files there as they are written?

userbenchmark/ddp_experiments/__init__.py
```diff
@@ -253,46 +337,61 @@ def get_backend_name(model_args):
...
+                        breakname = "withbreaks" if has_breaks else "nobreaks"
+                        if has_breaks:
+                            copied_model_args.append("--optimize_dynamo_ddp")
+                        batch_size = model_batch_size[model_name]
+                        args.model = model_name
+                        args.batch_size = batch_size
+                        args.nodes = nodes
+                        args.dist_url = get_init_file(args).as_uri()
+                        args.output_dir = args.job_dir
+                        args_copy = copy.deepcopy(args)
```
wconstab, line 361: hmm i guess you are sure in this case you'll never use 'args.model' again but maybe it would be cleaner to first do args_copy = deepcopy(args), then do args_copy.model = model_name, so you never modify the root `args` obj?

wconstab (approved):
LGTM! could you also update the PR-Desc / commit message for the squashed commit to include the run command that you use for a typical run, so its easy for folks to replicate?

userbenchmark/ddp_experiments/__init__.py
davidberard98: we can already do that, but I forgot to include the changes to parse_ddp.py in this PR - I'll open another PR with those changes.

xuzhao9 (approved):

PR #1251, opened 2022-10-17
Update DDP result parser to gather partial results

davidberard98:
* __->__ #1251

Usage:

First, start a job with
```
python userbenchmark/ddp_experiments/__init__.py --job_dir /full/path/to/dir
```
Take note of the jobid that gets printed.

Then you can run the following to get the results in a more easily readable format:

```
python userbenchmark/ddp_experiments/parse_ddp.py --results_dir path/to/dir --job_id 12345
```

It will look up path/to/dir/12345_0_log.out and parse out results wrapped like `<RESULT>{RESULT_JSON}</RESULT>`

Parsing results that are dumped into the stdout logs means that we can view partial results before the job completes.

The `--csv_out` flag can be used to dump the results in csv format for use in google sheets, etc.

This PR relies on https://github.com/pytorch/benchmark/pull/1245.

Differential Revision: [D40454556](https://our.internmc.facebook.com/intern/diff/D40454556)

### DIFF
--- a/userbenchmark/ddp_experiments/parse_ddp.py
+++ b/userbenchmark/ddp_experiments/parse_ddp.py
@@ -1,4 +1,6 @@
 import csv
+import json
+import copy
 import argparse
 from typing import OrderedDict
 from dataclasses import dataclass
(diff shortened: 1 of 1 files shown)


davidberard98:
@davidberard98 has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D40454556).

userbenchmark/ddp_experiments/parse_ddp.py
```diff
@@ -8,47 +10,32 @@
...
-    with open(args.csv) as f:
-        runs = csv.DictReader(f)
-        for row in runs:
-            model = row["model"]
+    model_data = defaultdict(lambda: defaultdict(lambda: defaultdict(lambda: defaultdict(list))))
+    rank_id = 0
+    log_path = os.path.join(args.results_dir, f"{args.job_id}_{rank_id}_log.out")
+    with open(log_path, "r") as f:
+        content = f.read()
+        pieces = content.split("<RESULT>")
```
xuzhao9, line 24: How is the log file format defined? Will it be subjected to change and break this code?

xuzhao9 (approved):
LGTM with minor inline comments about log schema

userbenchmark/ddp_experiments/parse_ddp.py
davidberard98: The log format is defined in ddp_experiments.py, it's just the `print(f"<RESULT>...` statement. I can follow up with a PR that uses the pkl file by default and falls back to this only if the pkl file isn't available.

PR #1260, opened 2022-10-24
Update conda installation command to handle the installation difference with cudatoolkit and pytorch-cuda

weiwangmeta:
This PR fixes the conda installation issue with pytorch 1.12.1. It was observed that "conda install --force-reinstall pytorch=1.12.1 torchvision torchtext pytorch-cuda=11.6 -c pytorch -c nvidia" command installs pytorch 1.12.1 but with cuda 11.3 instead of 11.6.


userbenchmark/release-test/setup_env.sh
```diff
@@ -19,13 +19,18 @@ if [ -e "/usr/local/cuda" ]; then
...
 conda uninstall -y pytorch torchvision torchtext pytorch-cuda
+conda uninstall -y pytorch torchvision torchtext cudatoolkit
 # make sure we have a clean environment without pytorch
 pip uninstall -y torch torchvision
 
 # install magma
 conda install -y -c pytorch ${MAGMA_VERSION}
 # install pytorch and pytorch-cuda
-conda install --force-reinstall pytorch=${PYTORCH_VERSION} torchvision torchtext pytorch-cuda=${CUDA_VERSION} -c ${PYTORCH_CHANNEL} -c nvidia
+if [ $PYTORCH_CHANNEL = "pytorch" ]; then
```
xuzhao9, line 29: Is this still true for the future pytorch releases? E.g. in the future 1.14.0, we should install `cudatoolkit` for pytorch channel, and use `pytorch-cuda` for pytorch-test channel?

xuzhao9 (approved):
Approved with minor inline comments

userbenchmark/release-test/setup_env.sh
weiwangmeta: Good point. For 1.14.0 testing, running 1.13.0 from pytorch channel would need pytorch-cuda instead of cudatoolkit... 
So, this code will very likely break. Let me continue working on this to see how these testing and conda installation can be unified.

weiwangmeta: Ended up hard-coding 1.12.1. Please let me know if you think of better ways.

xuzhao9: I am okay with hard-coding as long as it won't break on future 1.14.0.

weiwangmeta: Ok, thanks. It should not break, but let's see when 1.14.0 comes. Merging it now.

PR #1261, opened 2022-10-26
Migrate to enable Python 3.10

xuzhao9:
To support Python 3.10, we need to update a few model dependencies such as fairseq and spacy.

We need to add torchaudio dependency because newer version of fairseq now depends on torchaudio: https://github.com/facebookresearch/fairseq/blob/main/setup.py#L190
Torchtext also adds new dependency torchdata so we need to include that as well: https://github.com/pytorch/text/pull/1961

Currently, the CircleCI test runs on Python 3.8, and the GHA test runs on Python 3.10.


torchbenchmark/util/model.py
```diff
@@ -169,9 +169,12 @@ def determine_batch_size(self, batch_size=None):
...
             self.batch_size = self.DEFAULT_TRAIN_BSIZE if self.test == "train" else self.DEFAULT_EVAL_BSIZE
             # use the device suggestion on CUDA inference tests
-            if self.test == "eval" and self.device == "cuda":
-                current_device_name = torch.cuda.get_device_name()
-                assert current_device_name, f"torch.cuda.get_device_name() returns None when device is set to cuda, please double check."
+            if self.test == "eval":
+                if self.device == "cuda":
+                    current_device_name = torch.cuda.get_device_name()
+                    assert current_device_name, f"torch.cuda.get_device_name() returns None when device is set to cuda, please double check."
+                elif self.device == "cpu":
```
desertfire, line 176: Curious what is this change for?

desertfire (approved):
Awesome! Thank you so much for getting this done so quickly!

desertfire (approved):
Awesome! Thank you so much for getting this done so quickly!

torchbenchmark/util/model.py
xuzhao9: We now support specifying a smaller batch size for the CPU device. In the code, we try to use the same batch size as upstream. However, upstream batch sizes are often optimized for GPU, for CPU+inference tests, we want to use a smaller batch size by default to save test time.

Issue #1271, opened 2022-10-31
soft_actor_critic does not use the new gym API and crash on env initialization


Delaunay:
Gym deprecated the method `env.seed` and needs to use `env.reset(seed=seed)` instead

The method `end.seed` was removed in https://github.com/openai/gym/commit/3a8daafce113c6161190c0252c6a3a340cc38fd4#diff-8f3ae46be618d936b191b5f32bad0dcc4674ea5385bafb31a6c02840a43fe76aL196

xuzhao9:
Fixed in https://github.com/pytorch/benchmark/pull/1272

PR #1272, opened 2022-10-31
replace env.seed by end.reset(seed=seed)  #1271

Delaunay:


xuzhao9 (approved):
LGTM. Thanks for fixing.

PR #1278, opened 2022-11-03
Parse opt_args even when dynamo is used

davidberard98:
* __->__ #1278

Previously we had two versions of opt_args - one generated by dynamo and
one generated by the non-dynamo path. If we add this change, then we'll
be able to add arguments into opt_args that we would like to use in both
dynamo and non-dynamo contexts

Differential Revision: [D41008361](https://our.internmc.facebook.com/intern/diff/D41008361)


davidberard98:
@davidberard98 has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D41008361).

davidberard98:
Opt_args is intended for backend-specific args, so any args that might be expected for all backends should go in dargs.

PR #1281, opened 2022-11-04
[ddp+dynamo] use cpus_per_task=12 instead of 10

davidberard98:
* __->__ #1281

AWS cluster has machines with 96 cpus and 8 gpus, so 12 cpus per task
means we use all the CPUs.

I suspect that we were running into issues with bad CPU affinity when we
had 10 cpus_per_task, because often the 4th rank would be slower than
the other ranks.

Differential Revision: [D41016736](https://our.internmc.facebook.com/intern/diff/D41016736)

### DIFF
--- a/userbenchmark/ddp_experiments/__init__.py
+++ b/userbenchmark/ddp_experiments/__init__.py
@@ -321,7 +321,7 @@ def main():
         gpus_per_node=args.ngpus,
         # one task per GPU
         tasks_per_node=args.ngpus,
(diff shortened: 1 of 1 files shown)


davidberard98:
@davidberard98 has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D41016736).

wconstab (approved):

xuzhao9 (approved):

PR #1294, opened 2022-11-09
[ddp+dynamo] Add distributed correctness checks

davidberard98:
* #1304
* __->__ #1294

DDP+Dynamo experiments run in different processes, and DDP also occurs after the general correctness check - so they need to be handled separately.

Process:
1) When setting up the measurements in `ddp_experiments/__init__.py`, categorize some measurements as "reference" (e.g. the eager measurements) and others as "test". Also assign a file path.
2) For reference measurements, do an initial correctness measurement and dump the results into a file. For test measurements, load the reference measurements from the file and compare the initial correctness measurements.
3) Make sure to run the correctness measurement on all ranks, even if we're only doing the correctness check on rank 0.
4) Make sure to disable non-distributed correctness checks to avoid an additional iteration that might affect dynamo and eager parameters differently.

Currently hf_T5, hf_T5_large, hf_Bert, hf_GPT2_large, and timm_vision_transformer are passing; ~resnet50 is failing. Still investigating the resnet50 issue.~ resnet50 is also failing correctness with `python run.py resnet50 -t train -d cuda --torchdynamo inductor`, so this isn't a DDP-specific problem.

Usage:
```
python userbenchmark/ddp_experiments/__init__.py --job_dir /fsx/path/to/dir/shared/across/cluster --check_correctness_distributed
```

i.e. adding --check_correctness_distributed will add correctness checks. Note that the correctness checks here shouldn't be used if you care about performance; we modify the models with model.eval() to get rid of dropouts, so it's not representative of actual performance. Since we probably don't care about performance anyway when using this option, we also reduce the number of iterations to speed up the test time.

Differential Revision: [D41312110](https://our.internmc.facebook.com/intern/diff/D41312110)

### DIFF
--- a/torchbenchmark/util/model.py
+++ b/torchbenchmark/util/model.py
@@ -123,7 +123,6 @@ class BenchmarkModel(metaclass=PostInitProcessor):
     def __post__init__(self):
         should_check_correctness = check_correctness_p(self, self.opt_args, self.dargs)
         if should_check_correctness:
-            print("CORRECTNESS CHECK dberard")
             self.eager_output = stableness_check(self, cos_sim=False, deepcopy=self.DEEPCOPY, rounds=1)
             if isinstance(self.eager_output, Tuple):
                 self.eager_output = tuple((t.detach() if isinstance(t, torch.Tensor) else t) for t in self.eager_output)
--- a/torchbenchmark/util/model.py
+++ b/torchbenchmark/util/model.py
@@ -91,6 +91,20 @@ class BenchmarkModel(metaclass=PostInitProcessor):
         self.run_contexts = [
             enable_profiling_executor  # force JIT profiling executor to be enabled by default
         ]
+
+        # taken from torchdynamo benchmarks, this further controls randomness settings
+        def deterministic_torch_manual_seed(*args, **kwargs):
+            from torch._C import default_generator
+
+            seed = 1337
+            import torch.cuda
+
+            if not torch.cuda._is_in_bad_fork():
+                torch.cuda.manual_seed_all(seed)
+
+            return default_generator.manual_seed(seed)
+
+        torch.manual_seed = deterministic_torch_manual_seed
         set_random_seed()
         # sanity checks of the options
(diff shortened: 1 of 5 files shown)


torchbenchmark/util/model.py
```diff
@@ -91,6 +91,20 @@ def __init__(self, test: str, device: str, jit: bool=False, batch_size: Optional
...
         ]
+
+        # taken from torchdynamo benchmarks, this further controls randomness settings
+        def deterministic_torch_manual_seed(*args, **kwargs):
+            from torch._C import default_generator
+
+            seed = 1337
+            import torch.cuda
+
+            if not torch.cuda._is_in_bad_fork():
```
wconstab, line 102: what's this- any idea if you're in it?

wconstab: for my test i used 

```
def reset_rng_state():
    torch.manual_seed(1337)
    random.seed(1337)
    np.random.seed(1337)
```

any idea if you're using np or random libs in model init and therefore missing those?

wconstab: one debugging technique could be to add a function that pulls a random  number from torch and prints it; then call this function from various points in your benchmark (same points in both ref and test).  you could start at coarse grain, e.g. put these  calls at various places in the torchbench-distributed-ddp-experiment code, before/after calling the model etc.  then you could even experiment with adding it as a hook after each layer in the model or something like that.

davidberard98: hmm, I should take a look at this - it could be the problem, seems like it has something to do with forking processes. https://github.com/pytorch/pytorch/blob/master/torch/cuda/__init__.py#L228-L230.

re: reset_rng_state(), I think this is the same thing that set_random_seed() does: https://github.com/pytorch/benchmark/blob/main/torchbenchmark/util/env_check.py#L16-L22

wconstab (approved):

davidberard98:
@davidberard98 has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D41312110).

PR #1302, opened 2022-11-13
Enable E2E hf_bert dynamo+inductor train/eval

yanboliang:


yanboliang:
This PR should works well after https://github.com/pytorch/pytorch/pull/88963.

torchbenchmark/e2e_models/hf_bert/__init__.py
```diff
@@ -1,5 +1,6 @@
 from accelerate.utils.dataclasses import DeepSpeedPlugin
 import torch
+import torch._dynamo
```
xuzhao9, line 3: Let's wrap this in a `try` block, since other users might test this code on stable/old pytorch releases, where `torch._dynamo` is not available.

xuzhao9 (approved):
LGTM

PR #1306, opened 2022-11-15
Add --profile-options to support fine toggling of profile options

aaronenyeshi:
Summary: Support a new flag, --profile-options, to enable fine grained toggling of profiler options, such as record_shapes, profile_memory, with_stack, with_flops, and with_modules. This allows the user to pick which options to enable rather than using --profile-detailed which enables all options.

Reviewed By: xuzhao9

Differential Revision: D41314079

Pulled By: aaronenyeshi


xuzhao9 (approved):
Thanks!

PR #1308, opened 2022-11-15
Dynamo option to disable() optimizer.step

davidberard98:
* __->__ #1308

Previously was working around this with `torch._dynamo.disable(self.optimizer.step)()` in the model_factory.py code (e.g. [this branch](https://github.com/pytorch/benchmark/compare/main...davidberard98/skip-dynamo-optimizer)), but we probably don't want to land that. I think this does the same thing, but guarded with a flag.

Test:
```
ADAM_CAPTURABLE=1 python run.py hf_T5_large -t train -d cuda --torchdynamo inductor --torchinductor_cudagraph False --dynamo_disable_optimizer_step True
```

^ takes ~5 minutes, which is a lot slower than the ~30min expected when the optimizer gets compiled.

Differential Revision: [D41326442](https://our.internmc.facebook.com/intern/diff/D41326442)

### DIFF
--- a/torchbenchmark/util/backends/torchdynamo.py
+++ b/torchbenchmark/util/backends/torchdynamo.py
@@ -64,13 +64,13 @@ def apply_torchdynamo_args(model: 'torchbenchmark.util.model.BenchmarkModel', ar
     if bool(args.dynamo_disable_optimizer_step):
         found_optimizer_step = False
         try:
-            torch._dynamo.disallow_in_graph(model.cfg.optimizer.step)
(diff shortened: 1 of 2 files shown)


davidberard98:
@davidberard98 has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D41326442).

wconstab (approved):
this seems ok to me as a short term thing.  @albanD is working on a better plan for optimizers.

PR #1316, opened 2022-11-19
README.md fix

FlagZhao:
The torchaudio is missing when create new env and install.

Error message:
checking packages torch, torchvision, torchtext, torchaudio are installed...FAIL
Error: Users must first manually install packages ['torch', 'torchvision', 'torchtext', 'torchaudio'] before installing the benchmark.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ conda install -y -c pytorch magma-cuda116
 
 Then install pytorch, torchtext, and torchvision using conda:
 ```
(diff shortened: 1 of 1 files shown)


xuzhao9:
Thanks for fixing this! Yes now we require torchaudio because recent fairseq versions now require torchaudio: https://github.com/facebookresearch/fairseq/blob/main/setup.py#LL190C19-L190C24

xuzhao9 (approved):

PR #1317, opened 2022-11-20
Add --amp option to torchbench run.py

jansel:
Differential Revision:
D41429575

LaMa Project: L1124100


xuzhao9 (approved):

xuzhao9:
I want to note that this might not work for train tests, because in the train loop, amp should only apply to the forward pass, not backward or optimizer.

PR #1318, opened 2022-11-21
Add FSDP to ddp/dynamo benchmarks

davidberard98:
* #1322
* #1321
* __->__ #1318

- Make apply_trainer configurable (by providing your own implementation). This is needed because you might want to benchmark different configurations of FSDP, or configure differently per model. For example, see the T5_large or timm_vision_transformer FSDP wrapping functions in this PR.
- Make the ddp script work for DDP _and_ FSDP
- Add T5 and timm_vision_transformer_large for FSDP
- Monitor max memory in the distributed trainer

Differential Revision: [D41475962](https://our.internmc.facebook.com/intern/diff/D41475962)


davidberard98:
@davidberard98 has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D41475962).

davidberard98:
@davidberard98 has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D41475962).

torchbenchmark/util/distributed/core_model/trainer.py
```diff
@@ -93,7 +93,10 @@ def measure(self):
                 grad_params = {}
                 for name, param in self.benchmark.model.named_parameters():
                     if param.requires_grad:
-                        grad_params[name + ".grad"] = param.grad.cpu()
+                        if param.grad is not None:
+                            grad_params[name + ".grad"] = param.grad.cpu()
```
wconstab, line 97: is this happening in fsdp bc parameters are the views i hacked?

if so, does this mean params are gonna bypass accuracy check in fsdp bc of this?

userbenchmark/ddp_experiments/__init__.py
```diff
@@ -489,6 +469,222 @@ def get_backend_name(model_args):
     print(job.results())
 
 
+def apply_fsdp_hf_T5_large(model, trainer):
```
wconstab, line 472: would it be better to make the majority of this code live in a common 'apply_fsdp' helper, and just make the wrap_policy be determined by the model class or the args string or something?

userbenchmark/ddp_experiments/__init__.py
```diff
@@ -489,6 +469,222 @@ def get_backend_name(model_args):
...
+        use_orig_params=True,
+    )
+    print(fsdp_model, "local_rank", local_rank)
+    return fsdp_model
+
+def benchmark_fsdp(args, executor):
+    def get_backend_name(model_args):
+        if "--torchdynamo" in model_args:
+            return "torchdynamo_" + model_args[model_args.index("--torchdynamo") + 1]
+        return "eager"
```
wconstab, line 524: does this confuse between the case of `--torchdynamo eager` (using eager backend for dynamo) and actual eager?

torchbenchmark/util/distributed/core_model/trainer.py
davidberard98: good point, let me take a closer look into why this is happening.

userbenchmark/ddp_experiments/__init__.py
davidberard98: IMO it makes sense to keep this code here - a lot of the configuration here is specific to dynamo, e.g. use_orig_params=True. Does that seem reasonable or do you still think we should move this into some common folder?

userbenchmark/ddp_experiments/__init__.py
```diff
@@ -489,6 +469,222 @@ def get_backend_name(model_args):
...
+
+    experiments = []
+    for i in range(args.repeat):
+        for nodes in node_list:
+            for model_name, config_generator in model_configs.items():
+                for model_args in model_args_configs:
+                    experiments.append(config_generator(nodes, model_args))
+
+    allocation_nodes = max(node_list)
+    executor.update_parameters(
```
wconstab, line 632: does this need to be duplicated from below?

wconstab (approved):

torchbenchmark/util/distributed/core_model/trainer.py
davidberard98: I verified that this is happening in both eager and dynamo+inductor; i.e. all the params for which `param.grad == None`

cc @awgu do you have any idea why this would be? i.e. for some parameters, `param.requires_grad == True` but `param.grad is None`. This doesn't happen in DDP for these models.

userbenchmark/ddp_experiments/__init__.py
wconstab: i was less talking about the folder its in and more the duplication.

I think the only thin you need to customize per model is the wrap policy.  (hopefully).  

use_orig_params=True is going to be a broad requirement for any dynamo FSDP, (and i think @awgu said it will also become the new default at some point)

torchbenchmark/util/distributed/core_model/trainer.py
wconstab: this is happening bc of `use_orig_params` being set to true.  i think.  

my fsdp enablement PR puts orig_params views into module .parameters field, so  at least aot_autograd can find them and wrap them in faketensors.  But as they are views, I am not sure if their requires_grad field is set to true. or if that is only true on the underlying flat buffer...

awgu: For any original parameter `param` that is not in the rank's shard of the parameters, `param.grad = None`. (Otherwise, if `param` has at least some presence in the rank's shard, then `param.grad` is the 1D flattened shard of the gradient.) However, `param.requires_grad == True` because it preserved the original `requires_grad` and makes sense because the parameter logically should receive a gradient (ignoring the sharding).

Is this a problem/blocker?

awgu: Actually, I wonder if we can set `param.requires_grad = False` because the unsharded parameter is actually a different Python variable (that is a plain `Tensor`). Either way, let me know if these parameters having `requires_grad == True` but having `.grad = None` is a problem/blocker.

davidberard98:
@davidberard98 has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D41475962).

userbenchmark/ddp_experiments/__init__.py
davidberard98: de-duped a bunch of this

userbenchmark/ddp_experiments/__init__.py
davidberard98: sounds good, I de-duped some of this

torchbenchmark/util/distributed/core_model/trainer.py
davidberard98: @awgu not a problem - but will this strategy (iterate through parameters and check the ones with `param.requires_grad and param.grad is not None`) find all the gradients? or are there other gradients that I need to check that won't be caught in this check?

awgu:
I read over `trainer.py`, and it looks good to me. I will let @lessw2020 give his opinion on the benchmarking setup since he has more expertise.

davidberard98:
@davidberard98 has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D41475962).

Issue #1331, opened 2022-11-30
errors for configured CUDA_VISIBLE_DEVICES on multiple gpus


FindHao:
If we specify the visibility of a subset of installed GPUs via CUDA_VISIBLE_DEVICES, there will be an error 
```
components.model_analyzer.tb_dcgm_types.da_exceptions.TorchBenchAnalyzerException:  GPU with GPU-XXXX uuid is not present!
```

PR #1332, opened 2022-11-30
fix bug for multigpus when only a subset is visible

FindHao:
When there are multiple GPUs and only a subset is visible, e.g. specified the visible GPUs by `CUDA_VISIBLE_DEVICES`, to the process, errors occur as described in https://github.com/pytorch/benchmark/issues/1331

fixes  https://github.com/pytorch/benchmark/issues/1331

### DIFF
--- a/components/model_analyzer/tb_dcgm_types/da_exceptions.py
+++ b/components/model_analyzer/tb_dcgm_types/da_exceptions.py
@@ -1,6 +1,16 @@
 class TorchBenchAnalyzerException(Exception):
     """
-    A custom exception specific to the Triton Model
-    Analyzer
+    A custom exception specific to the TorchBench Model Analyzer
     """
     pass
+
+
+class TorchBenchAnalyzerException_GPU_Inavailable(Exception):
+    """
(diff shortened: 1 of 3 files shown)


components/model_analyzer/tb_dcgm_types/da_exceptions.py
```diff
@@ -1,6 +1,16 @@
...
-    A custom exception specific to the Triton Model
-    Analyzer
+    A custom exception specific to the TorchBench Model Analyzer
     """
-    pass
\ No newline at end of file
+    pass
+
+
+class TorchBenchAnalyzerException_GPU_Inavailable(Exception):
```
xuzhao9, line 8: Should we rename this to `TorchBenchAnalyzerException_GPU_Unavailable`

FindHao: fixed by https://github.com/pytorch/benchmark/pull/1332/commits/9b74dd72b6c3cea086a6486018044d923cb9bf65

xuzhao9 (approved):

PR #1334, opened 2022-12-06
add dcgm standalone mode

FindHao:
dcgm has two modes:
- standalone mode: our tool needs to contact nv-hostengine
- embedded mode: our tool embeds the dcgm library and directly contacts the GPU driver .

These two modes can not exist at the same time which means if someone is using standalone mode and our tool will raise an error `the third-party profiling module returned an unrecoverable error`.


FindHao:
related discussion https://github.com/NVIDIA/DCGM/issues/56

PR #1338, opened 2022-12-07
[WIP] prelim script to run some models

janeyx99:
This does NOT return any metrics yet in the JSON, but it does print the benchmark results! Just wanted to save progress for now.

For example: adam bm results
image

### DIFF
--- a/userbenchmark/optim/__init__.py
+++ b/userbenchmark/optim/__init__.py
@@ -1,8 +1,12 @@
 from typing import Any, Dict, List, Tuple
 from torchbenchmark import load_model_by_name
 import torch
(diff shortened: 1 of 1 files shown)


xuzhao9:
I created https://github.com/pytorch/benchmark/pull/1339 which introduces a few other interfaces  in the module `torchbenchmark.util.experiment`. Hope those can make developing userbenchmarks easier

janeyx99:
Today (March 8, 2023) I have decided to switch my approach with more benchmarks. I am going to close this PR to preserve this attempt (in case I want anything from here in the future) and open a new one for a mental refresh.

PR #1343, opened 2022-12-13
Set cudnn.benchmark in timm and vision models to True by default

FindHao:
The upstream of [timm framework](https://github.com/rwightman/pytorch-image-models/blob/e7da205345dcf770ee4bedd62d06fad7a1458904/train.py#L372) has set `torch.backends.cudnn.benchmark` to True by default. It's better to keep it the same as the upstream.

The vision framework doesn't have an upstream repo, and it is supposed to use fixed input sizes each time in TorchBench. So it's better to enable cudnn.benchmark too.

### DIFF
--- a/torchbenchmark/models/alexnet/metadata.yaml
+++ b/torchbenchmark/models/alexnet/metadata.yaml
@@ -1,8 +1,8 @@
 devices:
   NVIDIA A100-SXM4-40GB:
     eval_batch_size: 1024
-eval_benchmark: false
+eval_benchmark: true
 eval_deterministic: false
 eval_nograd: true
-train_benchmark: false
+train_benchmark: true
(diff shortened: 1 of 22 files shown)


xuzhao9:
You also need to update the corresponding `metadata.yaml` values.

FindHao:
> You also need to update the corresponding `metadata.yaml` values, e.g.: https://github.com/pytorch/benchmark/blob/main/torchbenchmark/models/timm_resnest/metadata.yaml#L7

fixed by 256326ab4c1f368de6ea15fe7b24e6f650cb852d

xuzhao9 (approved):
LGTM!

davidberard98 (approved):

PR #1365, opened 2023-01-11
Update README.md

johnnynunez:
Update URL's
Update to cuda 11.7 by default

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,19 +24,19 @@ conda activate torchbenchmark
 conda install -y git-lfs
 ```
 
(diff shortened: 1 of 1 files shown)


xuzhao9:
Thanks for your PR. However, we can't accept it for now because the default CUDA version in README.md needs to be consistent with our cuda_utils.py (https://github.com/pytorch/benchmark/blob/main/utils/cuda_utils.py) and PR/nightly workflows: https://github.com/pytorch/benchmark/blob/main/.github/workflows/pr-gha-runner.yml#L41

We should probably automate it with some scripts (e.g. install.py should read it from cuda_utils.py and install the packages if not available), but we don't have time to do this right now.

PR #1367, opened 2023-01-12
drq: change view to reshape to support channels last

chunyuan-w:
The current implementation of `drq` does not support **channels last** input since `view` has constraints on the input size and stride (https://pytorch.org/docs/stable/generated/torch.Tensor.view.html):
```
Cannot view a tensor with shape torch.Size([1, 32, 35, 35]) and strides (39200, 1, 1120, 32) as a tensor with shape (1, 39200)!
```

Change `view` to `reshape` which returns a view if the shapes are compatible, and copies (equivalent to calling `contiguous()`) otherwise.

### DIFF
--- a/torchbenchmark/models/drq/drq.py
+++ b/torchbenchmark/models/drq/drq.py
@@ -43,6 +43,8 @@ class Encoder(nn.Module):
             conv = torch.relu(self.convs[i](conv))
             self.outputs['conv%s' % (i + 1)] = conv
 
+        # Changed view to reshape here to support channels last input
+        # TODO: upstream this change to https://github.com/denisyarats/drq/blob/master/drq.py#L48
         h = conv.reshape(conv.size(0), -1)
         return h
 
--- a/torchbenchmark/models/drq/drq.py
+++ b/torchbenchmark/models/drq/drq.py
@@ -43,7 +43,7 @@ class Encoder(nn.Module):
             conv = torch.relu(self.convs[i](conv))
             self.outputs['conv%s' % (i + 1)] = conv
 
-        h = conv.view(conv.size(0), -1)
+        h = conv.reshape(conv.size(0), -1)
         return h
 
     def forward(self, obs, detach=False):


chunyuan-w:
@xuzhao9 could you please help review this PR? Thanks!

chunyuan-w:
cc @jgong5

jgong5 (approved):

xuzhao9 (approved):
LGTM, thanks for the improvement.
Just want to note that we should also add a comment in the code saying this is an update that doesn't exist in the upstream code: https://github.com/denisyarats/drq/blob/master/drq.py#L48

torchbenchmark/models/drq/drq.py
```diff
@@ -43,7 +43,7 @@ def forward_conv(self, obs):
             conv = torch.relu(self.convs[i](conv))
             self.outputs['conv%s' % (i + 1)] = conv
 
-        h = conv.view(conv.size(0), -1)
+        h = conv.reshape(conv.size(0), -1)
```
xuzhao9, line 46: Can you add one-line comment mention that this is an improvement that doesn't exist in the upstream code (with link to the upstream code at https://github.com/denisyarats/drq/blob/master/drq.py#L48)?

jgong5: @xuzhao9 We can submit a PR to upstream too. Does it sound good to you?

xuzhao9: > @xuzhao9 We can submit a PR to upstream too. Does it sound good to you?

Sure, if you can submit a PR to upstream and get accepted, we don't need the comment line.

chunyuan-w: I've added one-line comment as suggested for now.
I'll submit a PR to upstream this. If it is accepted, let me submit a following-up PR later to this benchmark repo to remove the one-line comment. Is it okay?

xuzhao9: Yes, that shoulds perfect to me!

chunyuan-w: FYI, I've submitted a PR to the upstream drq repo: https://github.com/denisyarats/drq/pull/27.

chunyuan-w:
For the CI failure, it seems unrelated with my changes:
image

And it also marks that there are 2 workflows awaiting approval:
image

@xuzhao9 could you please help on that?

xuzhao9:
> For the CI failure, it seems unrelated with my changes: image
> 
> And it also marks that there are 2 workflows awaiting approval: image
> 
> @xuzhao9 could you please help on that?

There is a pre-existing CI failure not because of this PR. Sorry but we are still working on it.
I manually tested this PR should be safe, we can merge it despite the CI problem.

Issue #1354, opened 2022-12-28
test resnet50 failed?


zkh2016:
Hi，I encountered a problem when executing the following test command :
```
 python run.py resnet50 -d cuda -t train --bs 32 --torchdynamo inductor
```
The result is:
```
...
[2022-12-28 06:50:30,069] torch._inductor.ir: [WARNING] Using FallbackKernel: torch.ops.aten.addcmul_.default
[2022-12-28 06:50:30,087] torch._inductor.ir: [WARNING] Using FallbackKernel: torch.ops.aten.addcmul_.default
[2022-12-28 06:50:42,387] torch._inductor.compile_fx: [WARNING] skipping cudagraphs due to multiple devices
grad of param conv1.weight after running with dynamo doesn't have gradient matching with eager mode
Running train method from resnet50 on cuda in eager mode with input batch size 32.
GPU Time:             38.122 milliseconds
CPU Total Wall Time:  38.148 milliseconds
Correctness:                        False
```
It shows that the correctness is false. Is it abnormal?

env: [pytorch2.0](https://pytorch.org/get-started/pytorch-2.0/#getting-started) , A100, cuda11.7

xuzhao9:
I can reproduce it on the latest PyTorch nightly and A100:

```
Running train method from resnet50 on cuda in eager mode with input batch size 32.
GPU Time:             37.284 milliseconds
CPU Total Wall Time:  37.334 milliseconds
Correctness:                        False
```
version: torch-2.0.0.dev20221222+cu116

@desertfire Can you help look into this and see if you can reproduce?

desertfire:
I can reproduce it locally, but running resnet50 through the script in PyTorch core does pass (`python benchmarks/dynamo/torchbench.py --training --accuracy --device cuda --inductor  --only resnet50`). So I suspect there is something inconsistent between two scripts.

xuzhao9:
Fixed by https://github.com/pytorch/benchmark/pull/1700

PR #1382, opened 2023-01-25
Use fp64 result for correctness checking in TorchBench

desertfire:
Summary: Compare model outputs with the fp64 eager result. The idea is to use fp64 as the golden baseline and compare the difference between eager_model/fp64_model and optimized_model/fp64_model. This also makes the comparison consistent with the approach used by PT2.

Differential Revision: D42637398

### DIFF
--- a/torchbenchmark/util/env_check.py
+++ b/torchbenchmark/util/env_check.py
@@ -7,6 +7,9 @@ import copy
 import warnings
 from typing import List, Dict, Tuple, Optional
 
+import torch
+from torch._dynamo.utils import same, clone_inputs
+
 MAIN_RANDOM_SEED = 1337
 # rounds for stableness tests
 STABLENESS_CHECK_ROUNDS: int = 3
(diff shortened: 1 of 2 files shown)


xuzhao9:
Does this PR fix https://github.com/pytorch/benchmark/issues/1354, or it is still unknown why there's inconsistency in the accuracy result?

desertfire:
> Does this PR fix #1354, or it is still unknown why there's inconsistency in the accuracy result on resnet50?

I just gave it a try, and the benchmark does pass now, although it hit `fp64 golden ref were not generated. Setting accuracy check to cosine`. This PR gets us one step closer to being consistent with the checking in PyTorch core. We can keep addressing those issues if there are more inconsistency coming up.

xuzhao9:
> > Does this PR fix #1354, or it is still unknown why there's inconsistency in the accuracy result on resnet50?
> 
> I just gave it a try, and the benchmark does pass now, although it hit `fp64 golden ref were not generated. Setting accuracy check to cosine`. This PR gets us one step closer to being consistent with the checking in PyTorch core. We can keep addressing those issues if there are more inconsistency coming up.

Curious does this problem also happen in PyTorch core torchbench.py?

xuzhao9:
https://github.com/pytorch/benchmark/pull/1700 should address this

PR #1384, opened 2023-01-28
Update the backend interface

xuzhao9:
This PR updates the flow of argument parsing. It parses all the backend arguments before initializing the model.

This makes model developers no need to worry about arguments in the backend.
Later, we will move fuser and tensorrt backends to the same format.

### DIFF
--- a/torchbenchmark/util/backends/__init__.py
+++ b/torchbenchmark/util/backends/__init__.py
@@ -12,7 +12,7 @@ def create_backend(fn):
             return None
 
         try:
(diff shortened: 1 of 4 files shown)


xuzhao9:
@zzpmiracle I want to make sure this won't break your logic to add BladeDISC backends.

zzpmiracle:
> @zzpmiracle I want to make sure this won't break your logic to add BladeDISC backends.

Little influence, take it free.

PR #1392, opened 2023-02-01
enable peak memory measurement by default

FindHao:
The CPU and GPU peak memory measurements will be enabled by default.
The output is like the following:
```
$ python3 run.py -d cuda -t train   BERT_pytorch 
Running train method from BERT_pytorch on cuda in eager mode with input batch size 16.
GPU Time:            107.128 milliseconds
CPU Total Wall Time: 107.156 milliseconds
GPU 0 Peak Memory:                6.2205 GB
CPU Peak Memory:                2.8018 GB
```
The dependent package could be installed by the following command.
```
pip install pynvml
```

### DIFF
--- a/components/model_analyzer/TorchBenchAnalyzer.py
+++ b/components/model_analyzer/TorchBenchAnalyzer.py
@@ -115,8 +115,10 @@ class ModelAnalyzer:
         try:
             self.start_monitor_timestamp = time_ns()
             if self.gpu_metrics:
-                self.gpu_factory = GPUDeviceFactory()
+                self.gpu_factory = GPUDeviceFactory(self.gpu_monitor_backend)
                 self.gpus = self.gpu_factory.verify_requested_gpus(['all', ])
+                if not self.gpus:
+                    raise TorchBenchAnalyzerException('No GPU found')
                 if self.gpu_monitor_backend == 'dcgm':
                     self.gpu_monitor = DCGMMonitor(
                         self.gpus, self.config.monitoring_interval, self.gpu_metrics)
--- a/components/model_analyzer/TorchBenchAnalyzer.py
+++ b/components/model_analyzer/TorchBenchAnalyzer.py
@@ -30,7 +30,7 @@ from time import time_ns
 
 
 class ModelAnalyzer:
-    def __init__(self):
+    def __init__(self, export_metrics_file='', metrics_needed=[], metrics_gpu_backend='dcgm'):
         # For debug
         # set_logger(logging.DEBUG)
         set_logger()
@@ -56,6 +56,7 @@ class ModelAnalyzer:
         self.config = AnalayzerConfig()
         self.gpu_record_aggregator = RecordAggregator()
         self.export_csv_name = ''
+        self.set_export_csv_name(export_metrics_file)
         # the cpu metrics to be collected. available metrics are [CPUPeakMemory, ]
         self.cpu_metrics = []
         self.cpu_monitor = None
@@ -65,7 +66,25 @@ class ModelAnalyzer:
         self.cpu_metric_value = {}
         # GPU Monitor Backend
         self.gpu_monitor_backend = 'dcgm'
+        self.start_monitor_timestamp = None
         self.stop_monitor_timestamp = None
+        self.metrics_backend_mapping = {}
+        self.process_metrics(metrics_needed, metrics_gpu_backend)
+    
+    def process_metrics(self, metrics_needed, metrics_gpu_backend):
+        if 'gpu_peak_mem' in metrics_needed:
+            self.add_metric_gpu_peak_mem()
+            self.metrics_backend_mapping['gpu_peak_mem'] = 'dcgm' if metrics_gpu_backend == 'dcgm' else 'nvml'
+        if 'flops' in metrics_needed:
+            if metrics_gpu_backend == 'dcgm':
+                self.add_metric_gpu_flops()
+                self.metrics_backend_mapping['flops'] = 'dcgm'
+            else:
+                self.metrics_backend_mapping['flops'] = 'fvcore'
+        if 'cpu_peak_mem' in metrics_needed:
+            self.add_metric_cpu_peak_mem()
+        if metrics_gpu_backend == "default":
+            self.set_gpu_monitor_backend_nvml()
 
     def add_metric_gpu_peak_mem(self):
         self.gpu_metrics.append(GPUPeakMemory)
@@ -80,13 +99,21 @@ class ModelAnalyzer:
         self.gpu_monitor_backend = 'nvml'
 
     def set_export_csv_name(self, export_csv_name=''):
+        if export_csv_name == '':
+            return
         self.export_csv_name = export_csv_name
         # test for correct permission
         with open(export_csv_name, 'w') as fout:
             fout.write('')
 
+    def update_export_name(self, insert_str=''):
+        index = self.export_csv_name.find('.csv')
+        if index != -1:
+            self.export_csv_name = self.export_csv_name[:index] + insert_str + self.export_csv_name[index:]
+
     def start_monitor(self):
         try:
+            self.start_monitor_timestamp = time_ns()
             if self.gpu_metrics:
                 self.gpu_factory = GPUDeviceFactory()
                 self.gpus = self.gpu_factory.verify_requested_gpus(['all', ])
@@ -221,7 +248,7 @@ class ModelAnalyzer:
                 fout.write('\n')
                 last_timestamp = timestamp_start
                 for a_timestamp in timestamps:
-                    duration = (a_timestamp - last_timestamp) / 1000.0
+                    duration = (a_timestamp - last_timestamp) / 1e3
                     last_timestamp = a_timestamp
                     line = "%.2f, " % ((a_timestamp - timestamp_start) / 1000)
                     for record_type in csv_records[gpu_uuid]:
@@ -263,7 +290,6 @@ class ModelAnalyzer:
         """
         if gpu_uuid:
             if gpu_uuid in self.gpu_metric_value:
-                gpu = self.gpu_factory.get_device_by_uuid(gpu_uuid)
                 return self.gpu_metric_value[gpu_uuid][GPUPeakMemory].value() / 1024
             else:
                 raise TorchBenchAnalyzerException("No available GPU with uuid ", gpu_uuid, " found!")
@@ -283,6 +309,9 @@ class ModelAnalyzer:
             logger.warning("There are multiple available CPUs and will only return the first one's peak memory bandwidth.")
         cpu_uuid = next(iter(self.cpu_metric_value))
         return self.cpu_metric_value[cpu_uuid][CPUPeakMemory].value() / 1024
+    
+
+
 
 
 def check_dcgm():
--- a/components/model_analyzer/TorchBenchAnalyzer.py
+++ b/components/model_analyzer/TorchBenchAnalyzer.py
@@ -279,11 +279,11 @@ class ModelAnalyzer:
             else:
                 raise TorchBenchAnalyzerException("No available GPU with uuid ", gpu_uuid, " found!")
         else:
-            if len(self.gpu_metric_value) > 1:
-                logger.warning("There are multiple available GPUs and will only return the first one's flops.")
+            # Will only return the first one's peak memory bandwidth. So please use CUDA_VISIBLE_DEVICES to specify the GPU.
             gpu_uuid = next(iter(self.gpu_metric_value))
             gpu = self.gpu_factory.get_device_by_uuid(gpu_uuid)
-            return gpu._sm_count * gpu._fma_count * 2 * gpu._frequency * self.gpu_metric_value[gpu_uuid][GPUFP32Active].value() / 1e+9
+            device_id = self.gpu_factory.get_device_by_uuid(gpu_uuid).device_id()
+            return device_id, gpu._sm_count * gpu._fma_count * 2 * gpu._frequency * self.gpu_metric_value[gpu_uuid][GPUFP32Active].value() / 1e+9
 
     def calculate_gpu_peak_mem(self, gpu_uuid=None) -> float:
         """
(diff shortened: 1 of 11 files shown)


xuzhao9:
Let's add `nvidia-ml-py` to requirements.txt in this case.

xuzhao9:
Also, let's remove the logs 
```
Metric gpu_peak_mem is collected by nvml backend
Metric cpu_peak_mem is collected by psutil.Process.
```

xuzhao9:
@FindHao One question: can we measure the memory usage in different runs than latency measurement loop? This makes sure memory and latency measurements are independent.

xuzhao9:
> Do you think it is a good idea to use another [while loop](https://github.com/pytorch/benchmark/pull/1392/files#diff-d6af0459a37d985953d7040c14f53feb3b9cc9e58b543aa3c2b80256d276c5e0R131) to measure the memory usage?
Yes another while loop sounds good to me. Note that we don't need to warmup for memory usage.

> 
> What about TFLOPS?

TFLOPS is fine since it is not enabled by default.

FindHao:
There are two solutions to split the memory measurements.
1. use another independent model_analyzer
2. use the existing model_analyzer

The first one is easier to be implemented, but the code in run.py could be slightly ugly. It has more overhead although this overhead is not counted in our measurements. The second one needs more changes in ModelAnalyzer, but the code in run.py is neater. Which one do you think is better?

xuzhao9:
@FindHao Let's keep a neater run.py

torchbenchmark/util/experiment/metrics.py
```diff
@@ -38,19 +40,58 @@ def get_latencies(func, device: str, nwarmup=WARMUP_ROUNDS, num_iter=BENCHMARK_I
...
+    if 'gpu_peak_mem' in metrics_needed:
+        gpu_peak_mem = mem_model_analyzer.calculate_gpu_peak_mem()
+    else:
+        gpu_peak_mem = None
+    if 'cpu_peak_mem' in metrics_needed:
+        cpu_peak_mem = mem_model_analyzer.calculate_cpu_peak_mem()
+    else:
+        cpu_peak_mem = None
+    if export_metrics_file:
+        mem_model_analyzer.update_export_name("_peak_memory")
```
xuzhao9, line 69: What is this metrics file?

torchbenchmark/util/experiment/metrics.py
```diff
@@ -15,6 +15,8 @@
 @dataclasses.dataclass
 class TorchBenchModelMetrics:
     latencies: List[float]
+    cpu_peak_mem: float
```
xuzhao9, line 18: It should be Optional[float], not float.

torchbenchmark/util/experiment/metrics.py
FindHao: `--export-metrics` in `run.py` will export all records to help us debug or find more detailed information.

torchbenchmark/util/experiment/metrics.py
FindHao: fixed by https://github.com/pytorch/benchmark/pull/1392/commits/57f0649d86e35507ceb9bc1608e1f588744faa90

torchbenchmark/util/experiment/metrics.py
```diff
@@ -38,19 +40,58 @@ def get_latencies(func, device: str, nwarmup=WARMUP_ROUNDS, num_iter=BENCHMARK_I
         result_summary.append((t1 - t0) / NANOSECONDS_PER_MILLISECONDS)
     return result_summary
 
+def get_peak_memory(func, device: str, num_iter=BENCHMARK_ITERS, export_metrics_file='', metrics_needed=[], metrics_gpu_backend='dcgm') -> Tuple[float, float]:
```
xuzhao9, line 43: I guess we don't need to run 15 iterations to get the memory metrics?
Since we plan to print this metrics by default, it is better to save some time.

requirements.txt
```diff
@@ -20,3 +20,4 @@ numpy==1.21.2
 git+https://github.com/kornia/kornia.git@b7050c3
 scipy # for lazy_bench.py
 submitit
+nvidia-ml-py
```
xuzhao9, line 23: Should we make it pynvml to make it consistent with the internal version?

components/model_analyzer/TorchBenchAnalyzer.py
```diff
@@ -80,16 +99,26 @@ def set_gpu_monitor_backend_nvml(self):
         self.gpu_monitor_backend = 'nvml'
 
     def set_export_csv_name(self, export_csv_name=''):
+        if export_csv_name == '':
```
xuzhao9, line 102:
```suggestion
        if not export_csv_name:
```

components/model_analyzer/TorchBenchAnalyzer.py
```diff
@@ -80,16 +99,26 @@ def set_gpu_monitor_backend_nvml(self):
...
+            return
         self.export_csv_name = export_csv_name
         # test for correct permission
         with open(export_csv_name, 'w') as fout:
             fout.write('')
 
+    def update_export_name(self, insert_str=''):
+        index = self.export_csv_name.find('.csv')
+        if index != -1:
```
xuzhao9, line 111:
```suggestion
        if not index == -1:
```

run.py
```diff
@@ -321,6 +300,9 @@ def _validate_profile_options(profile_options: str):
     if args.amp:
         test = torch.autocast("cuda")(test)
     metrics_needed = [_ for _ in args.metrics.split(',') if _.strip()] if args.metrics else []
+    # enable cpu_peak_mem and gpu_peak_mem by default
+    metrics_needed.extend(['cpu_peak_mem', 'gpu_peak_mem'])
```
xuzhao9, line 304: Check if device is not `CUDA`, only collect `gpu_peak_mem`.

run.py
```diff
@@ -340,7 +322,7 @@ def _validate_profile_options(profile_options: str):
             exit(-1)
         export_metrics_file = "%s_all_metrics.csv" % args.model
     else:
-        export_metrics_file = False
+        export_metrics_file = ''
```
xuzhao9, line 325: Can we use `None` here?

torchbenchmark/util/experiment/metrics.py
FindHao: I use a new variable `MEMPROF_ITER` and set it to 2. It's better to set a bigger number like 4 if we want to get more stable and accurate results since some models' execution times are too short.

FindHao:
thanks for your suggestions! the last commit has been fixed most of them. the remaining 2 issues are 
1.  I use a new variable MEMPROF_ITER and set it to 2. Do you think it is ok for now? It's better to set a bigger number like 4 if we want to get more stable and accurate results since some models' execution times are too short. 
2. I found [pynvml](https://pypi.org/project/pynvml/#history) is more like one version of [nvidia-ml-py](https://pypi.org/project/nvidia-ml-py/#history). Which one do you think is better to use? nvidia-ml-py is more actively developed but the package name when we try to import in python is still `pynvml` which is a little bit confusing.

torchbenchmark/util/experiment/metrics.py
```diff
@@ -38,19 +41,59 @@ def get_latencies(func, device: str, nwarmup=WARMUP_ROUNDS, num_iter=BENCHMARK_I
         result_summary.append((t1 - t0) / NANOSECONDS_PER_MILLISECONDS)
     return result_summary
 
+def get_peak_memory(func, device: str, num_iter=MEMPROF_ITER, export_metrics_file='', metrics_needed=[], metrics_gpu_backend='dcgm') -> Tuple[float, float]:
```
xuzhao9, line 44: The return value is `Tuple[float, Optional[str], float]`

xuzhao9:
> thanks for your suggestions! the last commit has been fixed most of them. the remaining 2 issues are
> 
> 1. I use a new variable MEMPROF_ITER and set it to 2. Do you think it is ok for now? It's better to set a bigger number like 4 if we want to get more stable and accurate results since some models' execution times are too short.

Is it possible to run 2 iterations and check the result is stable and accurate, if not, keep running at a maximum of 15 iterations?

FindHao:
> Is it possible to run 2 iterations and check the result is stable and accurate, if not, keep running at a maximum of 15 iterations?

The problem is that several model's execution times are too short. model analyzer requests metric measurement records from DCGM backend by the specified interval. By default, model analyzer's frequency is every 10ms. But for DCGM backend, the frequency is about every 1.5ms. It means model analyzer will get about 6~7 records for each request. for short models, not all records has valid data. e.g., if the model execution time is 5ms, we only need the first 3 records. so I add [a timestamp](https://github.com/pytorch/benchmark/pull/1366) when the monitor stops. 

But, sometimes the DCGM timestamps are all larger than the stop timestamp, we have to keep [at least one record](https://github.com/pytorch/benchmark/pull/1391). It is the extreme case. 

DCGM's monitoring crosses multiple iterations, and actually we don't know the execution time for each iteration when model analyzer collects metrics. So it could be hard to check the results' stableness or accuracy online.
But if we have more iterations, we will get more accurate results. I think 2 iterations should be OK for us, even for short models. I'll do the accuracy tests offline to double check it.

xuzhao9:
Another idea is to keep running the model for some time threshold (say 100ms) if single iteration is too short, and if single iteration is long enough, just run 1 or 2 iterations minimum.

FindHao:
dcgm monitor starts before the iterations, ends after all iterations, and works on another thread. So it is hard to let dcgm monitor know each iteration's execution time. If we move the monitors' start and end inside the loop, there would be much more overhead, and it would be complicated to handle the collected data. Since the sampling granularity of dcgm is about 1.5ms, if we make it inside the loop, there would be more noise data.

FindHao:
A possible way is that after the 2 iterations end, we check the execution time. if it is too short, run another memory measurement with more iterations.

xuzhao9:
> A possible way is that after the default 2 iterations end, we check the execution time. if it is too short, run another memory measurement with more iterations.

Yes, that sounds good to me. The basic idea is that we don't need to measure the iteration execution time within the ModelAnalzyer.

xuzhao9:
I get this failure when running internally:

```
  File "/mnt/xarfuse/uid-234232/7d206292-seed-nspid4026531836_cgpid114977510-ns-4026531840/pytorch/benchmark/run.py", line 158, in run_one_step
    cpu_peak_mem, device_id, gpu_peak_mem = get_peak_memory(func, model.device, export_metrics_file=export_metrics_file, metrics_needed=metrics_needed, metrics_gpu_backend=metrics_gpu_backend)
  File "/mnt/xarfuse/uid-234232/7d206292-seed-nspid4026531836_cgpid114977510-ns-4026531840/torchbenchmark/util/experiment/metrics.py", line 70, in get_peak_memory
    mem_model_analyzer.start_monitor()
  File "/mnt/xarfuse/uid-234232/7d206292-seed-nspid4026531836_cgpid114977510-ns-4026531840/components/model_analyzer/TorchBenchAnalyzer.py", line 118, in start_monitor
    self.gpu_factory = GPUDeviceFactory(self.gpu_monitor_backend)
  File "/mnt/xarfuse/uid-234232/7d206292-seed-nspid4026531836_cgpid114977510-ns-4026531840/components/model_analyzer/tb_dcgm_types/gpu_device_factory.py", line 38, in __init__
    self.init_all_devices()
  File "/mnt/xarfuse/uid-234232/7d206292-seed-nspid4026531836_cgpid114977510-ns-4026531840/components/model_analyzer/tb_dcgm_types/gpu_device_factory.py", line 94, in init_all_devices
    gpu_device = GPUDevice(device_name, device_id, pci_bus_id, device_uuid)
  File "/mnt/xarfuse/uid-234232/7d206292-seed-nspid4026531836_cgpid114977510-ns-4026531840/components/model_analyzer/tb_dcgm_types/gpu_device.py", line 50, in __init__
    assert type(device_name) is str
AssertionError
```

components/model_analyzer/tb_dcgm_types/gpu_device_factory.py
```diff
@@ -47,34 +48,58 @@ def init_all_devices(self, dcgmPath=None):
...
+            import pynvml
+
+            logger.info("Initializing GPUDevice handles using NVML")
+            pynvml.nvmlInit()
+
+            # Create a GPU device for every supported NVML device
+            nvml_device_count = pynvml.nvmlDeviceGetCount()
+            for device_id in range(nvml_device_count):
+                handle = pynvml.nvmlDeviceGetHandleByIndex(device_id)
+                device_name = pynvml.nvmlDeviceGetName(handle)
```
xuzhao9, line 90: We need to decode the name here
```suggestion
                device_name = pynvml.nvmlDeviceGetName(handle).decode('ascii')
                pci_bus_id = pynvml.nvmlDeviceGetPciInfo(handle).busId.decode('ascii')
                device_uuid = pynvml.nvmlDeviceGetUUID(handle).decode('ascii')
```

Otherwise they won't pass the check at https://github.com/pytorch/benchmark/blob/main/components/model_analyzer/tb_dcgm_types/gpu_device.py#L51

FindHao: Sorry. I should do more compatibility checks. Testing on pynvml 11.4.1 now. Will fix them soon.

xuzhao9:
Also I found a TODO here: https://github.com/pytorch/benchmark/blob/main/components/model_analyzer/tb_dcgm_types/gpu_device.py#L16 
Maybe we can also migrate to nvml and deprecate numba?

FindHao:
> 

This error has been set to `info` level. Bye default, the logger level in model analyzer is `warning`. So this message will not be shown unless we manually change the logger level. 

numba could detect all GPUs even if you use `NVIDIA_VISIBLE_DEVICES` to limit which GPU could be visible. I think it's better to leave it here for debugging purposes.

xuzhao9:
>  I think it's better to leave it here for debugging purposes. But let me double check if we still need it after we migrate to pynvml.

Let's still remove this, otherwise people will complain for noisy logs.

FindHao:
> > I think it's better to leave it here for debugging purposes. But let me double check if we still need it after we migrate to pynvml.
> 
> Let's still remove this or manually move this to debug level, otherwise people will complain for noisy logs.

Sure. will fix it in the next commit.

FindHao:
@xuzhao9 the warning messages have been commented. It passed the test on pynvml 11.4.1. 
But I failed to get rid of numba. Some information like sm count  is not in pynvml.

xuzhao9:
@FindHao That's fine, we still need sm counts so let's keep numba.

xuzhao9:
@FindHao Can you also move the following logs to debug? We don't need them as INFO.
```
INFO:model_analyzer_logger:Initializing GPUDevice handles using NVML
INFO:numba.cuda.cudadrv.driver:init
INFO:model_analyzer_logger:Using GPU 0 NVIDIA AAA with UUID GPU-BBB
```

PR #1393, opened 2023-02-01
Fix cudagraph backend

xuzhao9:
Fix the CUDA Graph code on torchvision.

Test plan:
```
$ python run.py  resnet18 -d cuda -t eval --backend cudagraph
Running eval method from resnet18 on cuda in eager mode with input batch size 256.
GPU Time:             16.525 milliseconds
CPU Total Wall Time:  16.561 milliseconds
$ python run.py  resnet18 -d cuda -t train --backend cudagraph
Running train method from resnet18 on cuda in eager mode with input batch size 16.
GPU Time:              7.137 milliseconds
CPU Total Wall Time:   7.199 milliseconds
Correctness:                        False
```

### DIFF
--- a/torchbenchmark/util/backends/cudagraph.py
+++ b/torchbenchmark/util/backends/cudagraph.py
@@ -8,20 +8,28 @@ WARMUP_ITER = 3
 def cudagraph(model: 'torchbenchmark.util.model.BenchmarkModel', backend_args: List[str]):
     cudagraph_func_name = f"cudagraph_{model.test}"
     assert hasattr(model, cudagraph_func_name), f"CUDA Graph only works on models implement {cudagraph_func_name}()"
+    if model.test == "train":
+        assert hasattr(model, "SKIP_ZERO_GRAD"), f"The model must support skipping zero grad in its train test."
     def _cudagraph():
         # CUDAGraph can't be copied/pickled, disable copying in correctness checking
         model.DEEPCOPY = False
+        model.SKIP_ZERO_GRAD = True
         # warmup
         s = torch.cuda.Stream()
         s.wait_stream(torch.cuda.current_stream())
         with torch.cuda.stream(s):
             for _ in range(WARMUP_ITER):
+                model.opt.zero_grad(set_to_none=True)
                 model.invoke()
         torch.cuda.current_stream().wait_stream(s)
         # capture
         cuda_graph = torch.cuda.CUDAGraph()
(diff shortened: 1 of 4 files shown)


torchbenchmark/util/backends/cudagraph.py
```diff
@@ -1,26 +1,37 @@
...
+        cuda_graph = torch.cuda.CUDAGraph()
+        if model.test == "train":
+            model.opt.zero_grad(set_to_none=True)
+        with torch.cuda.graph(cuda_graph):
+            model.invoke()
+        model.g = cuda_graph
+        if model.test == "train":
+            model.train = getattr(model, cudagraph_func_name)
+        else:
+            model.eval = getattr(model, cudagraph_func_name)
```
davidberard98, line 36: nit: assuming only model.test == "train" or "eval" exist - if so maybe do `elif model.test == "eval"`?

torchbenchmark/util/backends/cudagraph.py
```diff
@@ -1,26 +1,37 @@
...
-    model.g = g
\ No newline at end of file
+WARMUP_ITER = 3
+
+@create_backend
+def cudagraph(model: 'torchbenchmark.util.model.BenchmarkModel', backend_args: List[str]):
+    cudagraph_func_name = f"cudagraph_{model.test}"
+    assert hasattr(model, cudagraph_func_name), f"CUDA Graph only works on models implement {cudagraph_func_name}()"
+    if model.test == "train":
+        assert hasattr(model, "SKIP_ZERO_GRAD"), f"The model must support skipping zero grad in its train test."
```
davidberard98, line 12: should this be `getattr(model, "SKIP_ZERO_GRAD", False)` ? in case SKIP_ZERO_GRAD is defined and is set to false

davidberard98 (approved):
minor comments, otherwise looks good!

torchbenchmark/util/backends/cudagraph.py
xuzhao9: `model.SKIP_ZERO_GRAD` should be False in eager mode, we should only set it to True when running with cudagraph. So it is fine to be False here.

torchbenchmark/util/backends/cudagraph.py
xuzhao9: Thanks for pointing out! Fixed.

PR #1395, opened 2023-02-02
check for fx2trt only for cuda device

snadampal:
this change is required to make the cpu device (-d cpu) work on the latest commit.

### DIFF
--- a/torchbenchmark/util/model.py
+++ b/torchbenchmark/util/model.py
@@ -164,7 +164,7 @@ class BenchmarkModel(metaclass=PostInitProcessor):
                 self.dargs.precision == "fp16"
                 or self.dargs.precision == "amp"
                 or (self.dynamo and self.opt_args.torchdynamo == "fx2trt")
-                or (not self.dynamo and self.opt_args.fx2trt)
+                or (not self.dynamo and (self.device == "cuda" and self.opt_args.backend == "fx2trt"))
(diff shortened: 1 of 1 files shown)


torchbenchmark/util/model.py
```diff
@@ -164,7 +164,7 @@ def __post__init__(self):
                 self.dargs.precision == "fp16"
                 or self.dargs.precision == "amp"
                 or (self.dynamo and self.opt_args.torchdynamo == "fx2trt")
-                or (not self.dynamo and self.opt_args.fx2trt)
+                or (not self.dynamo and (self.device == "cuda" and self.opt_args.fx2trt))
```
xuzhao9, line 167: It should be:

```suggestion
                or (not self.dynamo and (self.device == "cuda" and self.opt_args.backend == "fx2trt"))
```

xuzhao9 (commented):
Thanks for the PR! It is a bug caused by our recent code change.

We should use `self.opt_args.backend == "fx2trt"` to check the fx2trt backend now.

torchbenchmark/util/model.py
snadampal: thanks for the quick review, will update.
yes, looks like the recent cuda fx2trt changes introduced this.

xuzhao9 (approved):

PR #1397, opened 2023-02-04
Add model torchrec dlrm

xuzhao9:
Original model file: https://github.com/facebookresearch/dlrm/blob/main/torchrec_dlrm/dlrm_main.py
Using the default config in the model file.
Need to modify a few places to run on single GPU device, including removing DDP, change device from "meta" to concrete devices, and remove fused optimizer.

Test Plan:
Experiment on T4 16 GB
```
$ python run.py torchrec_dlrm -d cuda --bs 4096 -t eval
Running eval method from torchrec_dlrm on cuda in eager mode with input batch size 4096.
GPU Time:              5.040 milliseconds
CPU Total Wall Time:   5.066 milliseconds

$ python run.py torchrec_dlrm -d cuda --bs 4096 -t train
Running train method from torchrec_dlrm on cuda in eager mode with input batch size 4096.
GPU Time:             32.600 milliseconds
CPU Total Wall Time:  32.625 milliseconds
```

```
$ python run.py torchrec_dlrm -d cuda --bs 4096 -t eval --torchdynamo inductor
[2023-02-04 14:14:36,265] torch._inductor.compile_fx: [WARNING] skipping cudagraphs due to multiple devices
Segmentation fault (core dumped)

$ python run.py torchrec_dlrm -d cuda --bs 4096 -t train --torchdynamo inductor
[2023-02-04 14:14:36,265] torch._inductor.compile_fx: [WARNING] skipping cudagraphs due to multiple devices
Segmentation fault (core dumped)
```

### DIFF
--- a/torchbenchmark/models/torchrec_dlrm/__init__.py
+++ b/torchbenchmark/models/torchrec_dlrm/__init__.py
@@ -140,8 +140,6 @@ class Model(BenchmarkModel):
         return self.model, self.example_inputs
 
     def forward(self):
-        if self.opt:
-            self.opt.zero_grad()
         losses, output = self.model(self.example_inputs)
         return losses
 
@@ -149,8 +147,7 @@ class Model(BenchmarkModel):
         torch.sum(losses, dim=0).backward()
 
     def optimizer(self):
-        if self.opt:
-            self.opt.step()
+        self.opt.step()
 
     def eval(self):
         with torch.no_grad():
--- a/torchbenchmark/models/torchrec_dlrm/__init__.py
+++ b/torchbenchmark/models/torchrec_dlrm/__init__.py
@@ -33,19 +33,18 @@ class Model(BenchmarkModel):
     def __init__(self, test, device, jit, batch_size=None, extra_args=[]):
         super().__init__(test=test, device=device, jit=jit, batch_size=batch_size, extra_args=extra_args)
         args = parse_args(self.extra_args)
+        backend = "nccl" if self.device == "cuda" else "gloo"
         # initialize example data
         if self.test == "train":
             args.batch_size = self.batch_size
-            train_dataloader = get_dataloader(args, backend, "train")
-            example_inputs = next(train_dataloader)
+            loader = get_dataloader(args, backend, "train")
         if self.test == "eval":
             args.test_batch_size = self.batch_size
-            test_dataloader = get_dataloader(args, backend, "test")
-            example_inputs = next(test_dataloader)
+            loader = get_dataloader(args, backend, "test")
         # prefetch data to device
-        self.example_inputs = example_inputs.to(self.device)
+        for data in loader:
+            self.example_inputs = data.to(self.device)
 
-        backend = "nccl" if self.device == "cuda" else "gloo"
         assert args.in_memory_binary_criteo_path == None and args.synthetic_multi_hot_criteo_path == None, \
             f"Torchbench only supports random data inputs."
 
--- a/torchbenchmark/models/torchrec_dlrm/__init__.py
+++ b/torchbenchmark/models/torchrec_dlrm/__init__.py
@@ -68,7 +68,7 @@ class Model(BenchmarkModel):
         if args.interaction_type == InteractionType.ORIGINAL:
             dlrm_model = DLRM(
                 embedding_bag_collection=EmbeddingBagCollection(
-                    tables=eb_configs, device=torch.device("meta")
+                    tables=eb_configs, device=self.device
                 ),
                 dense_in_features=len(DEFAULT_INT_NAMES),
                 dense_arch_layer_sizes=args.dense_arch_layer_sizes,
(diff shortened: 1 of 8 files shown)


torchbenchmark/models/torchrec_dlrm/__init__.py
```diff
@@ -0,0 +1,156 @@
...
+            args.batch_size = self.batch_size
+            loader = get_dataloader(args, backend, "train")
+        if self.test == "eval":
+            args.test_batch_size = self.batch_size
+            loader = get_dataloader(args, backend, "test")
+        limit_batches = 1
+        iterator = itertools.islice(iter(loader), limit_batches)
+        self.example_inputs = next(iterator).to(self.device)
+        # parse the args
+        args.dense_arch_layer_sizes = [int(x) for x in args.dense_arch_layer_sizes.split(',') if x.strip().isdigit()]
```
xuzhao9, line 49: Need to parse the arg from `str` to `List[int]`, a bug in the original model file?

torchbenchmark/models/torchrec_dlrm/__init__.py
```diff
@@ -0,0 +1,156 @@
...
+                ),
+                dense_in_features=len(DEFAULT_INT_NAMES),
+                dense_arch_layer_sizes=args.dense_arch_layer_sizes,
+                over_arch_layer_sizes=args.over_arch_layer_sizes,
+                dense_device=device,
+            )
+        elif args.interaction_type == InteractionType.DCN:
+            dlrm_model = DLRM_DCN(
+                embedding_bag_collection=EmbeddingBagCollection(
+                    tables=eb_configs, device=self.device
```
xuzhao9, line 81: Originally device was `meta`, need to update it to `self.device`. Otherwise there's error like "RuntimeError: Tensor on device meta is not on the expected device cuda:0!"

torchbenchmark/models/torchrec_dlrm/__init__.py
```diff
@@ -0,0 +1,156 @@
...
+                return lambda params: torch.optim.SGD(params, lr=args.learning_rate)
+
+        self.model = train_model
+        dense_optimizer = KeyedOptimizerWrapper(
+            dict(in_backward_optimizer_filter(self.model.named_parameters())),
+            optimizer_with_params(),
+        )
+        # opt = CombinedOptimizer([self.model.fused_optimizer, dense_optimizer])
+        # Only run dense optimizer
+        opt = CombinedOptimizer([dense_optimizer])
```
xuzhao9, line 130: Originally it was `[model.fused_optimizer, dense_optimizer]`. We don't have DDP model here, so removed `fused_optimizer`. Is there a better solution here?

torchbenchmark/models/torchrec_dlrm/args.py
```diff
@@ -0,0 +1,243 @@
...
+        "--num_embeddings_per_feature",
+        type=str,
+        default=None,
+        help="Comma separated max_ind_size per sparse feature. The number of embeddings"
+        " in each embedding table. 26 values are expected for the Criteo dataset.",
+    )
+    parser.add_argument(
+        "--dense_arch_layer_sizes",
+        type=str,
+        default="512,256,64",
```
xuzhao9, line 81: All use the default arch layer sizes, seems to be too small on A100. Do we need to update to larger model?

brad-mengchi (approved):

torchbenchmark/models/torchrec_dlrm/requirements.txt
```diff
@@ -0,0 +1,2 @@
+torchrec-nightly
```
colin2328, line 1: this will be using nightly of TorchRec/ FBGEMM GPU? is that typical for torchbench?

Is it also using nightly torch?

xuzhao9: Yes, all torchbench CI workflows are using nightly pytorch by default. My concern of using URL to install is that users may experience longer time to compile and install the packages.

torchbenchmark/models/torchrec_dlrm/metadata.yaml
```diff
@@ -0,0 +1,6 @@
+eval_benchmark: false
+eval_deterministic: false
+eval_nograd: true
+train_benchmark: false
+train_deterministic: false
+skip_cuda_memory_leak: true
```
xuzhao9, line 6: We noticed this model will "memory leak" (i.e., some CUDA memory are still in use after deleting the model instance). In the past, it happened on LSTM models which has cyclic references. Not sure if it is the same case  for torchrec_dlrm, any ideas? @colin2328

colin2328: strange, I haven't seen this before

colin2328 (commented):
@xuzhao9 , I don't think this model is actually running through trec sharder. will message offline

PR #1408, opened 2023-02-10
[model-stableness] Fix type in iter rounds

xuzhao9:
torch.cuda.empty_cache() is causing stableness issues in yolov3, so we are removing it from the benchmark test.


FindHao (approved):
LGTM

janeyx99:
@xuzhao9 @FindHao is the stableness issue referring to just my recent set_to_none change or has it been affecting yolov3 benchmarks more broadly?

FindHao:
I think it does affect the yolov3 stableness but the performance regression caused by set_to_none is also true.

PR #1410, opened 2023-02-13
change fastNLP_Bert's task name

FindHao:
fastNLP_Bert is the only one classified as `NLP.OTHER_NLP`. Since BERT_pytorch and hf_Bert are classified to `NLP.LANGUAGE_MODELING`, is it better to use `NLP.LANGUAGE_MODELING` here?


xuzhao9 (approved):
LGTM

Issue #1413, opened 2023-02-15
profile report CUDA failed in CPU only machine.


xuhancn:
Windows:
---------------------------------------------------------------------------------
PS D:\xu_github\benchmark> python.exe ./run.py -d cpu -m jit --bs=12 alexnet --stress 200 --profile
Running eval method from alexnet on cpu in jit mode with input batch size 12.
STAGE:2023-02-15 13:24:24 11468:11628 ..\third_party\kineto\libkineto\src\ActivityProfilerController.cpp:294] Completed Stage: Warm Up
STAGE:2023-02-15 13:24:24 11468:11628 ..\third_party\kineto\libkineto\src\ActivityProfilerController.cpp:300] Completed Stage: Collection
Traceback (most recent call last):
  File "D:\xu_github\benchmark\run.py", line 333, in <module>
    profile_one_step(test)
  File "D:\xu_github\benchmark\run.py", line 229, in profile_one_step
    start_event = torch.cuda.Event(enable_timing=True)
  File "C:\Users\1\AppData\Local\Programs\Python\Python310\lib\site-packages\torch\cuda\streams.py", line 164, in __new__
    return super(Event, cls).__new__(
TypeError: object.__new__() takes exactly one argument (the type to instantiate)

Linux:
---------------------------------------------------------------------------------
xu@xu-G3-3590:~/torchbench/benchmark$ python3 run.py -d cpu -m jit --bs=12 resnet50 --stress 200 --profile --profile-devices cpu
Running eval method from resnet50 on cpu in jit mode with input batch size 12.
STAGE:2023-02-15 21:20:54 2315:2315 ActivityProfilerController.cpp:294] Completed Stage: Warm Up
STAGE:2023-02-15 21:20:54 2315:2315 ActivityProfilerController.cpp:300] Completed Stage: Collection
Traceback (most recent call last):
  File "run.py", line 333, in <module>
    profile_one_step(test)
  File "run.py", line 233, in profile_one_step
    start_event.record()
  File "/home/xu/.local/lib/python3.8/site-packages/torch/cuda/streams.py", line 179, in record
    stream = torch.cuda.current_stream()
  File "/home/xu/.local/lib/python3.8/site-packages/torch/cuda/__init__.py", line 591, in current_stream
    _lazy_init()
  File "/home/xu/.local/lib/python3.8/site-packages/torch/cuda/__init__.py", line 229, in _lazy_init
    torch._C._cuda_init()
[...]

PR #1415, opened 2023-02-15
Fix vision model on inductor and profile on CPU only machines

xuzhao9:
Fixes https://github.com/pytorch/benchmark/issues/1413 at the same time.


FindHao (approved):
LGTM

PR #1424, opened 2023-02-22
Add torchtext to fix benchmark installation error

spzala:
TorchText is missing in the command,
`conda install pytorch torchvision torchaudio pytorch-cuda=11.7 -c pytorch-nightly -c nvidia` resulting in the following installation error:
```
python install.py
checking packages torch, torchvision, torchtext, torchaudio are installed...FAIL Error: Users must first manually install packages
['torch', 'torchvision', 'torchtext', 'torchaudio'] before installing the benchmark.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ conda install -y -c pytorch magma-cuda117
 
 Then install pytorch, torchtext, torchvision, and torchaudio using conda:
 ```
-conda install pytorch torchvision torchaudio pytorch-cuda=11.7 -c pytorch-nightly -c nvidia
+conda install pytorch torchvision torchtext torchaudio pytorch-cuda=11.7 -c pytorch-nightly -c nvidia
 ```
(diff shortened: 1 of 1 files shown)


spzala:
cc @xuzhao9 @FindHao

xuzhao9 (approved):
Thanks for the fix!

spzala:
@xuzhao9 thank you so much for the quick review!!! Also, I don't think the test failures are related to the PR changes but if anything that I may need to do, please let me know.

spzala:
> > @xuzhao9 thank you so much for the quick review!!! Also, I don't think the test failures are related to the PR changes but if anything that I may need to do, please let me know.
> 
> The test failures are not related to this PR and is a known issue caused by an upstream issue of fbgemm, we are still working on it.

Ok, sounds great. Thanks @xuzhao9 !!

PR #1430, opened 2023-02-23
Fix detectron2 models on cpu-only torch

### LINKED ISSUE #1429: Detectron2 models loading issues with cpu-only torch builds
The current model factory of `detectron2` will fetch the default cfg settings, which uses cuda as cfg.MODEL.DEVICE, and convert to whatever `device` user specified while loading the models. One caveat is that `detectrons` models cannot be loaded properly if the underlying torch is cpu only. Therefore making the `detectron2` models unusable for torchbenchmark with cpu-only torch.

Repro:
```
# the benchmarks script is from pytorch repo, make sure to run this on cpu-only torch
python benchmarks/dynamo/torchbench.py --only detectron2_fasterrcnn_r_101_c4 -dcpu --performance 
```

The error message while loading `detectron2` models:
```
Torch not compiled with CUDA enabled
```

cc: @xuzhao9

shingjan:
This PR intends to fix #1429 .

cc:
@xuzhao9

### DIFF
--- a/torchbenchmark/util/framework/detectron2/model_factory.py
+++ b/torchbenchmark/util/framework/detectron2/model_factory.py
@@ -90,7 +90,7 @@ class Detectron2Model(BenchmarkModel):
             args.fcos_use_bn = True
 
         cfg = setup(args)
-        if cfg.MODEL.DEVICE != self.device:
+        if getattr(cfg, "MODEL", None) and cfg.MODEL.DEVICE != self.device:
             cfg.MODEL.DEVICE == self.device
         if args.config_file.endswith(".yaml"):
             self.model = build_model(cfg).to(self.device)
--- a/torchbenchmark/util/framework/detectron2/model_factory.py
+++ b/torchbenchmark/util/framework/detectron2/model_factory.py
@@ -91,7 +91,9 @@ class Detectron2Model(BenchmarkModel):
 
         cfg = setup(args)
         if hasattr(cfg, "MODEL") and cfg.MODEL.DEVICE != self.device:
+            cfg.defrost()
             cfg.MODEL.DEVICE = self.device
+            cfg.freeze()
         if args.config_file.endswith(".yaml"):
             self.model = build_model(cfg).to(self.device)
         else:
--- a/torchbenchmark/util/framework/detectron2/model_factory.py
+++ b/torchbenchmark/util/framework/detectron2/model_factory.py
@@ -90,6 +90,8 @@ class Detectron2Model(BenchmarkModel):
             args.fcos_use_bn = True
 
         cfg = setup(args)
+        if cfg.MODEL.DEVICE != self.device:
+            cfg.MODEL.DEVICE == self.device
         if args.config_file.endswith(".yaml"):
             self.model = build_model(cfg).to(self.device)
         else:
--- a/torchbenchmark/util/framework/detectron2/model_factory.py
+++ b/torchbenchmark/util/framework/detectron2/model_factory.py
@@ -90,7 +90,7 @@ class Detectron2Model(BenchmarkModel):
             args.fcos_use_bn = True
 
         cfg = setup(args)
-        if getattr(cfg, "MODEL", None) and cfg.MODEL.DEVICE != self.device:
+        if hasattr(cfg, "MODEL") and cfg.MODEL.DEVICE != self.device:
             cfg.MODEL.DEVICE = self.device
         if args.config_file.endswith(".yaml"):
             self.model = build_model(cfg).to(self.device)
--- a/torchbenchmark/util/framework/detectron2/model_factory.py
+++ b/torchbenchmark/util/framework/detectron2/model_factory.py
@@ -91,7 +91,7 @@ class Detectron2Model(BenchmarkModel):
 
         cfg = setup(args)
         if getattr(cfg, "MODEL", None) and cfg.MODEL.DEVICE != self.device:
-            cfg.MODEL.DEVICE == self.device
+            cfg.MODEL.DEVICE = self.device
         if args.config_file.endswith(".yaml"):
             self.model = build_model(cfg).to(self.device)
         else:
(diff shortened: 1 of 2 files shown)


xuzhao9:
Thanks for the fix! Can you please add a test plan to this PR? For example:
 
```
python run.py detectron2_fasterrcnn_r_101_c4 -d cuda -t train
<output here...>
```

shingjan:
@xuzhao9 Thanks for the prompt review! That line is removed in the PR and a test plan for this PR is as follows:
```
python run.py detectron2_fasterrcnn_r_101_c4 -d cpu -t eval

# output

Running eval method from detectron2_fasterrcnn_r_101_c4 on cpu in eager mode with input batch size 1.
/home/yj/pytorch/torch/functional.py:504: UserWarning: torch.meshgrid: in an upcoming release, it will be required to pass the indexing argument. (Triggered internally at /home/yj/pytorch/aten/src/ATen/native/TensorShape.cpp:3483.)
  return _VF.meshgrid(tensors, **kwargs)  # type: ignore[attr-defined]
CPU Total Wall Time: 3277.488 milliseconds
```

Note that I installed detectron2 from source instead of from pypi as there is a undefined symbol issues.

xuzhao9 (approved):
LGTM!

xuzhao9:
@shingjan Can you take a look at the CI failure and fix it?

xuzhao9 (changes requested):
Can you help fix the CI error?

torchbenchmark/util/framework/detectron2/model_factory.py
```diff
@@ -90,6 +90,8 @@ def __init__(self, variant, test, device, jit=False, batch_size=None, extra_args
             args.fcos_use_bn = True
 
         cfg = setup(args)
+        if getattr(cfg, "MODEL", None) and cfg.MODEL.DEVICE != self.device:
+            cfg.MODEL.DEVICE == self.device
```
xuzhao9, line 94: Should it be `cfg.MODEL.DEVICE = self.device`?
```suggestion
            cfg.MODEL.DEVICE = self.device
```

shingjan: keen eyes! This is fixed

torchbenchmark/util/framework/detectron2/model_factory.py
```diff
@@ -90,6 +90,8 @@ def __init__(self, variant, test, device, jit=False, batch_size=None, extra_args
             args.fcos_use_bn = True
 
         cfg = setup(args)
+        if hasattr(cfg, "MODEL") and cfg.MODEL.DEVICE != self.device:
```
xuzhao9, line 93: What happens if cfg doesn't have the "MODEL" attr?

shingjan:
@xuzhao9 really appreciate the review! Seems like workflow now needs maintainer's approval.

torchbenchmark/util/framework/detectron2/model_factory.py
shingjan: From the [impl](https://github.com/facebookresearch/detectron2/blob/80e2673da161f57afe37ef769836a61976108ef1/detectron2/modeling/meta_arch/build.py#L23) of buliding detectron2 models, it looks to me that the assumption is that a `cfg` should always have the attribute `MODEL`. But per the last CI run, it seems that this assumption no longer holds at least for model `test_detectron2_fcos_r_50_fpn_eval_cuda`. Details [here](https://github.com/pytorch/benchmark/actions/runs/4258041173/jobs/7420088377#step:9:271). I am not familiar with this test and wonder what you recommendation would be for this case. Thanks!

xuzhao9:
@shingjan Can you help fix the new CI errors?
You can test it locally with the command 

```
python test.py -k detectron2_fasterrcnn_r_101_c4
```

shingjan:
@xuzhao9 With the latest change I ran `python run.py detectron2_fasterrcnn_r_101_c4 -d cpu -t eval` with output
```
Running eval method from detectron2_fasterrcnn_r_101_c4 on cpu in eager mode with input batch size 1.
/home/yj/pytorch/torch/functional.py:504: UserWarning: torch.meshgrid: in an upcoming release, it will be required to pass the indexing argument. (Triggered internally at /home/yj/pytorch/aten/src/ATen/native/TensorShape.cpp:3488.)
  return _VF.meshgrid(tensors, **kwargs)  # type: ignore[attr-defined]
CPU Total Wall Time: 3449.469 milliseconds
```
ran `python test.py -k detectron2_fasterrcnn_r_101_c4` with output
```
........
Ran 8 tests in 48.829s

OK
```
Its still possible that CI has a different setup so the outputs may vary. Thx!

torchbenchmark/util/framework/detectron2/model_factory.py
xuzhao9: Yes, this is because `detectron2_fcos_r_50_fpn_eval_cuda` doesn't use `MODEL.DEVICE` to initialize the model device. I am pretty sure this PR only enables the detectron2_fasterrcnn and detectron2_maskrcnn models for cpu-only torch, but [detectron2_fcos_r_50_fpn](https://github.com/pytorch/benchmark/tree/main/torchbenchmark/models/detectron2_fcos_r_50_fpn) is probably still GPU-only.

I am okay with that, and we can enable CPU with detectron2_fcos_r_50_fpn in another PR.

xuzhao9 (approved):
LGTM, thanks for the improvement.

shingjan:
@xuzhao9 Thanks! Seems like CI for facebook internal failed, any hint that I can fix it on my end?

xuzhao9:
@shingjan Those are a few timeout internal tests, no need to worry.

shingjan:
@xuzhao9 sounds good! Is there a way we can merge this pr? Thx

xuzhao9:
Sorry I have some rebasing issues with internal landing - can you help rebase this PR on the latest main branch and push again? Thanks! @shingjan

PR #1433, opened 2023-02-24
Using FP32 for DRQ network.

kulinseth:
Moving the temperature numpy log to FP32.

### DIFF
--- a/torchbenchmark/models/drq/drq.py
+++ b/torchbenchmark/models/drq/drq.py
@@ -167,7 +167,8 @@ class DRQAgent(object):
         # tie conv layers between actor and critic
         self.actor.encoder.copy_conv_weights_from(self.critic.encoder)
 
-        self.log_alpha = torch.tensor(np.log(cfg.init_temperature)).to(device)
+        self.log_alpha = torch.tensor(np.log(cfg.init_temperature,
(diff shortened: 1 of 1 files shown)


xuzhao9:
I am wondering what is the motivation of this change?

If we are changing this, I hope we can update the upstream code at the same time: https://github.com/denisyarats/drq/blob/master/drq.py#L188

kulinseth:
> I am wondering what is the motivation of this change?
> 
> If we are changing this, I hope we can update the upstream code at the same time: https://github.com/denisyarats/drq/blob/master/drq.py#L188

Hi @xuzhao9 , the motivation is that Metal doesn't support FP64. And as a result this benchmark fails on MPS device. From the change it seems like by default `np.log` returns FP64 , if it won't have any impact on the networks performance. It will be nice to just use FP32.

xuzhao9:
@kulinseth Since we don't do any end-to-end accuracy checks in the core models set, we need to be very careful when modifying the model code.

Therefore, I believe we can only accept this PR if upstream model code repo uses FP32, otherwise we will keep it as-is.

cc @adnanaziz for ideas and suggestions.

kulinseth:
Hi @xuzhao9 , that makes sense. I will create an upstream PR to propose similar fix.

PR #1434, opened 2023-02-24
Making BERT_pytorch work with MPS backend

kulinseth:


torchbenchmark/models/BERT_pytorch/__init__.py
```diff
@@ -86,7 +86,7 @@ def __init__(self, test, device, batch_size=None, jit=False, extra_args=[]):
             '--vocab_path', f'{root}/data/vocab.small',
             '--output_path', 'bert.model',
         ]) # Avoid reading sys.argv here
-        args.with_cuda = self.device == 'cuda'
+        args.with_cuda = self.device == 'cuda' or self.device == 'mps'
```
kulinseth, line 89: Piggybacking on the cuda flag as i didn't want to create additional changes in the function signature.

torchbenchmark/models/BERT_pytorch/bert_pytorch/trainer/pretrain.py
```diff
@@ -35,8 +35,12 @@ def __init__(self, bert: BERT, vocab_size: int,
         """
 
         # Setup cuda device for BERT training, argument -c, --cuda should be true
-        cuda_condition = torch.cuda.is_available() and with_cuda
-        self.device = torch.device("cuda:0" if cuda_condition else "cpu")
+        if (with_cuda and torch.cuda.is_available()):
```
xuzhao9, line 38: Same here, we can make it easier by using `self.device = device`

torchbenchmark/models/BERT_pytorch/__init__.py
xuzhao9: I suggest we remove the `with_cuda` flag and replace it with `device` and pass in `self.device`.

PR #1437, opened 2023-02-25
Add MPS device to YoloV3

kulinseth:


torchbenchmark/models/yolov3/yolo_utils/torch_utils.py
```diff
@@ -25,6 +25,10 @@ def init_seeds(seed=0):
 
 def select_device(device='', apex=False, batch_size=None):
     # device = 'cpu' or '0' or '0,1,2,3'
+
+    if (torch.backends.mps.is_available()):
```
davidberard98, line 29: This will skip a specified device if mps is available, is that intended?

kulinseth: Good catch, i will add the CPU check.

PR #1446, opened 2023-03-03
Add LLAMA

msaroufim:
- LLAMA
- Add LLAMA

### DIFF
--- a/torchbenchmark/models/llama/__init__.py
+++ b/torchbenchmark/models/llama/__init__.py
@@ -1,0 +1,43 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# This software may be used and distributed according to the terms of the GNU General Public License version 3.
+
+
+
+from ...util.model import BenchmarkModel
+from typing import Tuple
+import os
+import sys
+import torch
+import fire
+import time
+import json
+
+from pathlib import Path
+
+from fairscale.nn.model_parallel.initialize import initialize_model_parallel
+
+from llama import ModelArgs, Transformer, Tokenizer, LLaMA
+import torch
+
+torch.backends.cudnn.deterministic = False
+torch.backends.cudnn.benchmark = True
+
+class Model(BenchmarkModel):
+    def __init__(self, temperature: float = 0.8, top_p: float = 0.95):
+        self.model_args = ModelArgs()
+        self.generator = Transformer(self.model_args)
+        self.temperature = temperature
+        self.top_p = top_p
+    
+    def inference(self, prompts : str):
+        prompts = ["The capital of Germany is the city of", "Here is my sonnet in the style of Shakespeare about an artificial intelligence:"]
+        results = self.generator.generate(prompts, max_gen_len=256, temperature=self.temperature, top_p=self.top_p)
+
+        for result in results:
+            print(result)
+            print("\n==================================\n")
+
+
+
+
+
--- a/torchbenchmark/models/llama/llama/__init__.py
+++ b/torchbenchmark/models/llama/llama/__init__.py
@@ -1,0 +1,6 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# This software may be used and distributed according to the terms of the GNU General Public License version 3.
+
+from .generation import LLaMA
+from .model import ModelArgs, Transformer
+from .tokenizer import Tokenizer
--- a/torchbenchmark/models/llama/llama/generation.py
+++ b/torchbenchmark/models/llama/llama/generation.py
@@ -1,0 +1,77 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# This software may be used and distributed according to the terms of the GNU General Public License version 3.
+
+from typing import List
+
+import torch
+
+from llama.tokenizer import Tokenizer
+from llama.model import Transformer
+
+
+class LLaMA:
+    def __init__(self, model: Transformer, tokenizer: Tokenizer):
+        self.model = model
+        self.tokenizer = tokenizer
+
+    def generate(
+        self,
+        prompts: List[str],
+        max_gen_len: int,
+        temperature: float = 0.8,
+        top_p: float = 0.95,
+    ) -> List[str]:
+        bsz = len(prompts)
+        params = self.model.params
+        assert bsz <= params.max_batch_size, (bsz, params.max_batch_size)
+
+        prompt_tokens = [self.tokenizer.encode(x, bos=True, eos=False) for x in prompts]
+
+        min_prompt_size = min([len(t) for t in prompt_tokens])
+        max_prompt_size = max([len(t) for t in prompt_tokens])
+
+        total_len = min(params.max_seq_len, max_gen_len + max_prompt_size)
+
+        tokens = torch.full((bsz, total_len), self.tokenizer.pad_id).cuda().long()
+        for k, t in enumerate(prompt_tokens):
+            tokens[k, : len(t)] = torch.tensor(t).long()
+        input_text_mask = tokens != self.tokenizer.pad_id
+        start_pos = min_prompt_size
+        prev_pos = 0
+        for cur_pos in range(start_pos, total_len):
+            logits = self.model.forward(tokens[:, prev_pos:cur_pos], prev_pos)
+            if temperature > 0:
+                probs = torch.softmax(logits / temperature, dim=-1)
+                next_token = sample_top_p(probs, top_p)
+            else:
+                next_token = torch.argmax(logits, dim=-1)
+            next_token = next_token.reshape(-1)
+            # only replace token if prompt has already been generated
+            next_token = torch.where(
+                input_text_mask[:, cur_pos], tokens[:, cur_pos], next_token
+            )
+            tokens[:, cur_pos] = next_token
+            prev_pos = cur_pos
+
+        decoded = []
+        for i, t in enumerate(tokens.tolist()):
+            # cut to max gen len
+            t = t[: len(prompt_tokens[i]) + max_gen_len]
+            # cut to eos tok if any
+            try:
+                t = t[: t.index(self.tokenizer.eos_id)]
+            except ValueError:
+                pass
+            decoded.append(self.tokenizer.decode(t))
+        return decoded
+
+
+def sample_top_p(probs, p):
+    probs_sort, probs_idx = torch.sort(probs, dim=-1, descending=True)
+    probs_sum = torch.cumsum(probs_sort, dim=-1)
+    mask = probs_sum - probs_sort > p
+    probs_sort[mask] = 0.0
+    probs_sort.div_(probs_sort.sum(dim=-1, keepdim=True))
+    next_token = torch.multinomial(probs_sort, num_samples=1)
+    next_token = torch.gather(probs_idx, -1, next_token)
+    return next_token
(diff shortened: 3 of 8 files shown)


xuzhao9:
I get the following error when running the model locally with run.py:
```
$ python run.py llama -d cpu
Warning: Could not find dependent module llama for Model llama, skip it.
 No module named 'llama'
Unable to find model matching llama.
```

Can you help try and see if you can reproduce?

ezyang:
The limitations look fine; I prefer not downloading the checkpoint.

msaroufim:
Ok this should work 

```
(bench) ubuntu@ip-172-31-38-220:~/benchmark$ python run.py llama -d cpu
Running eval method from llama on cpu in eager mode with input batch size 32.
CPU Total Wall Time:  10.136 milliseconds
CPU Peak Memory:                2.4590 GB
(bench) ubuntu@ip-172-31-38-220:~/benchmark$ python run.py llama -d cuda
Running eval method from llama on cuda in eager mode with input batch size 32.
GPU Time:              8.040 milliseconds
CPU Total Wall Time:   8.067 milliseconds
GPU 0 Peak Memory:              1.9117 GB
CPU Peak Memory:                2.0645 GB
```

torchbenchmark/models/llama/__init__.py
```diff
@@ -0,0 +1,51 @@
...
+from .model import ModelArgs, Transformer
+import torch
+
+class Model(BenchmarkModel):
+    DEFAULT_EVAL_BSIZE = 32
+    task = NLP.LANGUAGE_MODELING
+
+    def __init__(self, test, device, jit=False, batch_size=DEFAULT_EVAL_BSIZE, extra_args=[]):
+        super().__init__(test=test, device=device, jit=jit, batch_size=batch_size, extra_args=extra_args)
+        self.model_args = ModelArgs(vocab_size=32) # TODO: Configuring arguments is breaking stuff: max_batch_size=batch_size, max_seq_len=1032 is breaking stuff
```
xuzhao9, line 18: Curious what is broken here?

torchbenchmark/models/llama/__init__.py
```diff
@@ -0,0 +1,51 @@
...
+from torchbenchmark.tasks import NLP
+import torch
+from .model import ModelArgs, Transformer
+import torch
+
+class Model(BenchmarkModel):
+    DEFAULT_EVAL_BSIZE = 32
+    task = NLP.LANGUAGE_MODELING
+
+    def __init__(self, test, device, jit=False, batch_size=DEFAULT_EVAL_BSIZE, extra_args=[]):
```
xuzhao9, line 16: Let's keep `batch_size=None`, in the parent class, it will setup the correct batch_size based on the value of `test`.

torchbenchmark/models/llama/__init__.py
```diff
@@ -0,0 +1,51 @@
...
+            The weights for this model are not publicly available and require a valid research reason to use
+            The publicly available github repo is inference only
+            https://github.com/facebookresearch/llama
+        """
+        return NotImplementedError(error_msg)
+
+    def eval(self):
+        self.model.eval()
+        with torch.no_grad():
+            for example_input in self.example_inputs:
```
xuzhao9, line 44: IMO we don't need the for loop since `self.example_inputs` length is 1.

msaroufim: I believe I still do since get_module expects the example input to be a list

torchbenchmark/models/llama/__init__.py
msaroufim: Fixed

torchbenchmark/models/llama/__init__.py
msaroufim: The LLAMA model need some extra validation for the ModelArgs since not all arg configs are valid

msaroufim:
@xuzhao9 I'm a bit confused by this error,

running locally it works but in CI getting a failure that start_pos is not given

```
(bench) ubuntu@ip-172-31-38-220:~/benchmark$ python test.py -k test_llama_example_cuda
s
Ran 1 test in 1.399s

OK (skipped=1)
```

torchbenchmark/models/llama/__init__.py
xuzhao9: We need to make sure `example_inputs` returned by `get_module()` returns *ALL* the data that will be used in `train()` or `eval()`. Otherwise, features like `channels_last` will be broken.

torchbenchmark/models/llama/__init__.py
xuzhao9: Thanks for the clarification!

torchbenchmark/models/llama/__init__.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+        if device == "cuda":
+            torch.set_default_device("cuda")
+            self.model.to(torch.device("cuda"))
+        self.example_inputs = [(torch.tensor([[1, 1], [1,1]], dtype=torch.int), 1)]
+
+        
+    def get_module(self):
+        return self.model, *self.example_inputs
```
xuzhao9, line 29: Sorry, what is `*self.example_inputs` here? Does it return the very first element of the list?

msaroufim: So I made this `*` change otherwise `test_llama_example_cuda` was broken but without it `test_llama_train/eval_cuda` were both working just fine. The intent is still to return all data but not sure how the `example` test is setup

torchbenchmark/models/llama/__init__.py
msaroufim: See my comment here https://github.com/pytorch/benchmark/pull/1446#discussion_r1132540502 - I'm a bit confused about how to proceed

msaroufim: NVM I understood what you meant and fixed this

torchbenchmark/models/llama/__init__.py
msaroufim: Fixed

torchbenchmark/models/llama/__init__.py
msaroufim: If I set batch size to anything except something like 32 I get this error

```
  File "/home/ubuntu/benchmark/run.py", line 298, in <module>
    m = Model(device=args.device, test=args.test, jit=(args.mode == "jit"), batch_size=args.bs, extra_args=extra_args)
  File "/home/ubuntu/benchmark/torchbenchmark/util/model.py", line 20, in __call__
    obj = type.__call__(cls, *args, **kwargs)
  File "/home/ubuntu/benchmark/torchbenchmark/models/llama/__init__.py", line 17, in __init__
    super().__init__(test=test, device=device, jit=jit, batch_size=batch_size, extra_args=extra_args)
  File "/home/ubuntu/benchmark/torchbenchmark/util/model.py", line 84, in __init__
    self.determine_batch_size(batch_size)
  File "/home/ubuntu/benchmark/torchbenchmark/util/model.py", line 216, in determine_batch_size
    raise NotImplementedError(f"Test {self.test} is not implemented.")
NotImplementedError: Test eval is not implemented.
```

xuzhao9 (approved):
LGTM!

torchbenchmark/models/llama/__init__.py
```diff
@@ -0,0 +1,48 @@
...
+class Model(BenchmarkModel):
+    task = NLP.LANGUAGE_MODELING
+    
+    def __init__(self, test, device, jit=False, batch_size=None, extra_args=[]):
+        super().__init__(test=test, device=device, jit=jit, batch_size=batch_size, extra_args=extra_args)
+        self.model_args = ModelArgs(vocab_size=32)
+        self.model = Transformer(self.model_args)
+
+
+        if device == "cuda":
```
xuzhao9, line 21: Oh sorry one more thing: we are supporting different device types including `cuda`, `cpu`, `mps`, and maybe more in the future, so it would be better to handle all of them like e.g., `self.model.to(self.device)`.

torchbenchmark/models/llama/__init__.py
```diff
@@ -0,0 +1,49 @@
...
+            torch.set_default_device("cuda")
+        
+        self.model.to(torch.device(device))
+        self.example_inputs = (torch.tensor([[1, 1], [1,1]], dtype=torch.int), 1)
```
xuzhao9, line 25: I think we also need to move the input data to the device
```suggestion
        self.example_inputs = (torch.tensor([[1, 1], [1,1]], dtype=torch.int).to(self.device), 1)
```

torchbenchmark/models/llama/__init__.py
```diff
@@ -0,0 +1,49 @@
...
+    task = NLP.LANGUAGE_MODELING
+    
+    def __init__(self, test, device, jit=False, batch_size=None, extra_args=[]):
+        super().__init__(test=test, device=device, jit=jit, batch_size=batch_size, extra_args=extra_args)
+        self.model_args = ModelArgs(vocab_size=32)
+        self.model = Transformer(self.model_args)
+
+
+        if device == "cuda":
+            torch.set_default_device("cuda")
```
xuzhao9, line 22: I am wondering why `torch.set_default_device('cuda')` is needed?

torchbenchmark/models/llama/tokenizer.py
```diff
@@ -0,0 +1,40 @@
...
+
+logger = getLogger()
+
+
+class Tokenizer:
+    def __init__(self, model_path: str):
+        # reload tokenizer
+        assert os.path.isfile(model_path), model_path
+        self.sp_model = SentencePieceProcessor(model_file=model_path)
+        logger.info(f"Reloaded SentencePiece model from {model_path}")
```
xuzhao9, line 18: Our customers like PT2 team usually doesn't like additional logs, maybe change the verbosity to `debug` so that it won't be printed by default?

torchbenchmark/models/llama/tokenizer.py
```diff
@@ -0,0 +1,40 @@
...
+
+        # BOS / EOS token IDs
+        self.n_words: int = self.sp_model.vocab_size()
+        self.bos_id: int = self.sp_model.bos_id()
+        self.eos_id: int = self.sp_model.eos_id()
+        self.pad_id: int = self.sp_model.pad_id()
+        logger.info(
```
xuzhao9, line 25: Same here, can we avoid logging as much as possible?

torchbenchmark/models/llama/__init__.py
xuzhao9: Thanks!

torchbenchmark/models/llama/__init__.py
```diff
@@ -0,0 +1,40 @@
...
+
+class Model(BenchmarkModel):
+    task = NLP.LANGUAGE_MODELING
+    
+    def __init__(self, test, device, jit=False, batch_size=None, extra_args=[]):
+        super().__init__(test=test, device=device, jit=jit, batch_size=batch_size, extra_args=extra_args)
+        self.model_args = ModelArgs(vocab_size=32)
+        self.model = Transformer(self.model_args)
+        
+        torch.set_default_device(device)
```
xuzhao9, line 20: I am curious why we don't need to explicitly move `self.model` and `self.example_inputs` to the device here? For example:
```
self.model = Transformer(self.model_args).to(self.device)
self.example_inputs = (torch.tensor([[1, 1], [1,1]], dtype=torch.int).to(self.device), 1)
```

msaroufim: Fixed

msaroufim:
This weeekend the internet was buzzing with news of this cpp llama implementation https://github.com/ggerganov/llama.cpp/issues/33#issuecomment-1465108022 - It seems like it might not be that tricky to also validate correctness for this model by loading in real checkpoints and tokenizer despite minor differences in the model

PR #1447, opened 2023-03-03
Add instructions to test model after submitting

msaroufim:
One more thing that's bugging me is it's not too clear how to get the tests to pass so will send out another PR soon


xuzhao9 (approved):
LGTM, thanks!

Note that depending on whether the machine local test runs on has a GPU, it might skip some tests that will be covered in the CI machine.

msaroufim:
Thanks @xuzhao9 speaking of tests, so far in this PR for example I see a lot of errors around NoneType https://github.com/pytorch/benchmark/pull/1446 - it's still my first time contributing to torchbench so I might be missing basic stuff but I feel like it'd be worth instead making it clearer what folks should do in this case because it's not quite implementing `test_my_model_cpu/cuda`

Issue #1293, opened 2022-11-09
[Roadmap] Standardize and increase coverage for TorchBench


yanbing-j:
## Motivation
`TorchBench` is a collection of open-source benchmarks used to evaluate PyTorch performance. It provides a standardized API for benchmark drivers, both for evaluation (eager/jit) and training. Plenty of popular models are involved in `TorchBench`. Users are convenient to debug and profile.

In order to standardize the performance evluation and increase coverage, `TorchBench` can be enhanced in the following 3 aspects in CPU:

- Fit for typical user scenarios
- Well integrate new features of PyTorch
- Increase benchmark coverage

### Fit for typical user scenarios (especially in `test.py`/`test_bench.py`)

add CPU runtime configuration options into launch script for user
- [ ] Add core binding option
- [ ] Add gomp/iomp option
- [ ] Add memory allocator option

support performance metrics in `test.py`/`test_bench.py`
- [ ] Add throughput: Total time / samples
- [ ] Add latency: Samples / Total time
- [ ] Add fps-like report

### Well integrate new features of PyTorch

- [ ] Enable bf16 datatype support both for inference and training
- [ ] Fully support channels_last both for inference and training
- [ ] Extend a complier option to support Dynamo
- [ ] Support JIT tracing and cover more models with JIT support
 
### Increase benchmark coverage
Increase model coverage
- [ ] Add models from community with popularity (e.g, )
- [ ] Add models from real customers (Multi-Band MelGAN, ViT and Wav2vec)

Port OpBench to TorchBench
- [ ] Increase OpBench coverage
- [ ] Complete support of dtypes, memory-format and inplace version for ops

xuzhao9:
To better help with different stakeholders, we are migrating away from `test_bench.py` to the new "userbenchmark" approach. In `benchmark/userbenchmark` (https://github.com/pytorch/benchmark/tree/main/userbenchmark), we encourage users to develop their customized benchmarks with TorchBench models and use the "run_benchmark.py" driver to drive their benchmark.

Benefits of using TorchBench userbenchmark:
- We can decouple benchmarks with their infrastructures, and run different benchmarks on different machines. For example, CPU benchmarks don't need GPU machines, and GPU benchmarks don't need too powerful CPU.
- We are also decoupling the benchmark model code with the experiments we are running on them. We believe this design is much clearer and can easily attribute the code to the correct owners.
[...]

PR #1449, opened 2023-03-03
Enabled fuser for jit backend and added fuser3 for llga path

chuanqi129:
Enabled fuser for jit backend and added fuser3 for llga path

Works for Roadmap #1293 for jit support, below is an example on CLX machine

```
$ python run.py alexnet -d cpu -m jit    -t eval
Running eval method from alexnet on cpu in jit mode with input batch size 128.
CPU Total Wall Time:  70.918 milliseconds
$ python run.py alexnet -d cpu -m jit  --fuser fuser0  -t eval
Running eval method from alexnet on cpu in jit mode with input batch size 128.
CPU Total Wall Time:  70.741 milliseconds
$ python run.py alexnet -d cpu -m jit  --fuser fuser3  -t eval
Running eval method from alexnet on cpu in jit mode with input batch size 128.
CPU Total Wall Time:  64.179 milliseconds
```


torchbenchmark/util/backends/jit.py
```diff
@@ -8,13 +8,20 @@ def parse_torchscript_args(args: List[str]) -> argparse.Namespace:
...
+    parser.add_argument("--fuser", type=str, default="", choices=["fuser0", "fuser1", "fuser2", "fuser3"], help="enable fuser")
     args, unknown_args = parser.parse_known_args(args)
     return args, unknown_args
 
 @create_backend
 def torchscript(model: 'torchbenchmark.util.model.BenchmarkModel', backend_args: List[str]):
     model.jit = True
     backend_args, extra_args = parse_torchscript_args(backend_args)
+    if model.device == "cpu" and backend_args.fuser == "fuser2":
+        raise NotImplementedError("Fuser only works with GPU.")
```
xuzhao9, line 20: In the error message, can we have the concrete fuser names like `f"{backend_args.fuser} only works with GPU`?

torchbenchmark/util/backends/jit.py
```diff
@@ -8,13 +8,20 @@ def parse_torchscript_args(args: List[str]) -> argparse.Namespace:
...
+    if model.test != "eval" and backend_args.fuser == "fuser3":
+        raise NotImplementedError("Fuser only works with eval mode.")
```
xuzhao9, line 22: Same here, it will be helpful to add the actual fuser name.

torchbenchmark/util/backends/jit.py
```diff
@@ -27,7 +34,10 @@ def _torchscript():
             module = torch.jit._script_pdt(module, example_inputs=[example_inputs, ])
         else:
             module = torch.jit.script(module, example_inputs=[example_inputs, ])
-        if model.test == "eval" and not backend_args.no_ofi:
+        if model.test == "eval" and backend_args.fuser != "fuser3" and not backend_args.no_ofi:
             module = torch.jit.optimize_for_inference(module)
+        if model.test == "eval" and backend_args.fuser == "fuser3":
```
xuzhao9, line 39: @davidberard98 can you help confirm this is the correct way of using `fuser3`?

davidberard98: it seems reasonable, but maybe it wold be better to add a separate option for `jit.freeze` since it is generally applicable for any fuser? Also, jit.freeze doesn't necessarily support all torchscript features so it's possible that it would fail for some models. Just curious @chuanqi129, is freezing required for onednn?

torchbenchmark/util/backends/jit.py
chuanqi129: Updated, thanks

torchbenchmark/util/backends/jit.py
chuanqi129: Updated

torchbenchmark/util/backends/jit.py
chuanqi129: Thanks @davidberard98, you're right, `jit.freeze` is generally applicable for any fuser with `eval` test, however, the `jit.ofi` also include `jit.freeze`, so I think we may don't need add a separate option for it. How do you think? 

No, oneDNN fuser isn't freezing required. I did some code change to merge the `jit.freeze` with `jit.ofi` into parameter `--no-ofi`, do you think it is a good solution?

```
$ python run.py alexnet -d cpu -m jit   --fuser fuser0  -t eval
Running eval method from alexnet on cpu in jit mode with input batch size 128.
CPU Total Wall Time:  70.924 milliseconds
$ python run.py alexnet -d cpu -m jit   --fuser fuser0  -t eval --no-ofi
Running eval method from alexnet on cpu in jit mode with input batch size 128.
CPU Total Wall Time:  93.563 milliseconds
$ python run.py alexnet -d cpu -m jit   --fuser fuser3  -t eval
Running eval method from alexnet on cpu in jit mode with input batch size 128.
CPU Total Wall Time:  64.205 milliseconds
$ python run.py alexnet -d cpu -m jit   --fuser fuser3  -t eval --no-ofi
Running eval method from alexnet on cpu in jit mode with input batch size 128.
CPU Total Wall Time: 101.545 milliseconds
```

Thanks @xuzhao9 & @davidberard98 for the quick pre-review comments, very appreciate~

xuzhao9 (approved):
LGTM! It is a nice improvement.

torchbenchmark/util/backends/jit.py
xuzhao9: The changes look good to me, merged. Thanks!

Issue #1444, opened 2023-03-02
Add instructpix2pix to benchmarks


msaroufim:
https://huggingface.co/timbrooks/instruct-pix2pix

PR #1451, opened 2023-03-06
Add instructpix2pix model

xuzhao9:
I place this model in the `canary_model` directory for the following reasons:
1. The model only has inference test, no train test
2. The model installation is very slow, taking ~5 mins and is downloading large model checkpoints (~10GB)
3. The inference test is slow (~4 seconds) and has noisy outputs

It needs further work to make it suitable to the core model set.

Fixes https://github.com/pytorch/benchmark/issues/1444

Test plan:
```
$ python run.py diffuser_instruct_pix2pix -d cuda
Fetching 15 files: 100%|███████████████| 15/15 [00:00<00:00, 27449.63it/s]
Running eval method from diffuser_instruct_pix2pix on cuda in eager mode with input batch size 1.
███████████████████████████████████████████| 100/100 [00:04<00:00, 23.96it/s]
GPU Time:            4123.949 milliseconds
CPU Total Wall Time: 4123.980 milliseconds
GPU 0 Peak Memory:              4.2791 GB
CPU Peak Memory:                3.7969 GB
```


msaroufim (approved):

msaroufim (approved):
Not sure about the changes to dlrm folder but everything else looks good

msaroufim:
Nice @ezyang as FYI

ezyang:
Nice!

Is there not an option to avoid downloading model checkpoints? I see that most of our benchmark models do download checkpoints, but for benchmarking purposes it seems to me like randomly initialized weights should be fine too?

ezyang:
This makes sense for detectron2, because they have data dependent flow. I guess we need to have some easy way to tell if a model is doing this or not. Dynamo can help.

PR #1457, opened 2023-03-07
Add get_optimizer and set_optimizer to BenchmarkModel API

janeyx99:
non ghstack version of #1452

PR #1459, opened 2023-03-07
Add explicit APIs in BenchmarkModel

janeyx99:
This is part 1 of adding ways to more granularly benchmark.

Also explicitly document what methods we expect BenchmarkModels to have.

### DIFF
--- a/torchbenchmark/util/model.py
+++ b/torchbenchmark/util/model.py
@@ -286,6 +286,32 @@ class BenchmarkModel(metaclass=PostInitProcessor):
                 out = self.eval()
         return out
 
+    def get_module(self):
+        raise NotImplementedError("To enroll into more granular benchmarking, every BenchmarkModel is exhorted "
+                                  "to implement a way to retrieve a torch module from it. It is okay if this "
+                                  "remains unimplemented for models that have vastly different architectures.")
+
+    # Should return loss
+    def forward(self):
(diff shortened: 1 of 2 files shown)


torchbenchmark/util/model.py
```diff
@@ -286,6 +286,30 @@ def invoke(self) -> Optional[Tuple[torch.Tensor]]:
                 out = self.eval()
         return out
 
+    def get_module(self):
+        raise NotImplementedError("To enroll into more granular benchmarking, every BenchmarkModel is exhorted "
+                                  "to implement a way to retrieve a torch module from it. It is okay if this "
+                                  "remains unimplemented for models that have vastly different architectures.")
+
+    # Should return loss
```
xuzhao9, line 294: Maybe we should clarify in the comment or error message that to implement the train loop, user can choose either implementing 1) `train()`, or 2) `forward()` `backward()` `optimizer_step()`.

It might also be good to add `get_optimizer()` and `set_optimizer()` stub here?

xuzhao9 (approved):

torchbenchmark/util/model.py
janeyx99: Good idea, I'll add that :D

janeyx99:
This will likely have to wait until https://github.com/pytorch/benchmark/pull/1457 is landed for the test cases to pass

PR #1467, opened 2023-03-08
[TEST] Ensure explicit API doesn't mess with get/set_optimizer

janeyx99:
DO NOT MERGE OR REVIEW


janeyx99:
Nice, they don't interfere!

PR #1469, opened 2023-03-08
init

janeyx99:


janeyx99:
for batches, waiting to do the optim part at the end is NOT functionally/mathematically equivalent! so don't try to split up training as such. closing as this is not really possible.

PR #1471, opened 2023-03-08
this needs to wait for the optim change

janeyx99:


janeyx99:
for batches, you cannot just split up the train into forward, backward, and optim separately (that would be the diff between gd and sgd). so closing as this is not worth doing, we should just instrument the steps ourselves

PR #1472, opened 2023-03-08
[attention is all you need] move optim step out of loop

janeyx99:
Local tests show this is equivalent, and I believe this is more correct, but I do not know.


janeyx99:
They are only equivalent cuz batchsize was 1 but are not otherwise equivalent.

PR #1473, opened 2023-03-08
Add optim.step() benchmarks

janeyx99:
In progress, but simpler and cleaner than #1338


xuzhao9 (approved):
Approve to unblock the development.

To test this, use the "T4 Metal" userbenchmark workflow, and type in "optim" as the userbenchmark name.

PR #1474, opened 2023-03-09
Remove all git-lfs files

xuzhao9:
Clean up the git-lfs files since they are causing issues in other CI workflows and local development.


wconstab:
this PR just deletes the files and the git-lfs integration.  i was expecting to also see that the files will be generated on-demand instead- is this not needed?

wconstab (approved):

PR #1478, opened 2023-03-11
Add time.sleep after subprocess git calls like git checkout 

weiwangmeta:
and git submodule operations. 

The cause of the failure for https://github.com/pytorch/benchmark/actions/runs/4195864401/jobs/7276018136 is that the git checkout 61ecaf1dd40 was tainted with an incompatible XNNPACK, e.g. 

image

Running "git submodule update --init --recursive" manually would clean up the changes. 

The reason might be because the "git submodule update --init --recursive" was executed too soon. So adding some delay to give time for the subprocess to fully finish execution.

### DIFF
--- a/torchbenchmark/util/gitutils.py
+++ b/torchbenchmark/util/gitutils.py
@@ -5,6 +5,7 @@ Utils for getting git-related information.
 
 import os
 import subprocess
+import time
 from datetime import datetime
 from typing import Optional, List
 
@@ -106,10 +107,13 @@ def checkout_git_commit(repo: str, commit: str) -> bool:
         assert len(commit) != 0
         command = f"git checkout {commit}"
         subprocess.check_call(command, cwd=repo, shell=True, stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
+        time.sleep(3)
         command = "git submodule sync"
         subprocess.check_call(command, cwd=repo, shell=True, stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
+        time.sleep(3)
         command = "git submodule update --init --recursive"
         subprocess.check_call(command, cwd=repo, shell=True, stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
+        time.sleep(3)
         return True
     except subprocess.CalledProcessError:
         print(f"Failed to checkout commit {commit} in repo {repo}")


xuzhao9:
I thought the `check_output()` call is synchronized and the parent process will wait until the subprocess finishes?
 
Maybe a better idea is to rewrite the function using gitpython, can you help try that?

weiwangmeta:
> I thought the `check_call()` call is synchronized and the parent process will wait until the subprocess finishes? (https://docs.python.org/3/library/subprocess.html#subprocess.check_call, the parent process will wait for the command to complete)
> 
> Maybe a better idea is to rewrite the function using gitpython, can you help try that?
> 
> Also, can you help attach the successful bisection workflow to this PR? You can manually trigger the bisection workflow on this branch. Curious if it finds the root cause commit of the speedup before 20230209.

Ok, this PR may not be needed but for some reason the workflow used a pytorch/benchmark version that was different from the master: 

It still used the following way: 

command = f"git submodule sync &> /dev/null && git submodule update --init --recursive &> /dev/null"
        subprocess.run(command, cwd=repo, shell=True) 

Maybe that was the issue? 
Why did the workflow check out an older version of pytorch/benchmark?

weiwangmeta:
The code on the instance was using the  v2.0 branch 
https://github.com/pytorch/benchmark/blob/v2.0/torchbenchmark/util/gitutils.py#L107 

Should I instead create the PR against v2.0 branch?

xuzhao9:
> The code on the instance was using the v2.0 branch https://github.com/pytorch/benchmark/blob/v2.0/torchbenchmark/util/gitutils.py#L107
> 
> Should I instead create the PR against v2.0 branch?

Ye,s, please do

weiwangmeta:
> > The code on the instance was using the v2.0 branch https://github.com/pytorch/benchmark/blob/v2.0/torchbenchmark/util/gitutils.py#L107
> > Should I instead create the PR against v2.0 branch?
> 
> Yes, please do

Maybe we should just cherry-pick https://github.com/pytorch/benchmark/commit/51a70428d890a26feab50569aba087d8da93b718 

Any reason why this commit was not part of 2.0?

xuzhao9:
> 
> Maybe we should just cherry-pick [51a7042](https://github.com/pytorch/benchmark/commit/51a70428d890a26feab50569aba087d8da93b718)
> 
> Any reason why this commit was not part of 2.0?

Ah, this should be a bug, I updated the main trunk bisection script, but forgot to update the 2.0 branch one.
The v2 nightly and bisection workflows all use 2.0 brank, causing this issue.

Let me cherry-pick the commit and try it again. Thanks for debugging this problem!!

xuzhao9:
Updated in v2.0 branch in https://github.com/pytorch/benchmark/commit/844578a7f622fe7743d454abc0f3fc70215f5b7b, retrying the bisection workflow at https://github.com/pytorch/benchmark/actions/runs/4395434110

weiwangmeta:
> Updated in v2.0 branch in [844578a](https://github.com/pytorch/benchmark/commit/844578a7f622fe7743d454abc0f3fc70215f5b7b), retrying the bisection workflow at https://github.com/pytorch/benchmark/actions/runs/4395434110

The above workflow failed. I reproduced it on the runner. 
Again, adding "import time; time.sleep(5)" seems to have avoided the git checkout issue.  I'm proposing this PR instead https://github.com/pytorch/benchmark/pull/1480

PR #1487, opened 2023-03-16
Expose runtime errors in the stableness test

xuzhao9:
We should treat the stableness test as sort of stress-test. We should fix any runtime errors that exposed in the stableness test.

Test workflow: https://github.com/pytorch/benchmark/actions/runs/4471135231


xuzhao9:
For unknown reasons, the GPU memory used by the subprocess worker is larger than in-process.
For example, maximum alexnet batch size is 512 when using ModelTask.

xuzhao9:
Soft_actor_critic test: https://github.com/pytorch/benchmark/actions/runs/4447958472

xuzhao9:
tacotron2 test: https://github.com/pytorch/benchmark/actions/runs/4458475180

xuzhao9:
5 Rounds workflow: https://github.com/pytorch/benchmark/actions/runs/4479875923

Issue #1486, opened 2023-03-16
issues with userbenchmark/ddp_experiments/__init__.py


ShiboXing:
Hi everyone, 

the --timeout flag doesn't seem to work due to the hard coding here: https://github.com/pytorch/benchmark/blob/355dc8002ea89f2899cc048553d5975c2ea02954/userbenchmark/ddp_experiments/__init__.py#L220

The command I'm using is `python userbenchmark/ddp_experiments/__init__.py --ngpus 4 --distributed ddp --nodes 1 --cluster local --filter_models resnet50 --timeout 2`

Additionally, would it be better if the script leaves an option flag for the network interface name: https://github.com/pytorch/benchmark/blob/355dc8002ea89f2899cc048553d5975c2ea02954/userbenchmark/ddp_experiments/__init__.py#L374 ?

please correct if I'm not using the ddp benchmark correctly.

ShiboXing:
Looks the timeout is for slurm job, is there any plan to support timeout for local?

xuzhao9:
@wconstab Can you help answer this question?

wconstab:
No plans per se, but happy to accept a PR to improve the script.  also cc @davidberard98

davidberard98:
I can open a PR

ShiboXing:
thanks everyone!

PR #1489, opened 2023-03-16
ddp_experiments script: add nccl-socket-ifname arg, fix timeout

davidberard98:
Fixes: #1486.

The timeout flag was being set incorrectly. This fixes that.

The NCCL_SOCKET_IFNAME was not configurable on the command line. This adds a --nccl-socket-ifname arg.

Testing: manually tested with:
```
python userbenchmark/ddp_experiments/__init__.py --ngpus 8 --distributed ddp --nodes 1 --filter_models resnet50 --timeout 15 --nccl-socket-ifname asdf
```

* verified that the slurm job gets killed after ~15 minutes
* in the logs, I observe `NCCL INFO NCCL_SOCKET_IFNAME set to asdf` followed by `NCCL WARN Bootstrap : no socket interface found`

### DIFF
--- a/userbenchmark/ddp_experiments/__init__.py
+++ b/userbenchmark/ddp_experiments/__init__.py
@@ -142,6 +142,12 @@ def parse_args(args: List[str]=None):
         default=None,
         help="Precision (e.g. amp, fp32, fp16)",
     )
+    parser.add_argument(
+        "--nccl-socket-ifname",
+        type=str,
+        default="ens",
+        help="Value to use for NCCL_SOCKET_IFNAME environment variable",
+    )
 
 
     try:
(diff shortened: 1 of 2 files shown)


userbenchmark/ddp_experiments/__init__.py
```diff
@@ -679,7 +685,7 @@ def main():
         slurm_exclude=args.exclude,
     )
 
-    executor.update_parameters(name="distbench", slurm_array_parallelism=1, timeout_min=1000)
+    executor.update_parameters(name="distbench", slurm_array_parallelism=1, timeout_min=args.timeout)
```
davidberard98, line 688: cc @ShiboXing I think the timeout issue was actually slightly different. The other timeouts are for things like file-based synchronization, and waiting for initialization. This timeout is for how long before we kill the slurm job. Let me know if this is the change you wanted!

ShiboXing: got it, this should fix it, thanks!

PR #1491, opened 2023-03-17
Update model task type

FindHao:
Model opacus_cifar10 is based on resnet18, so the task type should be COMPUTER_VISION.CLASSIFICATION.


xuzhao9:
I think opacus_cifar10 is mainly to test the opacus library which trains PyTorch models with differential privacy, so the task is not for image classification. cc @zou3519

FindHao:
oh, I see. Then maybe it's better to keep it in OTHERS.

FindHao:
thanks for your confirmation!

PR #1501, opened 2023-03-22
Disable CPU device on torchrec_dlrm

xuzhao9:
The CI is broken because of an upstream pytorch issue: https://github.com/pytorch/pytorch/issues/97308
Disable CPU device for now.


weiwangmeta (approved):
Oops. Will try to fix and unblock this.

PR #1508, opened 2023-03-27
[optim] Have batch_size=1 whenever possible, add more exclusions

janeyx99:
I had thought resetting the params and deleting the model would be enough, but it was not. Also added more exclusions for inductor issue https://github.com/pytorch/pytorch/issues/97361.

Will be running a test of the suite before landing this to ensure no other issues.


janeyx99:
Successful scheduled run: https://github.com/pytorch/benchmark/actions/runs/4534672192/jobs/7989157699 (I did use the -c flag to continue on error)

PR #1514, opened 2023-03-31
[optim] Move errors.txt cat'ing to the end, fix up some known issues with optim benchmarks

janeyx99:
I missed some changes when trying to exclude DALLE2_pytorch last time that were not caught as they ran into OOMs in the T4 machines. I will be trying this PR on A100s.


janeyx99:
https://github.com/pytorch/benchmark/actions/runs/4576842143

xuzhao9 (approved):
LGTM

PR #1518, opened 2023-04-03
Keep CI example generic

spzala:
The current link is broken because the nightly CI was disabled for the nvfuser benchmark and the `ci.yaml` file was removed. This can happen in the future for any specific benchmark, so let's keep the example generic.


spzala:
cc @xuzhao9

spzala:
Thanks @xuzhao9 !!

PR #1520, opened 2023-04-03
[BE] refactor S3 scripts and add new functionality for userbenchmarks

janeyx99:
Step 1 of adding a new userbenchmark regression detector

PR #1521, opened 2023-04-03
Add optim nightly benchmarking workflow

janeyx99:
Step 2 of https://github.com/pytorch/benchmark/pull/1520. This enables a nightly workflow for optim benchmarks that currently only PRINTS regressions + errors. 

The next step is to work these into actual issues.


userbenchmark/utils.py
janeyx99, line 1: No need to review this file here! It's in the base PR https://github.com/pytorch/benchmark/pull/1520

utils/s3_utils.py
janeyx99, line 1: No need to review this file here! It's in the base PR https://github.com/pytorch/benchmark/pull/1520

.github/workflows/userbenchmark-regression-detector.yml
```diff
@@ -0,0 +1,101 @@
...
+          cp -r ./.userbenchmark/optim ../benchmark-output
+      - name: Detect potential regressions
+        continue-on-error: true
+        run: |
+          . "${SETUP_SCRIPT}" && conda activate "${CONDA_ENV_NAME}"
+          pushd benchmark
+          RESULTS=($(find ${PWD}/../benchmark-output -name "metrics-*.json" -maxdepth 2 | sort -r))
+          for r in ${RESULTS[@]}; do
+            previous_metrics_json=$(python ./scripts/userbenchmark/download_previous_metrics.py -f "${r}" -p "${PLATFORM_NAME}")
+            python regression_detector.py --control "${previous_metrics_json}" --treatment "${r}"
```
xuzhao9, line 81: Need to consider the case when `previous_metrics_json` is empty.
In general it is tricky to handle exceptions in bash scripts. I feel we should have `regression_detector.py` handle S3 download.
We might want to post a GitHub issue in `pytorch/pytorch` or `pytorch/benchmark` if a regression happens.

janeyx99: Yea, I was going to get to the issue creation part after

regression_detector.py
```diff
@@ -23,13 +24,62 @@ def get_default_output_path(bm_name: str) -> str:
...
+        print(f"No performance signal detected between file {args.control} and {args.treatment}.")
+
 if __name__ == "__main__":
     parser = argparse.ArgumentParser()
     # Local metrics file comparison
     parser.add_argument("--control", default=None, help="The control group metrics file for comparison.")
     parser.add_argument("--treatment", default=None, help="The treatment metrics file for comparison.")
 
     # S3 metrics file comparison
+    parser.add_argument("--file", "-f", help="Name of the control metrics JSON file. When provided with "
```
xuzhao9, line 80: If we are comparing with the previous JSON file, the metrics file here should be treatment, not control group, correct?

I think a better interface is to directly use `--treatment` instead, and if `--control` is None, by default we should search for the most recent JSON file as the control group.

If `--treatment` specifies a local file path, we read from local file, otherwise, we try to read from Amazon S3.

janeyx99: oh yes, i meant treatment here and not control, the code has it as the treatment.

I can rework this to rely on treatment for clarity

Issue #1512, opened 2023-03-30
`doctr_det_predictor` failed installation without conda


shingjan:
with #1496, now torchbenchmark will fail the installation of `doctr_det_predictor` if `conda` is not installed. This looks like a breaking change as `conda` is explicitly mentioned as a optional, instead of required, dependency of `torchbench`.

cc: @xuzhao9

xuzhao9:
Thanks for the feedback. I am leaning to change `conda` to a hard requirement because `doctr_det_predictor` requires the `libglib` version provided in the `conda-forge` repo. Although users can install `libglib` provided by their OS, the version might not be compatible to the version accepted by `doctr_det_predictor` and will complicate the user's installation process. What do you think? 

Also cc the PR author @janeyx99

shingjan:
I feel like making `conda` a hard requirement just for one model is likely an overkill if there could be a workaround. If not then probably `conda` is the only way to go. `conda` also ships with some outdated runtimes so I will recommend to steer away from `conda` if possible.

BTW what about not failing the whole `python install.py` if some requirements of a model is not met? Like a graceful exit and warning about a specific model not available due to unmet dependencies.

janeyx99:
I think adding a warning here will be sufficient, as I understand that others like @shingjan may want to avoid conda for the reasons mentioned.

PR #1528, opened 2023-04-05
Avoid making conda a hard dependency

xuzhao9:
After some thoughts I believe we should keep conda an optional dependency because users might want to setup their own binary libraries and environment. We will first try installing binary libraries from conda, if failed, print a warning.

Fixes https://github.com/pytorch/benchmark/issues/1512.

### DIFF
--- a/torchbenchmark/models/doctr_det_predictor/install.py
+++ b/torchbenchmark/models/doctr_det_predictor/install.py
@@ -3,9 +3,6 @@ import subprocess
 import sys
 
 def pip_install_requirements():
-    # install deps from conda-forge
-    # model doctr_det_predictor needs weasyprint, which needs libglib and pango
-    subprocess.check_call(["conda", "install", "-y", "expecttest", "libglib", "pango", "-c", "conda-forge"])
     subprocess.check_call([sys.executable, '-m', 'pip', 'install', '-q', '-r', 'requirements.txt'])
 
 if __name__ == '__main__':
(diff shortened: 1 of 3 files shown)


xuzhao9:
cc @shingjan

shingjan (approved):
LGTM. Thanks for send this in!

torchbenchmark/models/doctr_det_predictor/install.py
```diff
@@ -3,9 +3,6 @@
 import sys
 
 def pip_install_requirements():
-    # install deps from conda-forge
-    # model doctr_det_predictor needs weasyprint, which needs libglib and pango
-    subprocess.check_call(["conda", "install", "-y", "expecttest", "libglib", "pango", "-c", "conda-forge"])
```
janeyx99, line 8: Could we just try and warn here instead of reverting the change entirely? maybe a
```
try:
    subprocess.check_call(["conda", "install", "-y", "expecttest", "libglib", "pango", "-c", "conda-forge"])
except Error:
    print("Failed to install weasyprint dependencies. Model doctr_det_predictor will not be able to be run.")
```

xuzhao9: I am thinking maybe we don't do this. This will make conda libraries override the libraries that already installed in user's environment.

janeyx99 (approved):

PR #1553, opened 2023-04-14
Strictly return Not Implemented for get/set_optimizer()

janeyx99:
This cannot land yet as it will now exclude certain models from the correctness check (like yolov3) which should not be skipped for things like dynamo!

@xuzhao9  has a proposal to untangle the dependencies there and this depends on landing that first.


janeyx99:
Closing for now due to staleness

PR #1554, opened 2023-04-15
Fix the issue of the dependency installation of GNN models

yanbing-j:
Add pyg-lib back as wheel is ready


yanbing-j:
PyG maintainer confirmed that the dependency wheel is compiled on PyTorch 2.0, could not be adaptive to PyTorch nightly. Will move the models to `canary_model` to disable in CI check. Will close this PR.

PR #1558, opened 2023-04-18
Run example_outputs under no_grad

Chillee:
I think example_outputs is keeping a lot of memory alive for longer than is needed.

### DIFF
--- a/torchbenchmark/util/framework/vision/model_factory.py
+++ b/torchbenchmark/util/framework/vision/model_factory.py
@@ -30,7 +30,8 @@ class TorchVisionModel(BenchmarkModel):
         self.example_inputs = (torch.randn((self.batch_size, 3, 224, 224)).to(self.device), )
         if test == "train":
             # compute loss
-            self.example_outputs = (torch.rand_like(self.model(*self.example_inputs)), )
+            with torch.no_grad():
+                self.example_outputs = (torch.rand_like(self.model(*self.example_inputs)), )
             self.model.train()
             # setup optimizer and loss_fn
             # if backend is cudagraph, must set optimizer to be capturable


xuzhao9:
Just curious - if we are loading data from dataloader, does dataloader also do this automatically for train?
I think we should keep consistent behavior with the dataloader's default way of loading data.

Chillee:
@xuzhao9 The issue isn't the dataloading, it's that we're running the model on the inputs (which then saves activations that are never freed).

xuzhao9:
> @xuzhao9 The issue isn't the dataloading, it's that we're running the model on the inputs (which then saves activations that are never freed).

Sorry let me clarify my question, I am wondering if we were running the model with inputs loaded from dataloaders, will we run the model with activations that are never freed, too?

xuzhao9:
cc @davidberard98 can you also help review? Thanks!

davidberard98 (approved):

davidberard98:
@xuzhao9 re: your comments on dataloaders - this is my mental model, can you correct me if I'm missing something?

Horace's changes here only affect the example_outputs that are created on initialization. example_outputs are only used to verify the numerical accuracy of outputs, and maybe understand the shape, strides, dtype of the output (I'm guessing this based on grepping for `example_outputs` in torchbench). Before this PR, computation of example_outputs also generates a number of activations that are saved so that the backward pass can occur; but, we don't ever run the backward pass, so saving those activations isn't important.

Meanwhile, my guess is that dataloader will be used to test the model / measure QPS for training. In that scenario, I'm guessing we would be running on an already-initialized model, and we'd be calling both the forward and backward pass. I think the activations should be freed after the backward pass in this scenario.

Does that sound correct or am I missing something?

xuzhao9:
> @xuzhao9 re: your comments on dataloaders - this is my mental model, can you correct me if I'm missing something?
> 
> Horace's changes here only affect the example_outputs that are created on initialization. example_outputs are only used to verify the numerical accuracy of outputs, and maybe understand the shape, strides, dtype of the output (I'm guessing this based on grepping for `example_outputs` in torchbench). Before this PR, computation of example_outputs also generates a number of activations that are saved so that the backward pass can occur; but, we don't ever run the backward pass, so saving those activations isn't important.
> 
> Meanwhile, my guess is that dataloader will be used to test the model / measure QPS for training. In that scenario, I'm guessing we would be running on an already-initialized model, and we'd be calling both the forward and backward pass. I think the activations should be freed after the backward pass in this scenario.
> 
> Does that sound correct or am I missing something?

Thanks for your response, @davidberard98 .
My question is, the code is trying to disable activations when `test == 'train'`. When we are running train tests, the assumption is we are running both forward and backward pass, so the activations should be saved.

When `test = 'eval'`, we don't need to save the activations, but in this case we should add the `no_grad()` context in the `eval()` function: https://github.com/pytorch/benchmark/blob/main/torchbenchmark/util/framework/timm/model_factory.py#L93

davidberard98:
@xuzhao9 I see.. but in this case, it's only affecting `example_outputs`. Since this is just for the example outputs, I don't think we can tell a difference (other than the reduced memory usage), because we only use the output tensor values (and we never need to run backward on the example_outputs)

xuzhao9:
> @xuzhao9 I see.. but in this case, it's only affecting `example_outputs`. Since this is just for the example outputs, I don't think we can tell a difference (other than the reduced memory usage), because we only use the output tensor values (and we never need to run backward on the example_outputs)

I see, yeah you are right, we should set `no_grad()` on `example_outputs`. Thanks a lot for helping me understand this!

PR #1565, opened 2023-04-22
Add option `none` to disable all metrics

FindHao:
CPU and GPU peak memory consumption measurements are enabled by default, but we don't provide an option to disable them.  By this PR, torchbench will check if `--metrics` includes `none`. If so, all metrics are removed.


xuzhao9 (approved):
LGTM. Thanks for the quick fix!

aaronenyeshi (approved):

PR #1568, opened 2023-04-22
Add new GPU arch supports for FMA count

FindHao:
Function [ConvertSMVer2Cores](https://github.com/pytorch/benchmark/blob/main/components/model_analyzer/tb_dcgm_types/gpu_device.py#L114) will return the # of FMA units of asked arch. It needs to be synced with the upstream [`_ConvertSMVer2Cores`](https://github.com/NVIDIA/cuda-samples/blob/master/Common/helper_cuda.h) file. 

This PR adds the supports for SM 8.7(Ada) SM 9.0(H100).


xuzhao9 (approved):
LGTM

aaronenyeshi (approved):

Issue #1561, opened 2023-04-20
maml_omniglot does not respect batch size


ezyang:
Found the same way as https://github.com/pytorch/benchmark/issues/1560

xuzhao9:
In maml_omniglot, we specify the variable `ALLOW_CUSTOMIZE_BSIZE = False`: https://github.com/pytorch/benchmark/blob/main/torchbenchmark/models/maml_omniglot/__init__.py#L50

This will throw error if user specifies anything different than the default batch size: https://github.com/pytorch/benchmark/blob/main/torchbenchmark/util/model.py#L224

ezyang:
Ah! In that case I think the default batch size is wrong here, it didn't match the tensor I got out

Issue #1560, opened 2023-04-20
speech_transformer does not respect batch_size


ezyang:
Even if you pass in train_bs to vary batch size (as is done in this backtrace

```
  /data/users/ezyang/c/pytorch/benchmarks/dynamo/torchbench.py(402)<module>()
-> torchbench_main()
  /data/users/ezyang/c/pytorch/benchmarks/dynamo/torchbench.py(398)torchbench_main()
-> main(TorchBenchmarkRunner(), original_dir)
  /data/users/ezyang/c/pytorch/benchmarks/dynamo/common.py(2000)main()
-> return maybe_fresh_cache(
  /data/users/ezyang/c/pytorch/benchmarks/dynamo/common.py(1017)inner()
-> return fn(*args, **kwargs)
  /data/users/ezyang/c/pytorch/benchmarks/dynamo/common.py(2377)run()
-> ) = runner.load_model(device, model_name, batch_size=batch_size)
  /data/users/ezyang/c/pytorch/benchmarks/dynamo/torchbench.py(291)load_model()
-> benchmark = benchmark_cls(
  /data/users/ezyang/c/torchbenchmark/torchbenchmark/util/model.py(19)__call__()
-> obj = type.__call__(cls, *args, **kwargs)
  /data/users/ezyang/c/torchbenchmark/torchbenchmark/models/speech_transformer/__init__.py(32)__init__()
-> self.traincfg = SpeechTransformerTrainConfig(prefetch=True, train_bs=self.batch_size, num_train_batch=NUM_TRAIN_BATCH, device=self.device)
  /data/users/ezyang/c/torchbenchmark/torchbenchmark/models/speech_transformer/config.py(73)__init__()
-> self.tr_dataset = AudioDataset(self.train_json, train_bs,
> /data/users/ezyang/c/torchbenchmark/torchbenchmark/models/speech_transformer/speech_transformer/data/data.py(84)__init__()
-> self.minibatch = minibatch
```

when you construct the dataset:

```
        self.cv_dataset = AudioDataset(self.valid_json, train_bs,
                                       self.maxlen_in, self.maxlen_out,
                                       batch_frames=self.batch_frames)
```

batch_frames is always specified, which means train_bs is ignored:

```
        # Method 2: Generate minibatch based on batch_frames
        # i.e. each batch contains approximately #batch_frames frames
        else:  # batch_frames > 0
            start = 0
            while True:
                total_frames = 0
                end = start
[...]

PR #1569, opened 2023-04-22
Fix the batch size for maml_omniglot and speech_transformer

xuzhao9:
The default batch size for maml_omniglot should be the number of tasks, which is 5.

Disable the batch size customization of speech_transformer because of upstream issue.

Fixes https://github.com/pytorch/benchmark/issues/1561 and https://github.com/pytorch/benchmark/issues/1560


ezyang (approved):
Thanks!

aaronenyeshi (approved):

Issue #1581, opened 2023-04-26
Tests (LearningToPaint, timm_vision_transformer_large) currently timing out in CI


janeyx99:
Example logs: https://github.com/pytorch/benchmark/actions/runs/4810529752/jobs/8563280755

It seems to occur consistently after #1547: https://github.com/pytorch/benchmark/actions/workflows/pr-gha-runner.yml and the PR test for that verifies that it is likely culprit.

PR #1582, opened 2023-04-26
Fix the broken CI by disabling warmup by default

xuzhao9:
We should not run warmup rounds to get eager latency when eager mode is already used.
However, in non-eager mode, we need to do the following:
1. Run correctness rounds with fp32 precision to get accurate eager output
2. Run warmup rounds with default precision to get accurate eager latency
3. Run dynamo rounds to get the correctness result

Note that the warmup rounds here are different from the warmup iterations in `run.py`. Here, we use a few warmup rounds to approximate eager latency, which is later used to calculate PT2 compilation time. In `run.py`, we run warmup rounds for performance testing.

Fix https://github.com/pytorch/benchmark/issues/1581

### DIFF
--- a/torchbenchmark/util/model.py
+++ b/torchbenchmark/util/model.py
@@ -151,7 +151,8 @@ class BenchmarkModel(metaclass=PostInitProcessor):
                     self.set_optimizer(current_optimizer)
         # apply decoration args
         apply_decoration_args(self, self.dargs)
-        eager_latency = warmup(self)
+        if should_check_correctness:
+            eager_latency = warmup(self)
         # apply optimization args
         if self.dynamo:
             from torchbenchmark.util.backends.torchdynamo import apply_torchdynamo_args


torchbenchmark/util/model.py
```diff
@@ -151,7 +151,8 @@ def __post__init__(self):
                     self.set_optimizer(current_optimizer)
         # apply decoration args
         apply_decoration_args(self, self.dargs)
-        eager_latency = warmup(self)
+        if should_check_correctness:
```
janeyx99, line 154: How does should_check_correctness become different for eager vs pt2? This seems to suggest that we do not check correctness for pt2?

janeyx99 (approved):
Stamping to unblock, but we should see green CI before landing

torchbenchmark/util/model.py
xuzhao9: We do not check correctness for eager, because we use eager as the baseline to calculate the correctness of pt2.

janeyx99: Oh sorry yes I got that flipped. Would it be good to include a comment about that somewhere?

xuzhao9: Yep I can update when we have the CI result

PR #1583, opened 2023-04-26
[optim][ub] Add more exceptions, improve -s and -m interaction

janeyx99:
This PR is another step towards subprocessing the optim benchmarks.

1. Fixes more bugs introduced in latest nightlies (yolov3 expects a batchsize that is a multiple of # GPUs).
2. Change -s and -m interaction. When a subset of models is asked for, we will run the intersection of the models asked by the user (with -m) and the subset. Previously, we just ignored -m.

Test plan:
Tested locally, and also https://github.com/pytorch/benchmark/actions/runs/4813239564/jobs/8569513495 which should ONLY run yolov3 and not fail at it.


xuzhao9 (approved):
LGTM!

PR #1584, opened 2023-04-26
[optim] Run userbenchmarks in subprocess in CI

janeyx99:
This depends on #1580 and #1583!

Runs CI in subprocesses so it should be strictly safer!


userbenchmark/optim/run_optim_benchmarks.py
```diff
@@ -0,0 +1,82 @@
...
+    # consolidate them into one file.
+    aggregated_metrics = {}
+    for file_path in Path(OUTPUT_DIR).glob("metrics*.json"):
+        with open(file_path, 'r') as f:
+            json_data = json.load(f)
+            aggregated_metrics.update(json_data['metrics'])
+    dump_output(BM_NAME, get_output_json(BM_NAME, aggregated_metrics))
+
+    # Gotta delete the tmp folder--otherwise the nightly CI will think there are multiple metrics jsons!
+    shutil.rmtree(OUTPUT_DIR)
```
xuzhao9, line 78: Oh I thought the runner Docker is transient - apparently they are not, and jobs will reuse previous job's directory.

If so, we should update the userbenchmark-regression-detector.yml file to always delete the working directory regardless the job is successful or not. It should be the CI workflow file's job to cleanup, because we shouldn't assume the benchmark can always reach this point.

xuzhao9 (approved):
Approved to unblock, see inline comments.

userbenchmark/optim/run_optim_benchmarks.py
janeyx99: Is the working directory benchmark? This should be done in another PR as it seems to have some scope. I have to remove this file anyway for the regression detector to only detect one metrics.json later down the line.

xuzhao9: Yeah sounds good to me, just reminder that if something weird happens, it may because of last failed run.

PR #1585, opened 2023-04-26
[yolov3] remove asserts when selecting cuda device

janeyx99:
We do not want model code to mess with global device stuff


xuzhao9 (approved):
LGTM!

PR #1594, opened 2023-04-27
[optim][ub] Skip most pt2 forloop impls, add option to ignore skips

janeyx99:
1. The working benchmarks are now taking very long to run. I've rootcaused it as the dynamo tracing part of for-loop optimizers, which are known to be slow (see the comment). Thus, we should not waste resources => let's skip most of them! Now, we will run pt2 on only our designated subset of models (defined by SUBSET_OF_MODELS)
2. Since we now have many exclusions and @mlazos may want to start unskipping certain things, I've added an overall option to ignore all skips. This way, one does not have to parse the skips and remove the relevant exclusion to run a certain config.

Lastly, reorder skips so that the most likely to be hit ones are first (though this probably is negligible for perf).

### DIFF
--- a/userbenchmark/optim/__init__.py
+++ b/userbenchmark/optim/__init__.py
@@ -19,6 +19,7 @@ BM_NAME: str = 'optim'
 
 continue_on_error: bool = False
 run_on_subset: bool = False
+ignore_skips: bool = False
 
 MODEL_NAMES: List[str] = list_models()
 SUBSET_OF_MODEL_NAMES: List[str] = [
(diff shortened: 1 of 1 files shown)


userbenchmark/optim/__init__.py
```diff
@@ -206,6 +207,32 @@
...
+    # moco uses DDP and DistributedDataParallel/allgather requires cuda
+    {'model': 'moco', 'device': 'cpu'},
+    # pyhpc_equation_of_state and pyhpc_isoneutral_mixing have no parameters
+    {'model': 'pyhpc_equation_of_state'},
+    {'model': 'pyhpc_isoneutral_mixing'},
+    {'model': 'pyhpc_turbulent_kinetic_energy'},
+    # fused/capturable requires params to be floats on CUDA
+    {'defaults': ['fused'], 'device': 'cpu'},
+    {'defaults': ['capturable'], 'device': 'cpu'},
+] + [
```
janeyx99, line 227: This is the new skip

albanD (approved):
SGTM

mlazos (approved):

PR #1595, opened 2023-04-27
[optim][ub] Clean benchmark workspace

janeyx99:
It turns out files persist in the workspace between runs. This PR attempts to clean the workspace before doing anything, but I'm not sure if the scope here is too narrow (we may have to clean a bigger folder).

### DIFF
--- a/.github/workflows/userbenchmark-regression-detector.yml
+++ b/.github/workflows/userbenchmark-regression-detector.yml
@@ -24,6 +24,17 @@ jobs:
         uses: actions/checkout@v3
         with:
           path: benchmark
+      - name: Clean workspace
+        shell: bash
+        run: |
+          echo "About to wipe and recreate ${GITHUB_WORKSPACE}"
+          rm -rf "${GITHUB_WORKSPACE}"
+          mkdir "${GITHUB_WORKSPACE}"
+      - name: Clean TorchBench checkout
+        run: |
+          # Remove any artifacts from the previous checkouts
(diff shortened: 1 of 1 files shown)


.github/workflows/userbenchmark-regression-detector.yml
```diff
@@ -24,6 +24,17 @@ jobs:
         uses: actions/checkout@v3
         with:
           path: benchmark
+      - name: Clean workspace
+        shell: bash
```
xuzhao9, line 28: We can move it to the end of this workflow and add `if: always()` to make sure we run it even if the previous steps fail.

xuzhao9 (commented):
LGTM. An alternative solution is to move cleanup to the end of the workflow and add `if: always()` to make sure they always run.

.github/workflows/userbenchmark-regression-detector.yml
```diff
@@ -24,6 +24,17 @@ jobs:
...
+        run: |
+          echo "About to wipe and recreate ${GITHUB_WORKSPACE}"
+          rm -rf "${GITHUB_WORKSPACE}"
+          mkdir "${GITHUB_WORKSPACE}"
+      - name: Clean TorchBench checkout
+        run: |
+          # Remove any artifacts from the previous checkouts
+          git clean -fxd
+        working-directory: benchmark
```
janeyx99, line 37:
```suggestion
```

PR #1603, opened 2023-05-01
Add regression detection steps for torch-nightly

xuzhao9:
Add regression detection feature to the torch-nightly userbenchmark.

The next step will be adding bisection.

### DIFF
--- a/.github/workflows/v3-nightly.yml
+++ b/.github/workflows/v3-nightly.yml
@@ -65,6 +65,27 @@ jobs:
           if [ -d .userbenchmark ]; then rm -Rf .userbenchmark; fi
           python run_benchmark.py torch-nightly -c v3-cuda-tests.yaml
           cp -r ./.userbenchmark/torch-nightly ../benchmark-output
+      - name: Detect potential regressions
+        continue-on-error: true
+        run: |
+          . "${SETUP_SCRIPT}" && conda activate "${CONDA_ENV_NAME}"
+          pushd benchmark
+          RESULTS=($(find ${PWD}/../benchmark-output -name "metrics-*.json" -maxdepth 2 | sort -r))
+          # TODO: the following assumes only one metrics-*.json is found. It will keep
+          # overwriting gh-issue.md if multiple are found.
+          for r in ${RESULTS[@]}; do
+            python regression_detector.py --platform "${PLATFORM_NAME}" --treatment "${r}" --owner @xuzhao9 \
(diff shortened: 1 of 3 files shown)


.github/workflows/v3-nightly.yml
```diff
@@ -65,6 +65,27 @@ jobs:
...
+          for r in ${RESULTS[@]}; do
+            python regression_detector.py --platform "${PLATFORM_NAME}" --treatment "${r}" --owner @xuzhao9 \
+            --gh-issue-path gh-issue.md --errors-path errors.txt
+          done
+      - name: Create the github issue
+        continue-on-error: true
+        if: env.TORCHBENCH_REGRESSION_DETECTED
+        uses: peter-evans/create-issue-from-file@v4
+        with:
+          title: V2 Performance Signal Detected by TorchBench Userbenchmark "torch-nightly" on ${{ env.TORCHBENCH_REGRESSION_DETECTED }}
```
janeyx99, line 85: Should this be V2 still? Or V3 haha since it says V3 nightly in the file. Also what is `TORCHBENCH_REGRESSION_DETECTED` supposed to represent?

xuzhao9: Sorry, it should be V3 instead.
In https://github.com/pytorch/benchmark/blob/main/regression_detector.py#LL134C17-L134C47, we will set TORCHBENCH_REGRESSION_DETECTED to the treatment commit hash.

userbenchmark/torch-nightly/regression_detector.py
```diff
@@ -1,17 +1,20 @@
...
 from ..utils import TorchBenchABTestResult, TorchBenchABTestMetric
 
+DEFAULT_REGRESSION_DELTA_THRESHOLD = 0.07
+
 def run(control, treatment) -> Optional[TorchBenchABTestResult]:
     control_env = control["environ"]
     treatment_env = treatment["environ"]
     control_metrics = control["metrics"]
-    treatment_metrics = control["metrics"]
+    treatment_metrics = treatment["metrics"]
```
janeyx99, line 10: OH MY GOSH WHAT A GOOD CATCH.

--facepalm that i landed this code no wonder the deltas were always 0 :'(

janeyx99 (approved):

userbenchmark/torch-nightly/regression_detector.py
xuzhao9: My bad, I didn't actually run the regression detection code when wrote this...

PR #1609, opened 2023-05-02
Detect regression based on dates

xuzhao9:
Support using `--start_date` and `--end_date` to detect regressions by using files on Amazon S3.

Test plan (Need to set AWS credentials):
```
python regression_detector.py --platform gcp_a100 --end-date 2023-05-02 --name torch-nightly
```
```
control_env:
  pytorch_git_version: 521b386e57f357386caa821adb0d9bae140f272e
treatment_env:
  pytorch_git_version: f38a04a5c5ce632ba8b0534e3689e89f98e89d73
bisection: null
details:
  test_eval[phlippe_resnet-cuda-eager]_latency:
    control: 2.978329
    treatment: 3.53091
    delta: 0.18553390172811668
  test_eval[timm_resnest-cuda-eager]_gmem:
    control: 4.0126953125
    treatment: 4.5927734375
    delta: 0.14456072036991968
control_only_metrics: {}
treatment_only_metrics: {}
```

```

TorchBench CI has detected a performance signal or runtime regression.

Base PyTorch commit: 521b386e57f357386caa821adb0d9bae140f272e

Affected PyTorch commit: f38a04a5c5ce632ba8b0534e3689e89f98e89d73

Affected Tests:
- test_eval[phlippe_resnet-cuda-eager]_latency: +0.18553%
- test_eval[timm_resnest-cuda-eager]_gmem: +0.14456%


Tests that were no longer run on affected commit:


Tests that were newly added on affected commit:


Runtime regressions found?
No runtime errors were found in the new benchmarks run--you are all good there!

GitHub workflow that triggered this issue: No URL found, please look for the failing action in https://github.com/pytorch/benchmark/actions

cc @xuzhao9
```

### DIFF
--- a/regression_detector.py
+++ b/regression_detector.py
@@ -231,7 +231,7 @@ if __name__ == "__main__":
         assert args.name, f"To detect regression with S3, you must specify a userbenchmark name."
         userbenchmark_name = args.name
         end_date = datetime.strptime(args.end_date, "%Y-%m-%d")
-    available_metrics_jsons = get_latest_jsons_in_s3_from_last_n_days(7, userbenchmark_name, args.platform, end_date)
+    available_metrics_jsons = get_latest_jsons_in_s3_from_last_n_days(userbenchmark_name, args.platform, end_date, ndays=7)
     # Download control from S3
     if len(available_metrics_jsons) == 0:
         raise RuntimeWarning(f"No previous JSONS in a week found to compare towards the end date {end_date}. No regression info has been generated.")
--- a/regression_detector.py
+++ b/regression_detector.py
@@ -176,6 +176,7 @@ def get_metrics_by_date(latest_metrics_jsons: List[str], pick_date: datetime):
         # Use the latest metric file on on the same day
         if metric_datetime.date() == pick_date.date():
             pick_metrics_json_key = metrics_json_key
+            break
     assert pick_metrics_json_key, f"Selected date {pick_date} is not found in the latest_metrics_jsons: {latest_metrics_jsons}"
     s3 = S3Client(USERBENCHMARK_S3_BUCKET, USERBENCHMARK_S3_OBJECT)
     metrics_json = s3.get_file_as_json(pick_metrics_json_key)
--- a/regression_detector.py
+++ b/regression_detector.py
@@ -10,7 +10,7 @@ import yaml
 from pathlib import Path
 import time
 from datetime import datetime
-from typing import Any, Dict, Optional
+from typing import Any, List, Dict, Optional
 from userbenchmark.utils import PLATFORMS, USERBENCHMARK_OUTPUT_PREFIX, REPO_PATH, \
                                 TorchBenchABTestResult, get_date_from_metrics, \
                                 get_ub_name, get_latest_n_jsons_from_s3
@@ -159,14 +159,26 @@ def process_regressions_into_gh_issue(regressions_dict, owner: str, output_path:
         f.write(issue_body)
 
 
-def get_best_start_date(latest_metrics_jsons, end_date: str):
+def get_best_start_date(latest_metrics_jsons, end_date: str) -> Optional[str]:
     """Get the date closest to `end_date` from `latest_metrics_jsons`"""
-    pass
+    end_datetime = datetime.strptime("%Y%m%d", end_date)
+    for metrics_json in latest_metrics_jsons:
+        start_datetime = datetime.strptime(metrics_json.split('/')[-1].split('-')[-1].split('.')[0], '%Y%m%d%H%M%S')
+        if start_datetime < end_datetime:
+            return datetime.strftime("%Y%m%d", start_datetime)
+    return None
 
 
-def get_metrics_by_date(date_str: str):
+def get_metrics_by_date(latest_metrics_jsons: List[str], date_str: str):
+    chosen_metrics_json_key: Optional[str] = None
+    for metrics_json in latest_metrics_jsons:
+        metric_datetime = datetime.strftime(datetime.strptime(metrics_json.split('/')[-1].split('-')[-1].split('.')[0],
+                                                            '%Y%m%d%H%M%S'), '%Y%m%d')
+        if metric_datetime == date_str:
+            chosen_metrics_json_key = metrics_json
+    assert chosen_metrics_json_key, f"Selectd date {date_str} is not found in the latest_metrics_jsons: {latest_metrics_jsons}"
     s3 = S3Client(USERBENCHMARK_S3_BUCKET, USERBENCHMARK_S3_OBJECT)
-    metrics_json_file = s3.get_file_as_json(latest_metrics_jsons[date_str])
+    metrics_json_file = s3.get_file_as_json(chosen_metrics_json_key)
     with open(metrics_json_file, "r") as metrics_fp:
         return json.load(metrics_fp)
 
--- a/regression_detector.py
+++ b/regression_detector.py
@@ -13,7 +13,7 @@ from datetime import datetime
 from typing import Any, List, Dict, Optional
 from userbenchmark.utils import PLATFORMS, USERBENCHMARK_OUTPUT_PREFIX, REPO_PATH, \
                                 TorchBenchABTestResult, get_date_from_metrics, \
-                                get_ub_name, get_latest_n_jsons_from_s3
+                                get_ub_name, get_latest_n_day_jsons_from_s3
 from utils.s3_utils import S3Client, USERBENCHMARK_S3_BUCKET, USERBENCHMARK_S3_OBJECT
 
 GITHUB_ISSUE_TEMPLATE = """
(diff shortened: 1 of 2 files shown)


regression_detector.py
```diff
@@ -158,6 +159,31 @@ def process_regressions_into_gh_issue(regressions_dict, owner: str, output_path:
         f.write(issue_body)
 
 
+def get_best_start_date(latest_metrics_jsons, end_date: str) -> Optional[datetime]:
+    """Get the date closest to `end_date` from `latest_metrics_jsons`"""
+    end_datetime = datetime.strptime("%Y-%m-%d", end_date)
+    for metrics_json in latest_metrics_jsons:
+        start_datetime = datetime.strptime(metrics_json.split('/')[-1].split('-')[-1].split('.')[0], '%Y%m%d%H%M%S')
```
janeyx99, line 166: We should probably abstract out a function for this by now haha, maybe something called get_date_from_metrics_filename

```
metrics_json.split('/')[-1].split('-')[-1].split('.')[0]
```

regression_detector.py
```diff
@@ -158,6 +159,31 @@ def process_regressions_into_gh_issue(regressions_dict, owner: str, output_path:
         f.write(issue_body)
 
 
+def get_best_start_date(latest_metrics_jsons, end_date: str) -> Optional[datetime]:
```
janeyx99, line 162:
```suggestion
def get_best_start_date(latest_metrics_jsons: List[str], end_date: str) -> Optional[datetime]:
```
is this the right type annotation?

regression_detector.py
```diff
@@ -158,6 +159,31 @@ def process_regressions_into_gh_issue(regressions_dict, owner: str, output_path:
...
+
+def get_metrics_by_date(latest_metrics_jsons: List[str], pick_date: datetime):
+    pick_metrics_json_key: Optional[str] = None
+    for metrics_json in latest_metrics_jsons:
+        metric_datetime = datetime.strptime(metrics_json.split('/')[-1].split('-')[-1].split('.')[0],
+                                                            '%Y%m%d%H%M%S')
+        # Use the latest metric file on on the same day
+        if metric_datetime.date() == pick_date.date():
+            pick_metrics_json_key = metrics_json
+    assert pick_metrics_json_key, f"Selectd date {pick_date} is not found in the latest_metrics_jsons: {latest_metrics_jsons}"
```
janeyx99, line 180:
```suggestion
    assert pick_metrics_json_key, f"Selected date {pick_date} is not found in the latest_metrics_jsons: {latest_metrics_jsons}"
```

regression_detector.py
```diff
@@ -158,6 +159,31 @@ def process_regressions_into_gh_issue(regressions_dict, owner: str, output_path:
...
+            return start_datetime
+    return None
+
+
+def get_metrics_by_date(latest_metrics_jsons: List[str], pick_date: datetime):
+    pick_metrics_json_key: Optional[str] = None
+    for metrics_json in latest_metrics_jsons:
+        metric_datetime = datetime.strptime(metrics_json.split('/')[-1].split('-')[-1].split('.')[0],
+                                                            '%Y%m%d%H%M%S')
+        # Use the latest metric file on on the same day
```
janeyx99, line 177: In this case, would it be more effective to reverse the latest_metrics_jsons (or traverse it in reverse order) and just return when the first hit occurs?

^ definitely a nit

regression_detector.py
```diff
@@ -158,6 +159,31 @@ def process_regressions_into_gh_issue(regressions_dict, owner: str, output_path:
...
+        if metric_datetime.date() == pick_date.date():
+            pick_metrics_json_key = metrics_json
+    assert pick_metrics_json_key, f"Selectd date {pick_date} is not found in the latest_metrics_jsons: {latest_metrics_jsons}"
+    s3 = S3Client(USERBENCHMARK_S3_BUCKET, USERBENCHMARK_S3_OBJECT)
+    metrics_json_file = s3.get_file_as_json(pick_metrics_json_key)
```
janeyx99, line 182: This should just be a json already and not a file

regression_detector.py
```diff
@@ -188,30 +215,41 @@ def process_regressions_into_gh_issue(regressions_dict, owner: str, output_path:
...
-        process_regressions_into_yaml(regressions_dict, output_path, args.control, args.treatment)
-
-        owner = " ".join(args.owner) if args.owner else "@xuzhao9"
-        process_regressions_into_gh_issue(regressions_dict, owner, args.gh_issue_path, args.errors_path)
     else:
-        # S3 path
-        print("Comparison for metrics from Amazon S3 given a userbenchmark name is WIP.")
+        assert args.name, f"To detect regression with S3, you must specify a userbenchmark name."
+        end_date = datetime.strptime(args.end_date, "%Y-%m-%d")
+        userbenchmark_name = args.name
```
janeyx99, line 236: more intuitive to switch these lines given the assertion above
```suggestion
        userbenchmark_name = args.name
        end_date = datetime.strptime(args.end_date, "%Y-%m-%d")
```

userbenchmark/utils.py
```diff
@@ -85,27 +85,27 @@ def get_output_dir(bm_name) -> Path:
     return target_dir
 
 
-def get_latest_n_jsons_from_s3(n: int, bm_name: str, platform_name: str, date: str) -> List[str]:
-    """Retrieves the most recent n metrics json filenames from S3 the WEEK BEFORE the given date, exclusive of that date.
-       If fewer than n items are found, returns all found items without erroring, even if there were no items. """
+def get_latest_n_day_jsons_from_s3(ndays: int, bm_name: str, platform_name: str, date: datetime, limit: int=100) -> List[str]:
+    """Retrieves the most recent n day metrics json filenames from S3 the WEEK BEFORE the given date, inclusive of that date.
```
janeyx99, line 89: should be modified to not be WEEK BEFORE

regression_detector.py
```diff
@@ -188,30 +215,41 @@ def process_regressions_into_gh_issue(regressions_dict, owner: str, output_path:
...
+    available_metrics_jsons = get_latest_n_day_jsons_from_s3(7, userbenchmark_name, args.platform, end_date)
```
janeyx99, line 237: The 7 here is not super intuitive.

Could we have ndays in get_latest_n_day_jsons_from_s3 be a kwarg that defaults to 7?

regression_detector.py
```diff
@@ -188,30 +215,41 @@ def process_regressions_into_gh_issue(regressions_dict, owner: str, output_path:
...
+    # Download control from S3
+    if len(available_metrics_jsons) == 0:
+        raise RuntimeWarning(f"No previous JSONS in a week found to compare towards the end date {end_date}. No regression info has been generated.")
+    print(f"Found metrics json files on S3: {available_metrics_jsons}")
+    start_date = args.start_date if args.start_date else get_best_start_date(available_metrics_jsons, end_date)
```
janeyx99, line 242: I think the start date should be derived earlier in the code (before retrieving jsons. Then, the ndays could be calculated from start date, and the random 7 from above would not need to occur unless the start date was not specified.

regression_detector.py
xuzhao9: Good point!

janeyx99:
What's the test plan?

userbenchmark/utils.py
```diff
@@ -85,27 +85,32 @@ def get_output_dir(bm_name) -> Path:
...
 
-def get_latest_n_jsons_from_s3(n: int, bm_name: str, platform_name: str, date: str) -> List[str]:
-    """Retrieves the most recent n metrics json filenames from S3 the WEEK BEFORE the given date, exclusive of that date.
-       If fewer than n items are found, returns all found items without erroring, even if there were no items. """
+def get_date_from_metrics_s3_key(metrics_s3_key: str) -> datetime:
+    metrics_s3_json_filename = metrics_s3_key.split('/')[-1]
+    return datetime.strptime(metrics_s3_json_filename, 'metrics-%Y%m%d%H%M%S.json')
+
+
+def get_latest_n_day_jsons_from_s3(ndays: int, bm_name: str, platform_name: str, date: datetime, limit: int=100) -> List[str]:
```
janeyx99, line 93:
```suggestion
def get_latest_jsons_in_s3_from_last_n_days(ndays: int, bm_name: str, platform_name: str, date: datetime, limit: int=100) -> List[str]:
```
would be a clearer name

regression_detector.py
xuzhao9: Yes it is correct, will fix.

regression_detector.py
xuzhao9: Nice catch! Thanks!

regression_detector.py
xuzhao9: Sorry I don't get this, I believe `latest_metrics_jsons ` is already sorted in the reverse order?

regression_detector.py
xuzhao9: We use the user-specified start_date if that's available. Otherwise, we will try to derive it with the retrieved jsons.

Without retrieving jsons, we don't know which date are available (the workflow could fail on some date).

regression_detector.py
janeyx99: Oh, in this case, shouldn't you break after a value is found? Otherwise, if multiple jsons are found on the same day, won't the earlier ones overwrite the key?

janeyx99 (approved):
Looks okay if passes local S3 expectations after addressing my comment about the reverse order traversal

regression_detector.py
xuzhao9: Ah, I see what you mean. Yes, a loop break is needed here. Thanks for pointing it out!

xuzhao9:
> Looks okay if passes local S3 expectations after addressing my comment about the reverse order traversal

Thanks! I have fixed that issue.

xuzhao9:
@janeyx99 Can I get a stamp on the Diff? Thanks!

janeyx99:
Done!

PR #1613, opened 2023-05-03
Remove downloading the nightly, check only if the workflow is scheduled

xuzhao9:
On A10G, we pull the torchbench:latest docker and run the tests in the docker.
On GCP A100 instances, the runner is already in the docker, so we can just skip the pytorch installation.

Test plan of v3 test workflow: https://github.com/pytorch/benchmark/actions/runs/4899087111

Test plan of A100 userbenchmark: https://github.com/pytorch/benchmark/actions/runs/4898890711


.github/workflows/userbenchmark-a100.yml
```diff
@@ -24,34 +24,19 @@ jobs:
...
-          sudo python benchmark/utils/cuda_utils.py --setup-cuda-softlink
       - name: Tune Nvidia GPU
         run: |
           . "${SETUP_SCRIPT}"
           sudo nvidia-smi -pm 1
           sudo nvidia-smi -ac 1215,1410
           nvidia-smi
-      - name: Install PyTorch nightly
+      - name: Check PyTorch nightly if scheduled
+        if: ${{ github.event_name == 'schedule' }}
```
janeyx99, line 34: Won't this just not install any nightly nor docker if the event was manual (not scheduled)? Should the step be duplicated below with an 
```
if: ${{ github.event_name != 'schedule' }}
```

xuzhao9: The nightly has already been installed in docker. If it is not 'schedule', we can assume the nightly has already been installed, but I can add a step below to make sure the versions are consistent.

THe installation step is at https://github.com/pytorch/benchmark/blob/main/docker/torchbench-nightly.dockerfile#L22

PR #1614, opened 2023-05-03
Use docker image for optim userbenchmark workflow

janeyx99:
Pro: no more infra issues

Con: will not be able to tell if we're using the most recent nightly

My thoughts: at this early stage, I'd like to just see more green and have stable benchmarks before worrying too much about upstream dependencies.

Test plan:
https://github.com/pytorch/benchmark/actions/runs/4886316447/jobs/8721515824 :D it worked

### DIFF
--- a/.github/workflows/userbenchmark-regression-detector.yml
+++ b/.github/workflows/userbenchmark-regression-detector.yml
@@ -9,7 +9,7 @@ on:
       userbenchmark_options:
         description: "Option of the user benchmark to run"
 env:
-  CONDA_ENV_NAME: "userbenchmarks-ci"
+  CONDA_ENV_NAME: "torchbench"
   PLATFORM_NAME: "gcp_a100"
   TORCHBENCH_USERBENCHMARK_SCRIBE_GRAPHQL_ACCESS_TOKEN: ${{ secrets.TORCHBENCH_USERBENCHMARK_SCRIBE_GRAPHQL_ACCESS_TOKEN }}
   AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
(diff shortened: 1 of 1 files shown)


.github/workflows/userbenchmark-regression-detector.yml
```diff
@@ -24,33 +24,19 @@ jobs:
...
-          sudo python benchmark/utils/cuda_utils.py --setup-cuda-softlink
       - name: Tune Nvidia GPU
         run: |
           . "${SETUP_SCRIPT}"
           sudo nvidia-smi -pm 1
           sudo nvidia-smi -ac 1215,1410
           nvidia-smi
-      - name: Install PyTorch nightly
+      - name: Check PyTorch nightly if scheduled
+        if: ${{ github.event_name == 'schedule' }}
```
xuzhao9, line 34: This line means, if the workflow is scheduled by GitHub, it will check the torch nightly package is the latest, and will fail if it is not (currently it will fail because torchvision build failure).

If the workflow is triggered by user (like clicking on the run workflow button), it will skip the check, and use whatever the latest docker version available.

janeyx99: Yea...I had blindly copied the code forgetting that this was not the docker pull step. Will be adding that in.

xuzhao9: See: https://github.com/pytorch/test-infra/blob/xz9/add-gcp-autoscaler/gcp/k8s/runnerdeployment.yaml#L10

.github/workflows/userbenchmark-regression-detector.yml
```diff
@@ -24,33 +24,12 @@ jobs:
...
-          . "${SETUP_SCRIPT}"
-          python benchmark/utils/python_utils.py --create-conda-env "${CONDA_ENV_NAME}"
-          sudo python benchmark/utils/cuda_utils.py --setup-cuda-softlink
       - name: Tune Nvidia GPU
         run: |
           . "${SETUP_SCRIPT}"
           sudo nvidia-smi -pm 1
           sudo nvidia-smi -ac 1215,1410
           nvidia-smi
-      - name: Install PyTorch nightly
```
xuzhao9, line 38: This means you are never checking the PyTorch nightly version in the docker even if it is staled.
Say if pytorch nightly fails to build for 10 days, you will keep running this workflow for 10 days on the same pytorch version. Just want to confirm is this desired behavior?

janeyx99: Yes. I'd still want signal since it is relatively early for the optim benchmarks so if regressions happen it would be good to note regardless of the pytorch nightly.

PR #1620, opened 2023-05-08
Add cm3leon

ezyang:


torchbenchmark/models/cm3leon/__init__.py
```diff
@@ -0,0 +1,39 @@
...
+        self.example_inputs = (
+            torch.randint(1, vocab_size, (self.batch_size, prompt_size)).to(self.device),
+        )
+
+    def get_module(self):
+        return self.model, self.example_inputs
+
+    # The code included here is specialized for eval
+    def train(self):
+        return NotImplementedError("training script not published")
```
msaroufim, line 35: Depending on how much you wanna make stuff up, you can have some sanity training script - I did this for segment anything https://gist.github.com/msaroufim/3091538604760282502019503f45f4be but it makes me feel gross

ezyang: I'd prefer not to make stuff up, because I don't want to optimize a training pipeline that isn't one that anyone uses! haha

torchbenchmark/models/cm3leon_generate/__init__.py
```diff
@@ -0,0 +1,38 @@
...
+
+    def get_module(self):
+        return self.model, self.example_inputs
+
+    # The code included here is specialized for eval
+    def train(self):
+        return NotImplementedError("training script not published")
+
+    def eval(self):
+        out = self.model(*self.example_inputs)
```
xuzhao9, line 37: Do we need `with torch.no_grad()` context here to disable gradients in inference code?

torchbenchmark/models/cm3leon_generate/model.py
```diff
@@ -0,0 +1,1203 @@
...
+        attn_probs = self.dropout_module(attn_weights)
+
+        assert v is not None
+        attn = torch.bmm(attn_probs, v)
+        assert list(attn.size()) == [bsz * self.num_heads, tgt_len, self.head_dim]
+        attn = attn.transpose(0, 1).contiguous().view(tgt_len, bsz, embed_dim)
+        attn = self.out_proj(attn)
+        return attn, None  # To match return type of F.multi_head_attention_forward
+
+    @torch.jit.export
```
xuzhao9, line 508: Same here, can we remove `@torch.jit.export`?

xuzhao9 (approved):
LGTM!

xuzhao9 (approved):
LGTM!

xuzhao9:
Is there an existing upstream repository of this model we can reference to? If yes, we should put it in the `origin` file in this model directory, like in https://github.com/pytorch/benchmark/blob/main/torchbenchmark/models/DALLE2_pytorch/origin

xuzhao9:
> There isn't yet; this is some code extracted from metaseq from Armen Aghajanyan that I got permission to put in torchbench, but the official release hasn't happened yet. When that happens we should point this at the official code.

I see, it is fine not adding this file then.

PR #1622, opened 2023-05-08
use g++ 10 for torchrec_dlrm

xuzhao9:
Use the consistent version with https://github.com/pytorch/FBGEMM/blob/main/.github/scripts/setup_env.bash#L102

### DIFF
--- a/docker/gcp-a100-runner-dind.dockerfile
+++ b/docker/gcp-a100-runner-dind.dockerfile
@@ -1,4 +1,5 @@
 # default base image: summerwind/actions-runner-dind:latest
+# base image: Ubuntu 20.04
 ARG BASE_IMAGE=summerwind/actions-runner-dind:latest
 FROM ${BASE_IMAGE}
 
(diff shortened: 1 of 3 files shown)


desertfire (commented):
Can we use the same `ln` trick as in https://github.com/pytorch/FBGEMM/blob/main/.github/scripts/setup_env.bash#L102? The steps here can be a good instructions to follow on our dev machine, where we can't `sudo apt`.

xuzhao9:
> Can we use the same `ln` trick as in https://github.com/pytorch/FBGEMM/blob/main/.github/scripts/setup_env.bash#L102? The steps here can be a good instructions to follow on our dev machine, where we can't `sudo apt`.

We can add this to the Docker env setup, however, it also means that we will be using the conda compiler and libraries, and we need to setup `LIBRARY_PATH`, `LD_LIBRARY_PATH`, and `PATH` correctly accordingly.

If we do not need to compile pytorch, this shouldn't be needed.

xuzhao9:
@desertfire Can you help take another look? Thanks!

PR #1625, opened 2023-05-08
Add hf_GPT2_generate and hf_T5_generate

ezyang:


voznesenskym (approved):

torchbenchmark/util/framework/huggingface/model_factory.py
```diff
@@ -146,3 +159,43 @@ def eval(self) -> Tuple[torch.Tensor]:
...
+        generation_config = GenerationConfig(
+            max_new_tokens=256,
+            pad_token_id=0,
+            eos_token_id=None,
+            do_sample=False,
+            num_beams=1,
+            use_cache=True,
+        )
+        self.model = GenerationWrapper(self.model, generation_config)
+
```
xuzhao9, line 186: It seems to me we should add `def train(self): raise NotImplementError()` here, because the old training loop wouldn't make sense with Generation model.

xuzhao9 (approved):
LGTM overall (see inline comments about training loop though).
Also, I am wondering if batch size == 1 makes sense on A100 though. What is the default batch size used in the upstream HuggingFace repository?

torchbenchmark/util/framework/huggingface/model_factory.py
ezyang: yes, will add

ezyang:
> Also, I am wondering if batch size == 1 makes sense on A100 though. What is the default batch size used in the upstream HuggingFace repository?

This is, uh, an interesting problem.

HuggingFace doesn't have a default "size" per se; this is not like a training script. If you're a hobbyist, you are definitely running autoregressive generation with batch size 1, since you're going to only run one prompt at a time. Most of the docs will look like this too.

What if you're running a serving service? Well, now it gets a bit tricky. In principle, you could batch multiple prompts together. But this is... kind of tricky to do? Because even if you have enough request volume that you can profitably batch, chances are all the prompts are varying sizes so you need some sort of nested tensor thing going on with the attention component... having a fixed size batch across everything is unrealistic unless you're padding (but padding is reaaaally bad).

I agree that for expert prod users, batch size = 1 is unrealistic. But I think I'd rather be a little more sensitive to overhead since there are real users who would be in this regime, and then figure out a more realistic batched benchmark setup.

xuzhao9:
@ezyang 
Yeah usually we should use the batch size number from upstream repo, or the original paper.
Previously we have a script to enumerate batch sizes [1, 2, 4,... 2^n], and pick the one with the highest GPU hardware FLOPS for inference tests. I hope we could make it a CI to run automatically on new models...

ezyang:
@xuzhao9 

> Previously we have a script to enumerate batch sizes [1, 2, 4,... 2^n], and pick the one with the highest GPU hardware FLOPS for inference tests. I wish we could make it a CI to run automatically on new models...

Isn't this... bad lol. It's cherry picking a batch size that makes us look good haha

PR #1627, opened 2023-05-09
Delay the nightly and daily workflows by 2 hours

xuzhao9:
In practice we found we need to wait for another 2 hours before all nightly packages release, so we also need to delay the nightly docker build and the workflows.

Also, use forked conda env in the workflows, so that the consequent workflows won't use the same environment.

Test optim workflow: https://github.com/pytorch/benchmark/actions/runs/4929351284


janeyx99 (commented):
Could you also modify the optim regression workflow as a part of this PR?

janeyx99 (approved):

PR #1631, opened 2023-05-10
[llama] Dont fail on training with random weights

anijain2305:
Enables testing torch.compile with random weights.


xuzhao9 (approved):
LGTM, thanks for the fix!

PR #1632, opened 2023-05-10
Add filter example to the pytest-benchmark driver section

spzala:
The`run.py` doesn't have `-k` flag. The filter example was originally created for the `pytest-benchmark driver`, where it works as expected. It seems like at one point the `run.py` section was added mistakenly before (instead of after) the filter example.


spzala:
cc @xuzhao9

xuzhao9 (approved):
Nice catch!

PR #1633, opened 2023-05-10
Add some better GNN benchmarks from pytorch_geometric

ezyang:
Unlike the ones in canaries, these work with only a source install of
latest pytorch_geometric, without requiring C++ libraries.  I've also
taken care to make the models jittable, so that you actually get speedup
from torch.compile them.

### DIFF
--- a/torchbenchmark/util/framework/gnn/model_factory.py
+++ b/torchbenchmark/util/framework/gnn/model_factory.py
@@ -1,9 +1,12 @@
 import torch
+import sys
 import typing
 from contextlib import nullcontext
 from torchbenchmark.util.model import BenchmarkModel
 
-from torch_geometric.nn import GAT, GCN, GraphSAGE
+import torch_geometric
+from torch_geometric.nn import GAT, GCN, GraphSAGE, GIN, EdgeCNN
+from torchbenchmark.tasks import GNN
 import torch.nn.functional as F
 from tqdm import tqdm
 from pathlib import Path
@@ -15,6 +18,8 @@ from torch import Tensor
 models_dict = {
     'gat': GAT,
     'gcn': GCN,
+    'edgecnn': EdgeCNN,
+    'gin': GIN,
     'sage': GraphSAGE,
 }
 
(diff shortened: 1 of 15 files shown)


ezyang:
cc @yanbing-j

torchbenchmark/util/framework/gnn/model_factory.py
```diff
@@ -107,3 +112,40 @@ def eval(self) -> typing.Tuple[torch.Tensor]:
...
+        # model jittable also probably reduces its performance; but this is
+        # matching the benchmark)
+        self.model = sys.modules["torch_geometric.compile"].to_jittable(self.model)
+        num_nodes, num_edges = 10_000, 200_000
+        x = torch.randn(num_nodes, 64, device=device)
+        edge_index = torch.randint(num_nodes, (2, num_edges), device=device)
+        self.example_inputs = (x, edge_index)
+
+    def eval(self):
+        return (self.model(*self.example_inputs),)
```
xuzhao9, line 141: Should we add `torch.no_grad()` context here?
What about `self.model.eval()`?

torchbenchmark/util/framework/gnn/model_factory.py
```diff
@@ -107,3 +112,40 @@ def eval(self) -> typing.Tuple[torch.Tensor]:
...
+
+    def train(self):
+        # NB: This is a little different than test_basic_gnn.py, as we
+        # are including the cost of randn_like in the overall computation here
+        out = self.model(*self.example_inputs)
+        out_grad = torch.randn_like(out)
```
xuzhao9, line 147: Here our benchmark latency will also include the time we generate the output tensor, but I guess it is fine if the tensor is relatively small.

Optionally we can generate `self.out_grad` in `__init__()` and reference it as `self.out_grad` here, then we won't include tensor generation in the measurement.

xuzhao9 (approved):

xuzhao9 (approved):

torchbenchmark/util/framework/gnn/model_factory.py
ezyang: I think it is more realistic to include the random inside the benchmark

torchbenchmark/util/framework/gnn/requirements.txt
```diff
@@ -0,0 +1 @@
+torch_geometric @ git+https://github.com/pyg-team/pytorch_geometric.git@cabcd4097442ba60aa1efa11e1619dd9bb8fb527
```
yanbing-j, line 1: Is it suitable to fix a certain commit here? Why not use `pip install pyg-nightly`?

yanbing-j:
This PR is to add benchmarks from PyG without NeighborLoader and real data. Overall looks good to me.

torchbenchmark/util/framework/gnn/requirements.txt
ezyang: I figured fixing a commit would make things more stable as we wouldn't arbitrarily break as people pushed things to the repo. Can't use the nightly as it's incompatible with PyTorch nightly

PR #1639, opened 2023-05-11
[optim][ub] add newly models to dense list to not run on sparseadam

janeyx99:
Fixes https://github.com/pytorch/benchmark/actions/runs/4943819025


xuzhao9 (approved):
LGTM, although we might add even more new models in the next few weeks...

Maybe consider using allowlist to only allow a list of models?

PR #1652, opened 2023-05-15
Use CUDA 11.8 versions as the default

xuzhao9:
Use CUDA 11.8 as the default as CUDA 11.7 is being deprecated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,19 +21,19 @@ conda create -n torchbenchmark python=3.10
 conda activate torchbenchmark
 ```
 
-If you are running NVIDIA GPU tests, we support CUDA 11.7+, and use CUDA 11.7 as default:
+If you are running NVIDIA GPU tests, we support CUDA 11.8+, and use CUDA 11.8 as default:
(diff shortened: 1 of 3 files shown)


xuzhao9:
qq @atalman: when are we deprecating CUDA 11.7 ?

spzala (commented):
@xuzhao9 a small review comment :) -  I think we should update the README (which currently says, `we support CUDA 11.7+, and use CUDA 11.7 as default` ) to change the default to 11.8 as well? https://github.com/pytorch/benchmark#installation Thanks!

xuzhao9:
> @xuzhao9 a small review comment :) - I think we should update the README (which currently says, `we support CUDA 11.7+, and use CUDA 11.7 as default` ) to change the default to 11.8 as well? https://github.com/pytorch/benchmark#installation Thanks!

Addressed - thanks for pointing it out!

spzala:
@xuzhao9 yrw, thank you!

PR #1656, opened 2023-05-16
Fix batch size for vision_maskrcnn

ezyang:


xuzhao9 (approved):
Stamped. I have a few comments though:

1. does this mean we can't tune the batch size for detectron2_maskrcnn models either?
2. it will be better to get comments from torchvision developers for this PR.

xuzhao9:
cc @fmassa

xuzhao9:
I have got response from @terrychenism that detectron2 accepts any batch size inputs, so this should be a limitation only for vision_maskrcnn.

xuzhao9:
@ezyang Can you give a code example? To me it seems we can pass more tensors in a list to the model easily.

PR #1660, opened 2023-05-17
Update README.md

H-Huang:
typo

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ pip install --pre torch torchvision torchtext torchaudio -f https://download.pyt
 
 Install other necessary libraries:
 ```
-pip install pyyaml pytest
+pip install pyyaml
 ```
 
 Install the benchmark suite, which will recursively install dependencies for all the models.  Currently, the repo is intended to be installed from the source tree.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -38,7 +38,7 @@ pip install --pre torch torchvision torchtext torchaudio -f https://download.pyt
 
 Install other necessary libraries:
 ```
-pip install pyyaml
+pip install pyyaml pytest
```
xuzhao9, line 41: If `distro` is also missing, we should add it here too.

xuzhao9 (approved):

README.md
H-Huang: Actually I noticed that pytest, distro, etc. are included in requirements.txt. Not sure how come it wasn't installed as part of `python install.py`. I will double check

H-Huang: seems like its my fault and the initial install.py didnt succeed for me. Now i see the packages. I'll remove the package update and only check in the typo change

xuzhao9: Sounds good to me, thanks!

xuzhao9:
Hi @H-Huang , can you help Import this diff to fbsource so that we can land it?

PR #1674, opened 2023-05-22
Add metrics param for cpu userbenchmark

chuanqi129:
Add `metrics` param for cpu userbenchmark, easy to extend others metrics in later. 

Works for roadmap #1293

### DIFF
--- a/userbenchmark/cpu/run_config.py
+++ b/userbenchmark/cpu/run_config.py
@@ -9,17 +9,14 @@ import numpy
 from typing import List, Dict, Optional
 from pathlib import Path
 
-from cpu_utils import add_path, REPO_PATH, get_output_dir, get_output_json, dump_output
+from cpu_utils import add_path, REPO_PATH, validate, parse_str_to_list, list_metrics, get_output_dir, get_output_json, dump_output
 with add_path(str(REPO_PATH)):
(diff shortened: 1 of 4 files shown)


userbenchmark/cpu/run_config.py
```diff
@@ -43,21 +40,19 @@ def dump_result_to_json(metrics, output_dir):
...
         # We do not allow RuntimeError in this test
         try:
-            # load the model instance within separate subprocess
-            model = load_model_isolated(config)
+            if "cpu_peak_mem" in metrics:
+                # load the model instance within separate subprocess
+                model = load_model_isolated(config)
+            else:
+                # load the model instance within current process
+                model = load_model(config)
```
xuzhao9, line 55: Curious why using `load_model` here?
We should try to avoid loading model instance within the same process. Different model code might interfere with each other, so it will be better to isolate them by running models in separate processes.

chuanqi129: The mainly reason is that the `load_model_isolated` can't show verbose / log when we set the `ONEDNN_VERBOSE` or `PYTORCH_JIT_LOG_LEVEL` and other verbose / log flags. Those flags can help us to double check whether we run into right pass or get more debug information. How do you think?

PR #1679, opened 2023-05-23
Upload the regression file to Amazon S3

xuzhao9:
When a regression happens, we should also upload the regression detector output to S3.

Test plan:

Local testing:
```
$ python ./scripts/userbenchmark/upload_s3.py --userbenchmark_json $LATEST_REGRESSION_RESULT --userbenchmark_platform "gcp_a100"
```

### DIFF
--- a/scripts/userbenchmark/upload_s3.py
+++ b/scripts/userbenchmark/upload_s3.py
@@ -1,8 +1,6 @@
 import argparse
-import json
 import sys
 from pathlib import Path
-from datetime import datetime
 
 REPO_ROOT = Path(__file__).parent.parent.parent.resolve()
 class add_path():
@@ -24,7 +22,8 @@ with add_path(str(REPO_ROOT)):
 
 def upload_s3(ub_name: str, platform_name: str, date_str: str, file_path: Path):
     """S3 path:
-        s3://ossci-metrics/torchbench_userbenchmark/<userbenchmark-name>/<platform-name>/<date>/metrics-<time>.json"""
+        s3://ossci-metrics/torchbench_userbenchmark/<userbenchmark-name>/<platform-name>/<date>/metrics-<YYmmddHHMMSS>.json
+        s3://ossci-metrics/torchbench_userbenchmark/<userbenchmark-name>/<platform-name>/<date>/regression-<YYmmddHHMMSS>.yaml"""
     s3client = S3Client(USERBENCHMARK_S3_BUCKET, USERBENCHMARK_S3_OBJECT)
     prefix = f"{ub_name}/{platform_name}/{date_str}"
     s3client.upload_file(prefix=prefix, file_path=file_path)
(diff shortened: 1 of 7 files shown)


scripts/userbenchmark/upload_s3.py
```diff
@@ -34,7 +33,7 @@ def upload_s3(ub_name: str, platform_name: str, date_str: str, file_path: Path):
     parser.add_argument("--userbenchmark_platform", required=True,
                         help='Name of the userbenchmark platform')
     parser.add_argument("--userbenchmark_json", required=True,
-                        help='Upload userbenchmark json data')
+                        help='Upload userbenchmark json or regression yaml file.')
```
janeyx99, line 36:
```suggestion
                        help='Upload userbenchmark json and regression yaml file.')
```
I forget what this flag does now...is it just a True/False whether we upload stats?

janeyx99 (approved):
Seems fine though I don't know if it'd be worth removing the slight duplication in get_date_from_metrics.

scripts/userbenchmark/upload_s3.py
xuzhao9: It specifies the file we want to upload to S3. Previously it should be just metrics-*.json file, now we want it to upload regression-*.yaml files as well.

janeyx99: Hmmmm would this be clearer as just "--file-to-upload" or "--files-to-upload" then?

userbenchmark/utils.py
```diff
@@ -68,15 +69,26 @@ def dump_output(bm_name, output, target_dir: Path=None) -> None:
...
-    return datetime.strftime(datetime_obj, "%Y-%m-%d")
-
+    if metrics_file.startswith("metrics-"):
+        datetime_obj = datetime.strptime(metrics_file, "metrics-%Y%m%d%H%M%S")
+        return datetime.strftime(datetime_obj, "%Y-%m-%d")
+    elif metrics_file.startswith("regression-"):
+        datetime_obj = datetime.strptime(metrics_file, "regression-%Y%m%d%H%M%S")
+        return datetime.strftime(datetime_obj, "%Y-%m-%d")
+    print(f"Unknown metrics or regression file name format: {metrics_file}")
+    exit(1)
```
janeyx99, line 79: Right, but we could refactor like:
```suggestion
    assert metrics_file.startswith("metrics-") or metrics_file.startswith("regression-"), f"Unknown metrics or regression file name format: {metrics_file}"
    # metrics_file usually looks like metrics-%Y%m%d%H%M%S or regression-%Y%m%d%H%M%S
    stripped_filename = metrics_file.split("-")[1]
    datetime_obj = datetime.strptime(metrics_file.split("-")[1], "%Y%m%d%H%M%S")
    return datetime.strftime(datetime_obj, "%Y-%m-%d")
```

xuzhao9: Right, this would be cleaner. Thanks!

scripts/userbenchmark/upload_s3.py
xuzhao9: Right, let me change it to "--upload-file".

PR #1681, opened 2023-05-23
[optim][ub] Set threshold to 30% and disable torch-nightly unstable models

janeyx99:
Skip models that are deemed unstable by torch-nightly entirely. Currently, that would be the following 16: 
```
['LearningToPaint', 'densenet121', 'detectron2_fasterrcnn_r_101_c4', 'detectron2_fasterrcnn_r_101_fpn', 'detectron2_fasterrcnn_r_50_c4', 'detectron2_fasterrcnn_r_50_c4', 'detectron2_fcos_r_50_fpn', 'detectron2_maskrcnn_r_50_c4', 'detectron2_maskrcnn_r_50_fpn', 'drq', 'functorch_dp_cifar10', 'functorch_maml_omniglot', 'lennard_jones', 'resnext50_32x4d', 'soft_actor_critic', 'timm_efficientdet']
```

Also drive the regression threshold to 30%!!! This _is_ a big leap, but we want to stop flooding torchbench with issues everyday and slowly push down the threshold. Based on stats in https://docs.google.com/spreadsheets/d/1r3a7UczDofej8p6KhAcbTf8moxnlbyQ1AWZKOo0kbh8/edit#gid=418753160, only 0.8% of noise will go beyond 30%, so this should make the issues much more manageable in conjunction with my nifty dashboard (unfortunately internal only atm) www.fburl.com/optim_benchmarks

### DIFF
--- a/userbenchmark/optim/__init__.py
+++ b/userbenchmark/optim/__init__.py
@@ -1,5 +1,5 @@
 from pathlib import Path
-from typing import Any, Dict, List, Tuple
+from typing import Any, Dict, List, Set, Tuple
 from torchbenchmark import load_model_by_name
 import torch
 from torch import _dynamo as torchdynamo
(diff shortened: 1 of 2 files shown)


xuzhao9 (approved):
LGTM, just note that the stableness in v3-cuda-tests.yaml is on the test level, not model level.

userbenchmark/optim/__init__.py
```diff
@@ -22,6 +23,19 @@
...
 
+# Models that are unstable in torch-nightly should not run in the optim world either
+def get_unstable_models() -> List[str]:
+    unstable_models: List[str]= []
+    yaml_file_path = REPO_PATH.joinpath('userbenchmark/torch-nightly/v3-cuda-tests.yaml')
+    with open(yaml_file_path, "r") as yf:
+        config_obj = yaml.safe_load(yf)
+    for d in config_obj['cuda']:
+        if not d['stable']:
+            unstable_models.append(d['model'])
```
xuzhao9, line 34: We may want to remove the duplicate ones because the original yaml file is on the test level, not model level.
So there might be multiple entries for a single model (for example, one entry for cuda_train and another for cuda_eval).

albanD (approved):

PR #1685, opened 2023-05-24
Support downloading yaml files from S3

xuzhao9:
In the bisection workflow, we need to download the regression yaml file from S3.

Test Plan:
```
python regression_detector.py --name torch-nightly --platform gcp_a100 --end-date 2023-05-19 --download-from-s3 --output 1.yaml
Downloaded the regression yaml file to path 1.yaml
```

### DIFF
--- a/regression_detector.py
+++ b/regression_detector.py
@@ -196,7 +196,7 @@ if __name__ == "__main__":
     parser.add_argument("--start-date", default=None, help="The start date to detect regression.")
     parser.add_argument("--end-date", default=None, help="The latest date to detect regression.")
     # download from S3
-    parser.add_argument("--download-from-s3", help="Download the regression yaml file from S3.")
+    parser.add_argument("--download-from-s3", action='store_true', help="Download the regression yaml file from S3.")
     # output file path
     parser.add_argument("--output", default=None, help="Output path to print the regression detection file.")
     # GitHub issue details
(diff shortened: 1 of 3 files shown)


regression_detector.py
```diff
@@ -195,6 +195,8 @@ def get_metrics_by_date(latest_metrics_jsons: List[str], pick_date: datetime):
     parser.add_argument("--platform", choices=PLATFORMS, default=None, help="The name of platform of the regression.")
     parser.add_argument("--start-date", default=None, help="The start date to detect regression.")
     parser.add_argument("--end-date", default=None, help="The latest date to detect regression.")
+    # download from S3
+    parser.add_argument("--download-from-s3", action='store_true', help="Download the regression yaml file from S3.")
```
janeyx99, line 199: If one chooses the option to download regression from S3, do we expect them to NOT specify treatment and control then? If so, can we specify that this is disjoint and that when specifying this option, control and treatment should not be specified and start-date and end-date should not be? 

The usual actions would create a regression file for reporting. What do we expect this option to be used for?

xuzhao9: This option is expected for the bisection workflow. The bisection workflow will need to get the regression detector output file as the input, for example:
```
python regression_detector.py --name torch-nightly --platform gcp_a100 --end-date 20230519 --download-from-s3 --output regression-20230519182382.yaml
python bisection.py --config regression-20230519182382.yaml
```

I feel it is easier to add downloading in `regression_detector.py` than in `bisection.py`.

xuzhao9: Correct - if users only want to download the regression result from S3, they don't need to specify control or treatment (they will be ignored).

janeyx99: Could the script be more documentative of this use case? e.g., specify that this is useful for bisection

janeyx99 (approved):
Approving to unblock, though I feel as if the help message could be more helpful.

regression_detector.py
xuzhao9: Sorry I missed this, will fix in a follow-up PR.

PR #1702, opened 2023-05-28
Add gpt-j-6b model into torchbench

chuanqi129:
Add HF model gpt-j-6b into torchbench

Work for roadmap #1293 

```shell
$ python run.py hf_GPTJ --torchdynamo inductor
Running eval method from hf_GPTJ on cpu in dynamo inductor mode with input batch size 1 and precision fp32.
CPU Total Wall Time: 3891.941 milliseconds
CPU Peak Memory:               47.6953 GB
PT2 Compilation time:      99.554 seconds
```

### DIFF
--- a/torchbenchmark/models/hf_GPTJ/__init__.py
+++ b/torchbenchmark/models/hf_GPTJ/__init__.py
@@ -1,0 +1,10 @@
+from torchbenchmark.tasks import NLP
+from torchbenchmark.util.framework.huggingface.model_factory import HuggingFaceModel
+
+class Model(HuggingFaceModel):
+    task = NLP.LANGUAGE_MODELING
+    DEFAULT_TRAIN_BSIZE = 1
+    DEFAULT_EVAL_BSIZE = 1
+
+    def __init__(self, test, device, jit=False, batch_size=None, extra_args=[]):
+        super().__init__(name="hf_GPTJ", test=test, device=device, jit=jit, batch_size=batch_size, extra_args=extra_args)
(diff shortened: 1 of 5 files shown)


torchbenchmark/models/hf_GPTJ/__init__.py
```diff
@@ -0,0 +1,10 @@
+from torchbenchmark.tasks import NLP
+from torchbenchmark.util.framework.huggingface.model_factory import HuggingFaceModel
+
+class Model(HuggingFaceModel):
+    task = NLP.LANGUAGE_MODELING
+    DEFAULT_TRAIN_BSIZE = 1
```
xuzhao9, line 6: Is this the correct batch size? We should use the default batch size in the upstream code if possible.

torchbenchmark/models/hf_GPTJ/metadata.yaml
```diff
@@ -0,0 +1,13 @@
+eval_benchmark: false
+eval_deterministic: false
+eval_nograd: true
+train_benchmark: false
+train_deterministic: false
+not_implemented:
+  # hf_GPTJ model doesn't support JIT
+  - jit: true
```
xuzhao9, line 8: Let's remove `jit` option since it has been deprecated in the default CI.

torchbenchmark/models/hf_GPTJ/__init__.py
```diff
@@ -0,0 +1,10 @@
...
+    DEFAULT_EVAL_BSIZE = 1
+
+    def __init__(self, test, device, jit=False, batch_size=None, extra_args=[]):
```
xuzhao9, line 9: Same here, remove the `jit=False` and `jit=jit` arguments.

chuanqi129:
Close it due to out-of-date

PR #1725, opened 2023-06-09
Move torchrec_dlrm to canary models

xuzhao9:
We are moving torchrec_dlrm to canary models for a couple of reasons:

1. The environment setup is difficult. It requires fbgemm_gpu_nightly, which doesn't work on older pytorch releases.
2. Doesn't support bisection. To bisect this model, we need to re-compile fbgemm_gpu, which takes very long time.


janeyx99 (approved):

desertfire:
cc @anijain2305

PR #1727, opened 2023-06-11
Add throughputs metric and enable in cpu userbenchmark

chuanqi129:
Add throughputs metric in torchbench model metrics, enable throughputs metric in cpu userbenchmark.

Work for roadmap #1293 to add fps-like report in cpu userbenchmark.


xuzhao9 (approved):
LGTM

PR #1730, opened 2023-06-12
Add lit-llama benchmarks (logits, autoregressive generation, lora fine tuning)

ezyang:


torchbenchmark/models/lit_llama_lora/__init__.py
```diff
@@ -0,0 +1,64 @@
...
+        )
+
+
+    def get_module(self):
+        return self.model, self.example_inputs
+
+    def train(self):
+        logits = self.model(*self.example_inputs)
+        logits.sum().backward()
+        # meh this sucks
```
msaroufim, line 58: xd, this might be a good dataset https://huggingface.co/datasets/OpenAssistant/oasst1

Even finetuning on two examples of questions you make up might be not bad as a sanity

torchbenchmark/models/lit_llama_generate/__init__.py
```diff
@@ -0,0 +1,41 @@
...
+        # max_new_tokens matches lit-llama/generate.py
+        self.example_inputs = (tokenizer.encode("The meaning of life is", bos=True, eos=False, device=device), 50)
+
+    def train(self):
+        return NotImplementedError("cannot train on autoregressive generation")
+
+    def eval(self):
+        self.model.eval()
+        with torch.no_grad():
+            y = self.model(*self.example_inputs)
```
msaroufim, line 40: do you mind printing the input prompt and the output, will be nice to do vibe checks later

torchbenchmark/models/lit_llama_generate/__init__.py
```diff
@@ -0,0 +1,41 @@
...
+    def forward(self, idx, max_new_tokens):
+        return lit_llama_generate.generate(self.model, idx, max_new_tokens)
+
+class Model(lit_llama.Model):
+    def __init__(self, test, device, jit=False, batch_size=None, extra_args=[]):
+        super().__init__(test=test, device=device, jit=jit, batch_size=batch_size, extra_args=extra_args)
+        self.model = GenerationWrapper(self.model)
+        tokenizer = Tokenizer(os.path.join(LIT_LLAMA_PATH, "checkpoints/lit-llama/tokenizer.model"))
+        # max_new_tokens matches lit-llama/generate.py
+        self.example_inputs = (tokenizer.encode("The meaning of life is", bos=True, eos=False, device=device), 50)
```
msaroufim, line 32: is 50 the max number of tokens to generate?

msaroufim (approved):
Some minor nits otherwise thanks! Let's see how long CI will take now lol

xuzhao9 (approved):
LGTM. How large is the checkpoint file and is there any rules on the accessing frequency? If we download it too frequently (every CI workflow and every nightly testing workflow), the server might ban our access.

msaroufim:
@xuzhao9 this will be a common workflow for LLM work, it might make sense to cache these files in a github artifact

torchbenchmark/models/lit_llama_generate/__init__.py
ezyang: yes

ezyang:
Do we have any precedent for hosting it in S3? I am happy to set it up if there is some example of doing it.

xuzhao9:
> Do we have any precedent for hosting it in S3? I am happy to set it up if there is some example of doing it.

I am not sure if it requires legal review for that.
The closest example I found is that Detectron2 models host their checkpoints at https://dl.fbaipublicfiles.com: https://github.com/pytorch/benchmark/blob/main/torchbenchmark/util/framework/detectron2/__init__.py#L12

ezyang:
So, it seems like we can only run this benchmark on the A100s anyway, so I'm going to disable the A10G configuratoin

torchbenchmark/models/lit_llama/__init__.py
```diff
@@ -0,0 +1,54 @@
...
+from torchbenchmark.tasks import NLP
+import torch
+import os
+from torchbenchmark import REPO_PATH
+import sys
+import lightning as L
+
+LIT_LLAMA_PATH = os.path.join(REPO_PATH, "submodules", "lit-llama")
+
+sys.path.insert(0, LIT_LLAMA_PATH)
```
xuzhao9, line 11: Is it better to use the `add_path()` context manager here (https://github.com/pytorch/benchmark/blob/ae3534ab071270d2d3bb02305c2a1d5089eeb169/torchbenchmark/__init__.py#L27)? If people are running 2 models in the same process, this might cause unexpected results.

torchbenchmark/models/lit_llama/__init__.py
```diff
@@ -0,0 +1,54 @@
...
+
+from lit_llama.utils import EmptyInitOnDevice, lazy_load, llama_model_lookup
+from lit_llama import LLaMA, Tokenizer
+
+class Model(BenchmarkModel):
+    task = NLP.LANGUAGE_MODELING
+    DEFAULT_EVAL_BSIZE = 1
+    DEFAULT_TRAIN_BSIZE = 32
```
xuzhao9, line 19: Curious why the default train batch size is 32?

xuzhao9 (approved):
LGTM, see minor inline comments

torchbenchmark/models/lit_llama/__init__.py
ezyang: okey. Note that there are a bunch of other models doing it the wrong way:

```
torchbenchmark/canary_models/fambench_dlrm/__init__.py:        sys.path.insert(0, self.path)
torchbenchmark/models/attention_is_all_you_need_pytorch/preprocess.py:        sys.path.insert(0, path)
torchbenchmark/models/fambench_xlmr/__init__.py:        sys.path.insert(0, self.path)
torchbenchmark/models/pytorch_CycleGAN_and_pix2pix/scripts/edges/batch_hed.py:sys.path.insert(0, caffe_root + 'python')
torchbenchmark/models/pytorch_CycleGAN_and_pix2pix/scripts/eval_cityscapes/cityscapes.py:        sys.path.insert(0, '{}/scripts/helpers/'.format(self.dir))
torchbenchmark/models/tacotron2/waveglow/mel2samp.py:sys.path.insert(0, 'tacotron2')
```

torchbenchmark/models/lit_llama/__init__.py
ezyang: I think I should just delete this, it's meaningless, you can't train 7B without some sort of distribution haha

torchbenchmark/models/lit_llama/__init__.py
xuzhao9: For fambench_dlrm and fambench_xlmr, they are part of the `add_path` context manager:  https://github.com/pytorch/benchmark/blob/main/torchbenchmark/canary_models/fambench_dlrm/__init__.py#L16

For other models (attention_is_all_you_need_pytorch, pytorch_CycleGAN_and_pix2pix, and tacotron2), I believe those code are not executed by the model (for example, they are preprocessing code)

ezyang:
oh thank god, pr-test is finally passing

torchbenchmark/models/lit_llama_generate/__init__.py
ezyang: Hmm, but I don't want to print it here, because then the detokenization would also count as part of the benchmark?

torchbenchmark/models/lit_llama_lora/__init__.py
ezyang: Will fix this later, I think. Not needed for dynamo benchmarks.

Issue #1717, opened 2023-06-06
install.py installs datasets for all models, even if model to install is specified


ezyang:
Steps to reproduce: Run `python install.py asdfjksdfd`

xuzhao9:
Some models are sharing the same input sets like coco2017, and if all models need to specify their input data, we need to think about how to avoid downloading the same input data files repeatedly.

Currently https://github.com/pytorch/benchmark/pull/1734 only checks the existence of the file and directories without checking their integrity, which can be future work.

PR #1734, opened 2023-06-15
Download the input data ondemand

xuzhao9:
Instead of always downloading all `.tar.gz` files, only download them when the model explicitly ask for them.

Fixes https://github.com/pytorch/benchmark/issues/1717

Note that this is a trade-off - some models share the same data set, and we don't want to download the same files multiple times. Here we are only checking the existence of the file and directory. An improvement could be using checksums or completion flag files.


ezyang (approved):
Thanks!

PR #1736, opened 2023-06-16
SAM medeval finetuning loop

msaroufim:


torchbenchmark/models/sam/__init__.py
```diff
@@ -47,14 +84,72 @@ def get_module(self):
...
+        self.model.train()
+        optimizer = torch.optim.Adam(self.model.mask_decoder.parameters(), lr=1e-5, weight_decay=0)
+        seg_loss = monai.losses.DiceCELoss(sigmoid=True, squared_pred=True, reduction='mean')
+
+        num_epochs = 100
+        losses = []
+        best_loss = 1e10
+
+        train_dataset = NpzDataset(npz_tr_path)
+        train_dataloader = DataLoader(train_dataset, batch_size=64, shuffle=True)
```
xuzhao9, line 108: We shouldn't involve any DataLoader code in `train()` or `eval()`. The principle is to prefetch just 1 batch of input in `__init__()`, and run it in `train()`.

torchbenchmark/models/sam/__init__.py
```diff
@@ -47,14 +84,72 @@ def get_module(self):
...
+        for epoch in range(num_epochs):
+            epoch_loss = 0
+            # train
+            for step, (image_embedding, gt2D, boxes) in enumerate(tqdm(train_dataloader)):
+                # do not compute gradients for image encoder and prompt encoder
+                with torch.no_grad():
```
xuzhao9, line 114: It is suggested to use the new `staged train` interface if possible, users implement 3 interfaces for train: `forward()`, `backward()`, and `optimizer_step()`,

msaroufim: You mean the same interface that the optimizer tests are using in the e2e benchmark folder?

xuzhao9: @msaroufim Not really, I am talking about these interfaces in core models: https://github.com/pytorch/benchmark/blob/main/torchbenchmark/util/model.py#L251-L258

PR #1744, opened 2023-06-23
Add --setup-only to install.py to prevent env. change during benchmark

thiagocrepaldi:
Although installing all requirements and configuring the torch benchmark in a single run is convenient for brand new environments, other times we already have existing environment which had all dependencies properly consolidated accros many projects, including pytorch, torchvision, torchaudio, torchtext, detectron2, benchmark and others

By running `pip install requirements.txt` a second time, package versions can be changed, breaking the current environment.

This PR introduces a flag that prevents `pip install` from running while keeping the `setup` phase intact. 

The current behavior (without the new flag) is not changed to preserve BC

### DIFF
--- a/install.py
+++ b/install.py
@@ -31,6 +31,7 @@ if __name__ == '__main__':
     parser.add_argument("--continue_on_fail", action="store_true")
     parser.add_argument("--verbose", "-v", action="store_true")
     parser.add_argument("--component", choices=["distributed"], help="Install requirements for optional components.")
+    parser.add_argument("--setup-only",  action="store_true", help="Only runs the setup routine without installing packages from requirements.txt")
     args = parser.parse_args()
 
     os.chdir(os.path.realpath(os.path.dirname(__file__)))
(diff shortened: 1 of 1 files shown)


install.py
```diff
@@ -44,21 +45,29 @@ def pip_install_requirements(requirements_txt="requirements.txt"):
...
-        sys.exit(0)
 
-    success, errmsg = pip_install_requirements()
-    if not success:
-        print("Failed to install torchbenchmark requirements:")
-        print(errmsg)
-        if not args.continue_on_fail:
-            sys.exit(-1)
     from torchbenchmark import setup
     success &= setup(models=args.models, verbose=args.verbose, continue_on_fail=args.continue_on_fail, test_mode=args.test_mode, allow_canary=args.canary)
```
xuzhao9, line 72: I think this is incorrect - even when running with `--setup-only`, the `setup()` function will still install all the requirements.txt files in each models directory, such as https://github.com/pytorch/benchmark/blob/main/torchbenchmark/util/framework/detectron2/requirements.txt and https://github.com/pytorch/benchmark/blob/main/torchbenchmark/models/drq/requirements.txt

xuzhao9 (commented):
The dependency files are scattered in the models/ and framework/ directories. With `--setup-only` option, it will still try to reinstall packages that might already exist in the environment.

install.py
thiagocrepaldi: where/when these extra requirements.txt are installed? They seem to be harmless as they are just a few and not pinning any particular version (which is the issue we want to prevent). I will double check whether we can improve the `--setup-only` by preventing these extra requirements too

xuzhao9: > where/when these extra requirements.txt are installed? They seem to be harmless as they are just a few and not pinning any particular version (which is the issue we want to prevent). I will double check whether we can improve the `--setup-only` by preventing these extra requirements too

Unfortunately, there are plenty of them and many of them are pinning to particular versions. Almost every model comes with its own `requirements.txt` file, to name a few:
https://github.com/pytorch/benchmark/blob/main/torchbenchmark/models/BERT_pytorch/requirements.txt
https://github.com/pytorch/benchmark/blob/main/torchbenchmark/models/soft_actor_critic/requirements.txt
https://github.com/pytorch/benchmark/blob/main/torchbenchmark/models/tacotron2/requirements.txt#L6
https://github.com/pytorch/benchmark/blob/main/torchbenchmark/models/fastNLP_Bert/requirements.txt
https://github.com/pytorch/benchmark/blob/main/torchbenchmark/models/doctr_det_predictor/requirements.txt

The reason we have to pin the versions of the dependencies is because some model code are old and there's no update for ~1year, therefore, they won't work on the latest dependency code and it is non-trivial to migrate. However, our main customer (PyTorch team) still want to keep those and test against them.

thiagocrepaldi: I am running a full test to check whether the "default models"(when we don't set `--models`) break an existing env

Do you know where I can find the code that would follow and install the `requirements.txt` file above? maybe this PR can go a step further and skip the installation of the model-specific requirements

xuzhao9: Yes that is another option, the code is at https://github.com/pytorch/benchmark/blob/main/torchbenchmark/__init__.py#L149

thiagocrepaldi:
Thanks, will think about how to proceed. Closing for now

PR #1745, opened 2023-06-23
Add boto3 to requirements.txt

thiagocrepaldi:
This PR adds boto3 to the requirements.txt as install.py will error out without it installed.

The README.md list boto3 and pyyaml as packages to be installed before running, but requirements.txt only has pyyaml in the list


xuzhao9 (approved):
LGTM, thanks!

PR #1752, opened 2023-07-06
Add special device mapping

FindHao:
To reuse the inference batch sizes already set in models' meta.yaml file for different kind of devices like AMD GPUs, this PR adds a `SPECIAL_DEVICE_MAPPING` and got the corresponding batch size when executes `determine_batch_size ()`

## Run
You can use the following command to run model `resnet18` inference by this userbenchmark. 
```
python run_benchmark.py rocm-test  -t eval -m resnet18 
```
You can run all models without `-m resnet18`.
## Results
The result json file is generated in the location `benchmark/.userbenchmark/rocm-test/metrics-20230706XX.json`. The content is like the following.
```json
{
    "name": "rocm-test",
    "environ": {
        "pytorch_git_version": "e9ebda29d87ce0916ab08c06ab26fd3766a870e5",
        "device": "AMD Instinct MI210"
    },
    "metrics": {
        "test_eval[resnet18-cuda-eager]_latency": 23.336431
    }
}
```
## Acknowlegement
This userbenchmark's code is mostly copied from `userbenchmark/torch-nightly`, some functionalities may not be available even though there are related codes.

### DIFF
--- a/userbenchmark/torch-nightly/run.py
+++ b/userbenchmark/torch-nightly/run.py
@@ -224,5 +224,8 @@ def run(args: List[str]):
             score_name = f"{score_version}_score"
             metrics[score_name] = score
         result = get_output_json(BM_NAME, metrics)
+        if args.device == 'cuda':
+            import torch
+            result["environ"]["device"] = torch.cuda.get_device_name()
         with open(args.output, 'w') as f:
             json.dump(result, f, indent=4)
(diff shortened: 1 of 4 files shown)


userbenchmark/torch-nightly/run.py
```diff
@@ -224,5 +224,8 @@ def run(args: List[str]):
             score_name = f"{score_version}_score"
             metrics[score_name] = score
         result = get_output_json(BM_NAME, metrics)
+        if args.device == 'cuda':
+            import torch
+            result["environ"]["device"] = torch.cuda.get_device_name()
```
FindHao, line 229: @xuzhao9 do you still want to keep this?

xuzhao9: Yes, let's keep this.

xuzhao9:
Can we have an example command and result json in the PR body? For example, the result of just running a single model.

FindHao:
> Can we have an example command and result json in the PR body? For example, the result of just running a single model.

updated.

xuzhao9 (approved):
Thanks!

PR #1753, opened 2023-07-07
[backends] Add functionality to TRT backend

gs-olive:
- Add argument parsing for backend arguments to be passed to Torch-TRT
- Add capability to specify IR and other Torch-TRT fields via command line interface
- Add functionality to compilation path and clean up code

cc: @narendasan

### DIFF
--- a/torchbenchmark/util/backends/trt.py
+++ b/torchbenchmark/util/backends/trt.py
@@ -114,11 +114,6 @@ def torch_trt(
         module, example_inputs = model.get_module()
         torch_dtype_precision = torch.half if FP16 else torch.float32
 
-        trt_input = [
-            torch_tensorrt.Input(shape=input_.shape, dtype=input_.dtype)
-            for input_ in example_inputs
-        ]
-
         print(
             f"Compiling {model.name} with batch size {model.batch_size}, precision {model.dargs.precision}, "
             + f"and {'default' if 'ir' not in torch_trt_kwargs else torch_trt_kwargs['ir']} IR"
@@ -126,7 +121,7 @@ def torch_trt(
 
         trt_module = torch_tensorrt.compile(
             module,
-            inputs=trt_input,
+            inputs=example_inputs,
             enabled_precisions={torch_dtype_precision},
             **torch_trt_kwargs,
         )
--- a/torchbenchmark/util/backends/trt.py
+++ b/torchbenchmark/util/backends/trt.py
@@ -1,8 +1,30 @@
 from typing import List
 import torch
+import argparse
 
 from torchbenchmark.util.backends import create_backend 
 from torchbenchmark.util.env_check import is_hf_model
+
+def parse_torch_trt_args(backend_args: List[str]):
+    """Parses CLI-provided backend arguments to extract Torch-TRT keywords
+
+    Returns kwargs dictionary and remainder arguments which were unrecognized
+    """
+    arg_parser = argparse.ArgumentParser()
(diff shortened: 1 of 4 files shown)


xuzhao9 (approved):
LGTM. Just curious will you be interested in setting up a CI to maintain this?

gs-olive:
Thank you for the review. I think we would be interested in setting up a CI to maintain it. @narendasan - what are your thoughts on this?

gs-olive:
@xuzhao9 - what would you need from us to set up a CI to maintain the TRT backend?

xuzhao9:
@gs-olive Here is the guide on how to develop a CI: https://github.com/pytorch/benchmark/blob/main/userbenchmark/ADDING_USERBENCHMARKS.md

gs-olive:
Hi @xuzhao9 - thanks for the reference. I've added functionality to this PR in a new commit which adds the `userbenchmark` folder and adapts functionality from the main [`run.py`](https://github.com/pytorch/benchmark/blob/745644f391b4d11da107b2c82fe2d7a3eacf561d/run.py). I also added a `ci.yaml` file. I wasn't sure how the benchmarks are invoked in CI, however. My intent was to have this sort of usage:

```bash
# Invokes a singular model
python run_benchmark.py torch_trt --model resnet18 --precision fp32 --bs 4

# Invokes all models in the directory
python run_benchmark.py torch_trt --precision fp32 --bs 4
```

torchbenchmark/util/backends/trt.py
```diff
@@ -31,27 +73,63 @@ def _fx2trt():
...
-        else:
-            torchtrt_dtype = torch_tensorrt.dtype.float
-            torch_dtype = torch.float32
-        trt_input = [torch_tensorrt.Input(shape=example_inputs[0].shape, dtype=torch_dtype)]
-        trt_module = torch_tensorrt.compile(module, inputs=trt_input, enabled_precisions=torchtrt_dtype)
+        torch_dtype_precision = torch.half if FP16 else torch.float32
+
+        trt_input = [
+            torch_tensorrt.Input(shape=input_.shape, dtype=input_.dtype)
+            for input_ in example_inputs
```
xuzhao9, line 119: Is it necessary to add check that the type of `example_inputs` is `List[tensor]`? Actually, many models have different types of inputs.

gs-olive:
@xuzhao9 - Thank you for the detailed comments and review. This sounds like a good idea to me; I have removed the `ci.yaml` file as requested, and I will follow up with a subsequent PR addressing the remaining comments.

xuzhao9 (approved):
LGTM, please address my inline comment about checking the input type of `example_inputs`, thanks!

torchbenchmark/util/backends/trt.py
xuzhao9: One suggestion is to use pytree to traverse the input and cast them to `torch_tensorrt.Input`. Similar to this: https://github.com/pytorch/benchmark/blob/9d84f9e6edcae65b43604e5781c828a05121df78/torchbenchmark/util/env_check.py#L186

gs-olive: Since the different `ir` choices for Torch-TRT process inputs differently, but all can handle Torch Tensor inputs, I passed the example inputs directly to the compiler instead, so the selected `ir` can handle the inputs/casting as necessary.

gs-olive:
I have addressed the comment regarding example inputs here: https://github.com/pytorch/benchmark/pull/1753#discussion_r1271062716

xuzhao9 (approved):
LGTM, thanks!

PR #1758, opened 2023-07-10
Add Falcon-7B HuggingFace model

petermcaughan:
This PR adds two features to torchbenchmark:

1. Include [tiiuae/falcon-7b model from HuggingFace](https://huggingface.co/tiiuae/falcon-7b) in torchbenchmark suite.
2. Include code path to pass arguments from model's `__init__.py` into model constructor during benchmarking. **Running Falcon-7B in HuggingFace requires the parameter `trust_remote_code` to be passed**. 
    - There are two files that are executed when Falcon is loaded: [configuration_RW](https://huggingface.co/tiiuae/falcon-7b/blob/main/configuration_RW.py) and [modelling_RW](https://huggingface.co/tiiuae/falcon-7b/blob/main/modelling_RW.py) that both contain what seem to be safe code related to modeling specifics.

### DIFF
--- a/torchbenchmark/models/hf_Falcon_7b/__init__.py
+++ b/torchbenchmark/models/hf_Falcon_7b/__init__.py
@@ -1,0 +1,10 @@
+from torchbenchmark.tasks import NLP
+from torchbenchmark.util.framework.huggingface.model_factory import HuggingFaceModel
+
+class Model(HuggingFaceModel):
+    task = NLP.LANGUAGE_MODELING
+    DEFAULT_TRAIN_BSIZE = 4
+    DEFAULT_EVAL_BSIZE = 1
+
+    def __init__(self, test, device, jit=False, batch_size=None, extra_args=[]):
+        super().__init__(name="hf_Falcon_7b", test=test, device=device, jit=jit, batch_size=batch_size, extra_args=extra_args)
--- a/torchbenchmark/models/hf_Falcon_7b/install.py
+++ b/torchbenchmark/models/hf_Falcon_7b/install.py
@@ -1,0 +1,14 @@
+
+import subprocess
+import sys
+import os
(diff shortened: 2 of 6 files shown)


torchbenchmark/models/hf_Falcon_7b/__init__.py
```diff
@@ -0,0 +1,10 @@
+from torchbenchmark.tasks import NLP
+from torchbenchmark.util.framework.huggingface.model_factory import HuggingFaceModel
+
+class Model(HuggingFaceModel):
+    task = NLP.LANGUAGE_MODELING
+    DEFAULT_TRAIN_BSIZE = 4
```
xuzhao9, line 6: Can we have a link to upstream code that reference the train batch size?

petermcaughan: The Falcon README shows the training batch size to be 2304 [link](https://huggingface.co/tiiuae/falcon-7b/blob/main/README.md#:~:text=Batch%20size,tokens%20ramp%2Dup)] - included link in comment. This seems prohibitive to run in testing suite though, so I kept the default size to 4.

xuzhao9: Thanks for making the reference! We should use the per-device batch size here. In the link it mentioned 2304 batch size on 384 A100 GPUs, so per-GPU batch size should be 6. Can you please update it to 6 and add the link ( https://huggingface.co/tiiuae/falcon-7b/blob/main/README.md#:~:text=Batch%20size,tokens%20ramp%2Dup) as the reference?

petermcaughan: Thanks for the clarification! Updated the size and the reference link.

PR #1760, opened 2023-07-10
export chrome trace when profiler is enabled

dshi7:
Summary: as title.  default off (enabled by `--profile-trace`)

Reviewed By: xuzhao9

Differential Revision: D47234285


aaronenyeshi (approved):
LGTM, wondering if the option should be a little more descriptive, like profile-export-chrome-trace or -profile-export-chrome

PR #1765, opened 2023-07-13
Change ibm runner tag

xuzhao9:


.github/workflows/userbenchmark-ibmcloud-testrunner.yml
```diff
@@ -16,7 +16,7 @@ env:
   SETUP_SCRIPT: "/workspace/setup_instance.sh"
 jobs:
   run-userbenchmark:
-    runs-on: [self-hosted, testrunner1545]
+    runs-on: [testrunner1545]
```
malfet, line 19: This is a label name, not a runner name. And this runner labels are Linux, X64
image

spzala: Thanks and yes, `testrunner1545` is the runner name. The rest are the labels.

xuzhao9: The `runs-on` tag can only accept labels, not runner name. We need to add the "testrunner1545" label to the runner to schedule the workflow.

PR #1771, opened 2023-07-17
Fix require_grad typo

kit1980:
Fix require_grad typos (should be requires_grad).
Before the fix, the code doesn't cause any errors but doesn't do what it's supposed to do.

Fixed with TorchFix https://github.com/pytorch/test-infra/tree/main/tools/torchfix

Upstream PR: https://github.com/codertimo/BERT-pytorch/pull/104

### DIFF
--- a/torchbenchmark/models/BERT_pytorch/bert_pytorch/model/embedding/position.py
+++ b/torchbenchmark/models/BERT_pytorch/bert_pytorch/model/embedding/position.py
@@ -10,7 +10,7 @@ class PositionalEmbedding(nn.Module):
 
         # Compute the positional encodings once in log space.
         pe = torch.zeros(max_len, d_model).float()
-        pe.require_grad = False
+        pe.requires_grad = False
 
         position = torch.arange(0, max_len).float().unsqueeze(1)
         div_term = (torch.arange(0, d_model, 2).float() * -(math.log(10000.0) / d_model)).exp()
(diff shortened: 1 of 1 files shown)


xuzhao9:
Thanks! Looks like torchfix is a very useful tool!

Note that the upstream code also has this bug: https://github.com/codertimo/BERT-pytorch/blob/master/bert_pytorch/model/embedding/position.py#L13

Can you help submit a PR to the upstream code, and reference it in this PR? Thanks!

kit1980:
I'll submit upstream.

xuzhao9:
@kit1980 Thanks! Can you please also add the upstream PR link (https://github.com/codertimo/BERT-pytorch/pull/104) in this PR?

xuzhao9 (approved):

torchbenchmark/models/BERT_pytorch/bert_pytorch/model/embedding/position.py
```diff
@@ -10,7 +10,7 @@ def __init__(self, d_model, max_len=512):
 
         # Compute the positional encodings once in log space.
         pe = torch.zeros(max_len, d_model).float()
-        pe.require_grad = False
+        pe.requires_grad = False
```
xuzhao9, line 13: Could you please add `https://github.com/codertimo/BERT-pytorch/pull/104` here so that we can track the upstream code change?

kit1980: done

PR #1775, opened 2023-07-19
llama v2

msaroufim:


msaroufim:
Ok weird there's an OOM error here even though this is passing locally on my A10G

torchbenchmark/models/llama_v2/requirements.txt
```diff
@@ -0,0 +1,3 @@
+sentencepiece
+datasets
+transformers==4.31.0
```
xuzhao9, line 3: This file is not needed -  we want to pin transformers version in the top-level requirements.txt file: https://github.com/pytorch/benchmark/blob/main/requirements.txt#L13

torchbenchmark/models/llama_v2/metadata.yaml
```diff
@@ -0,0 +1,15 @@
...
+eval_benchmark: false
+eval_deterministic: false
+eval_nograd: true
+not_implemented:
+- jit: true
+- device: cpu
+- device: cuda
+  test: train
+- device: cuda
+  test: example
```
xuzhao9, line 13: I am wondering why cuda example is disabled? Is it because of OOM?

The example test will check whether the output tensor is deterministic across runs, and it will be useful for the accuracy tests.

xuzhao9:
Could you please paste the command and output of running the model locally?
For example, the output of 
```
python run.py llama2 -d cuda
```

torchbenchmark/models/llama_v2/metadata.yaml
msaroufim: Yep at num_heads = 16 this was the only test that OOM'd because of some cloning - lemme get the error message

msaroufim: Here you go https://github.com/pytorch/benchmark/actions/runs/5623600850/job/15238632991#step:6:909

xuzhao9: If cloning is an issue, you can add `DEEPCOPY = False` to disable it, similar to https://github.com/pytorch/benchmark/blob/36e1ed1c435ff8a4953a7c5e4791e76e6714b269/torchbenchmark/models/fambench_xlmr/__init__.py#L57

Can you please add `DEEPCOPY = False` and try again?

msaroufim: Alright trying it now in CI

msaroufim: Ok with deep copy off it still ooms at the full model size

xuzhao9 (approved):
Thanks for adding llama v2 model!

PR #1779, opened 2023-07-24
[optim][ub] Print torch git version (commit) in workflow

janeyx99:
Print out the version so we can easily know which torch nightly we're running the regressions in the workflow.


xuzhao9 (approved):
LGTM!

PR #1783, opened 2023-07-24
[optim][ub] Print out runnable repros with better time metadata

janeyx99:
This way, it is easy to just copy/paste the command into a CLI

### DIFF
--- a/userbenchmark/optim/run.py
+++ b/userbenchmark/optim/run.py
@@ -377,14 +377,15 @@ def is_excluded(mn: str, d: str, on: str, func_str: str, defaults: Dict[str, Any
 def run_model(modelName, device, Optim, defaults, maybe_pt2_):
     try:
         params = get_model_params(modelName, device)
-        print('getting params: ', params[0].size(), params[0].dtype, len(params), params[0].device)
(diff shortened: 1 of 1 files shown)


xuzhao9 (approved):
Approved to unblock, see inline suggestion.

userbenchmark/optim/run.py
```diff
@@ -377,14 +377,15 @@ def is_excluded(mn: str, d: str, on: str, func_str: str, defaults: Dict[str, Any
 def run_model(modelName, device, Optim, defaults, maybe_pt2_):
     try:
         params = get_model_params(modelName, device)
-        print('getting params: ', params[0].size(), params[0].dtype, len(params), params[0].device)
+        print(datetime.datetime.now(), 'getting params: ', params[0].size(), params[0].dtype, len(params), params[0].device)
```
xuzhao9, line 380: Maybe it is better to use some standard library like `logging` here?

janeyx99: if i add another print statement i will switch to a logger, right now it doesn't seem to provide more value add.

PR #1786, opened 2023-07-25
[optim] fix regression detector after #1710

janeyx99:
1. Fix longstanding issue where the optim benchmarks weren't able to calculate regressions
2. Fix type annotations to be correct
3. Add {} as default for the *_only_metrics.


xuzhao9 (approved):
Sorry didn't capture this in the first place... and thanks for the fix!!

Issue #1777, opened 2023-07-20
Remove jit from model options


xuzhao9:
Torchscript is in archived mode, we can remove it from the model options.

PR #1787, opened 2023-07-25
Remove jit from the framework code

xuzhao9:
Jit(torchscript) is now in archive mode and there is no need that we add it to the explicit options.

Jit(torchscript) is still supported with the extra args option `--backend torchsript`.

Fixes https://github.com/pytorch/benchmark/issues/1777


msaroufim (approved):
farewell jit

PR #1792, opened 2023-07-28
Disable llama_v2_7b_16h on optim benchmarks as they currently OOM

janeyx99:
Fixes https://github.com/pytorch/benchmark/issues/1791

Test plan:
https://github.com/pytorch/benchmark/actions/runs/5693564120 a run of the optim benchmarks


torchbenchmark/models/llama_v2_7b_8h/__init__.py
```diff
@@ -0,0 +1,15 @@
...
+    DEFAULT_TRAIN_BSIZE = 1
+    DEFAULT_EVAL_BSIZE = 1
+    DEEPCOPY = False 
+
+    def __init__(self, test, device, batch_size=None, extra_args=[]):
+        HuggingFaceAuthMixin.__init__(self)
+        super().__init__(name="llama_v2_7b_8h", test=test, device=device, batch_size=batch_size, extra_args=extra_args)
+
+    def train(self):
+        return NotImplementedError("7b LLAMA model will OOM on CI GPU machines")
```
xuzhao9, line 15: Currently our criteria to accept model is that the eager mode test does not OOM on CI GPU machines (A10G and A100, respectively). We cannot guarantee that they won't OOM on the optim userbenchmark.

xuzhao9 (commented):
I am wondering how much value does it add since we already have llama_v2_7b_16h. My understanding is neither llama_v2_7b_16h nor llama_v2_7b_8h has train test, how it can be useful in the optimizer userbenchmark?

xuzhao9 (approved):
LGTM. We can add the test back when train is supported on this model.

PR #1797, opened 2023-07-31
Move jit from cpu userbenchmark supportted arg to an extra arg

chuanqi129:
Fixed cpu userbenchmark on jit mode 

works for #1293


xuzhao9 (approved):
LGTM!

PR #1799, opened 2023-07-31
Update `detectron2` to fix Pillow errors

HMellor:
As reported in https://github.com/facebookresearch/detectron2/issues/5010, `PIL.Image.LINEAR` no longer exists.

The Pillow requirement in `detectron2` isn't pinned, so old `detectron2` is installing the latest Pillow and causing errors in the TorchBench tests.


xuzhao9 (approved):
Thanks for fixing this!

PR #1803, opened 2023-08-02
all the llamas in canary

msaroufim:
All of these OOM on a single device but want to make them all available for the distributed tests @H-Huang is working on


xuzhao9 (approved):
More llamas!

torchbenchmark/canary_models/llama_v2_13b/__init__.py
```diff
@@ -0,0 +1,15 @@
+from torchbenchmark.tasks import NLP
```
H-Huang, line 1: just curious, these models are in `canary_models` as opposed to `models` dir. Does this effect how these models are pulled in either from torchbench or dyanmorunner and if so in what way?

H-Huang (approved):
Awesome! Thanks for adding :)

torchbenchmark/canary_models/llama_v2_13b/__init__.py
msaroufim: This error message explains it super well, should be no difference

```
  File "/var/lib/jenkins/workspace/benchmarks/dynamo/torchbench.py", line 302, in load_model
    raise ImportError(f"could not import any of {candidates}")
ImportError: could not import any of ['torchbenchmark.models.stable_diffusion', 'torchbenchmark.canary_models.stable_diffusion', 'torchbenchmark.models.fb.stable_diffusion']
ERROR
```

PR #1805, opened 2023-08-02
Make `test.py` device agnostic

HMellor:
Allows the user to specify an arbitrary device and an arbitrary backend import using environment variables. The three added environment variables are:
- `TIMEOUT` - allows the timeout to be configured to suit the characteristics of the arbitrary device
- `ACCELERATOR` - a string which is added to the devices list, just like `'cpu'`, `'cuda'` & `'mps'` currently are
- `ACCELERATOR_BACKEND` - a string which is the name of the backend package (which doesn't yet exist in upstream `torch`) needed to use `ACCELERATOR`

These environment variables are used as follows:
```
TIMEOUT=100 ACCELERATOR=my_accelerator ACCELERATOR_BACKEND=my_accelerator_backend \
    python test.py -k "test_hf_Bert_check_device_my_accelerator"
```

### DIFF
--- a/torchbenchmark/__init__.py
+++ b/torchbenchmark/__init__.py
@@ -295,6 +295,8 @@ class ModelTask(base_task.TaskBase):
         diagnostic_msg = ""
         try:
             module = importlib.import_module(f'.models.{model_name}', package=package)
+            if accelerator_backend := os.getenv("ACCELERATOR_BACKEND"):
+                setattr(module, accelerator_backend, importlib.import_module(accelerator_backend))
             Model = getattr(module, 'Model', None)
             if Model is None:
                 diagnostic_msg = f"Warning: {module} does not define attribute Model, skip it"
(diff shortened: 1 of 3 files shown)


torchbenchmark/__init__.py
```diff
@@ -295,6 +295,8 @@ def _maybe_import_model(package: str, model_path: str) -> Dict[str, Any]:
         diagnostic_msg = ""
         try:
             module = importlib.import_module(f'.models.{model_name}', package=package)
+            if accelerator_backend := os.getenv("ACCELERATOR_BACKEND"):
+                setattr(module, accelerator_backend, importlib.import_module(accelerator_backend))
```
xuzhao9, line 299: It seems the `accelerator_backend` is not being used. Can you give an example of its usage?

xuzhao9 (approved):
LGTM. Note that `test.py` is mainly for unit test purpose, not benchmarking. Since we only have CPU/CUDA devices in our CI machines, we can't test it against customized accelerators or accelerator backends.

HMellor:
Thanks for the speedy review.

> `test.py` is mainly for unit test purpose

Understood, this is what we intend to use `test.py` for 😄

PR #1819, opened 2023-08-10
Cleanup the unused files in the root directory

xuzhao9:
Clean up the unused run scripts in the root directory.
Now all scripts that "sweep through models and collect metrics" should be implemented as userbenchmark.

Test upgraded numpy version on nightly docker build:
https://github.com/pytorch/benchmark/actions/runs/5823289368


FindHao (approved):
LGTM

FindHao:
there is a related issue https://github.com/opencv/opencv-python/issues/885 . Looks like we may need to delay the upgrade of numpy.

xuzhao9:
> there is a related issue [opencv/opencv-python#885](https://github.com/opencv/opencv-python/issues/885) . We may need to delay the upgrade of numpy.

I see, thanks for digging this up! Will update the PR to keep the same numpy version

PR #1821, opened 2023-08-11
[backends] List models function call replacement

gs-olive:
- Replace `_list_model_paths` call with `list_models`

### DIFF
--- a/userbenchmark/torch_trt/run.py
+++ b/userbenchmark/torch_trt/run.py
@@ -13,12 +13,12 @@ from typing import List, Union
 from torchbenchmark.util.experiment.instantiator import (
     TorchBenchModelConfig,
     load_model_isolated,
+    list_models,
 )
 from torchbenchmark import (
     ModelTask,
     load_canary_model_by_name,
     load_model_by_name,
-    _list_model_paths,
     ModelNotFoundError,
 )
 from torchbenchmark.util.model import BenchmarkModel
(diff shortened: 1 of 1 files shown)


xuzhao9 (approved):
Thanks for fixing this!

xuzhao9:
BTW, we tested the running the userbenchmark at: https://github.com/pytorch/benchmark/actions/runs/5833882739

Unfortunately, the workflow failed because we currently don't have tensorrt or torch_trt installed in our nightly docker.
Can you please help fix our docker file? https://github.com/pytorch/benchmark/blob/main/docker/torchbench-nightly.dockerfile

gs-olive:
Hi @xuzhao9 - thanks for the reference! I looked into the Dockerfile and there's a few considerations we'll need to address on the Torch-TRT side to ensure the import will work. I have a few questions to address this:

1. At what time is the nightly CI run?
2. What Python and CUDA versions does the Docker container use?
3. Does the Docker container use the latest-available Torch nightly?

This will help me to understand the requirements to provide a nightly build of Torch-TRT which will be compatible with the versions in the container.

xuzhao9:
1. The nightly CI will run everyday at 4PM UTC (https://github.com/pytorch/benchmark/blob/main/.github/workflows/userbenchmark-a100.yml#L4)
2. The default Python and CUDA versions are defined at https://github.com/pytorch/benchmark/blob/main/utils/cuda_utils.py#L11 and https://github.com/pytorch/benchmark/blob/main/utils/python_utils.py#L4, respectively
3. Yes, the Docker container uses the latest-available Torch nightly release, we will update the Docker everyday at https://github.com/pytorch/benchmark/actions/workflows/build-nightly-docker.yml

gs-olive:
Thank you for the detailed answers, this is very helpful. We are working on publishing nightly releases of Torch-TRT via Github Actions and I will update the Dockerfile as soon as those are ready.

PR #1829, opened 2023-08-16
fix the version of opencv-python

FindHao:
According to the [upstream issue](https://github.com/opencv/opencv-python/issues/885), opencv-python has a compatibility problem with numpy. Since we fixed the numpy version in the requirements.txt, we also need to fix the opencv-python's verison too. I install the nightly Torch and TorchBench, there is an TorchBench installation error same with the above issue with opencv-python 4.8. But opencv-python 4.7.0.72 works well. I think we'd better fix its version until the latest opencv-python 4.9.0 released.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -17,7 +17,7 @@ pyyaml
 # pytorch build script pins numpy version
 # https://github.com/pytorch/builder/blob/ae5c82e65cb3d8bac6df50e742a195019af91ad3/wheel/build_wheel.sh#L145
 numpy==1.21.2
-# https://github.com/pytorch/benchmark/tree/opencv-python-compatibility
+# https://github.com/opencv/opencv-python/issues/885
 opencv-python==4.7.0.72
 # Need https://github.com/kornia/kornia/commit/53808e5 to work on PyTorch nightly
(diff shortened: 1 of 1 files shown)


requirements.txt
```diff
@@ -17,6 +17,7 @@ pyyaml
 # pytorch build script pins numpy version
 # https://github.com/pytorch/builder/blob/ae5c82e65cb3d8bac6df50e742a195019af91ad3/wheel/build_wheel.sh#L145
 numpy==1.21.2
+opencv-python==4.7.0.72
```
xuzhao9, line 20: Can we add link to the upstream bug here? This is helpful to track why we are adding this constraint and when we can remove it.

FindHao: added in https://github.com/pytorch/benchmark/pull/1829/commits/fce2df3619c366c5e574dd73682be76f9c80d5de

xuzhao9: Sorry, I mean to add link to the upstream GitHub issue, not the branch URL.

FindHao: ah, it's a typo.

xuzhao9 (approved):
LGTM

PR #1833, opened 2023-08-17
Remove attention_is_all_you_need and pytorch_struct

xuzhao9:
This is a follow-up of https://github.com/pytorch/benchmark/pull/1831

Related to the missing metrics in https://github.com/pytorch/benchmark/issues/1832


janeyx99 (approved):

xuzhao9:
@janeyx99 It should only be related to the missing metrics of nvidia_deeprecommender, attention_is_all_you_need_pytorch, and pytorch_struct.

The metric increase in DALLE2_pytorch should be an independent issue.

janeyx99:
yes that was why i had specified runtime errors—i will look into the two perf regressions

PR #1837, opened 2023-08-18
Move clip to canary_models

xuzhao9:
This is a follow-up of https://github.com/pytorch/benchmark/pull/1831.

Clip depends on torchmultimodal, which further depends on torchtext:
https://github.com/pytorch/benchmark/actions/runs/5895605535/job/15991723413

Move the clip model to canary_models.


msaroufim:
hmm, @ankitade is there any way we can keep clip without depending on torchtext?

msaroufim (approved):
:(

PR #1865, opened 2023-08-31
[optim][ub] Distinguish compile time from run time

janeyx99:
This PR teases out the compile time and run time to track two numbers, where previously, the metrics measured e2e compile+runtime.

SGD with foreach, momentum, CUDA on pt2 today takes ~2.36s on average. I anticipate the runtime to be close to that after this change still, whereas the compile time will be a separate much larger value.

Current results:
image

The metrics dictionary used to have entries like:
```
{
    "name": "optim",
    "environ": {
        "pytorch_git_version": "c527d0fadd27595ba2f98dd0f57aae5c56658d71"
    },
    "metrics": {
        "resnet18, Adam, cuda, (pt2) default": 0.0018993107215413507,
        "resnet18, Adam, cuda, default": 0.0010943956114351748,
        "resnet18, Adam, cuda, (pt2) amsgrad, maximize": 0.002033790648736135,
        "resnet18, Adam, cuda, amsgrad, maximize": 0.0013529009232297541,
        "resnet18, Adam, cuda, (pt2) no_foreach": 0.005578947072434757,
        ...
    }
}
```

But now, the keys will contain "compile_time" at the beginning if we're measuring compile time:
```
{
    "name": "optim",
    "environ": {
        "pytorch_git_version": "a005f70a4284152e00c8f6603feaf4ab9636f6aa"
    },
    "metrics": {
        "resnet18, SGD, cuda, (pt2) no_foreach": 0.0017500566132366657,
        "resnet18, SGD, cuda, no_foreach": 0.0025729038193821907,
        "resnet18, SGD, cuda, (pt2) foreach": 0.0017613966017961502,
        ...
        "resnet18, SGD, cpu, foreach, momentum=0.9": 0.08240865767002106,
        "compile_time, resnet18, SGD, cuda, (pt2) no_foreach": 14.877577589824796,
        "compile_time, resnet18, SGD, cuda, (pt2) foreach": 0.6698574535548687,
        "compile_time, resnet18, SGD, cuda, (pt2) foreach, momentum=0.9, nesterov": 0.32723781156043213,
        ...
        "compile_time, resnet18, SGD, cpu, (pt2) foreach, momentum=0.9": 0.29321490600705147
    }
}
```

### DIFF
--- a/userbenchmark/optim/regression_detector.py
+++ b/userbenchmark/optim/regression_detector.py
@@ -2,6 +2,7 @@ from typing import Optional
 from ..utils import TorchBenchABTestResult, TorchBenchABTestMetric
 
 DEFAULT_REGRESSION_DELTA_THRESHOLD = 0.3
+COMPILE_TIME_REGRESSION_DELTA_THRESHOLD = 2.0
 
 def run(control, treatment) -> Optional[TorchBenchABTestResult]:
     control_env = control["environ"]
@@ -11,10 +12,11 @@ def run(control, treatment) -> Optional[TorchBenchABTestResult]:
     details = {}
     for control_metric_name, control_metric in control_metrics.items():
         if control_metric_name in treatment_metrics:
+            regression_threshold = COMPILE_TIME_REGRESSION_DELTA_THRESHOLD if "compile_time" in control_metric_name else DEFAULT_REGRESSION_DELTA_THRESHOLD
             treatment_metric = treatment_metrics[control_metric_name]
             delta = (treatment_metric - control_metric) / control_metric
             # Trigger on BOTH slowdowns and speedups
-            if abs(delta) > DEFAULT_REGRESSION_DELTA_THRESHOLD:
+            if abs(delta) > regression_threshold:
                 details[control_metric_name] = TorchBenchABTestMetric(control=control_metric, treatment=treatment_metric, delta=delta)
     # control_only_metrics/treatment_only_metrics will be filled in later by the main regression detector
     return TorchBenchABTestResult(name=control["name"],
--- a/userbenchmark/optim/run.py
+++ b/userbenchmark/optim/run.py
@@ -11,6 +11,7 @@ import gc
 import sys
 import itertools
 import datetime
+import time
 import yaml
 
 with add_path(REPO_PATH):
@@ -368,14 +369,11 @@ def generate_random_gradients(parameters):
 def optimizer_step(optimizer):
     optimizer.step()
 
-def compile_step(optimizer):
+def pt2_optimizer_step(optimizer):
     @torchdynamo.optimize('inductor')
     def f():
         optimizer.step()
-    return f
-
-def pt2_optimizer_step(compiled_step):
-    compiled_step()
+    f()
 
 def defaults_to_str(defaults: Dict[str, Any]) -> str:
     # We define lr for SGD, but we don't currently vary lr so it is effectively the default.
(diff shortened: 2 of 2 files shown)


userbenchmark/optim/run.py
```diff
@@ -368,11 +368,14 @@ def generate_random_gradients(parameters):
 def optimizer_step(optimizer):
     optimizer.step()
 
-def pt2_optimizer_step(optimizer):
+def compile_step(optimizer):
     @torchdynamo.optimize('inductor')
     def f():
         optimizer.step()
```
janeyx99, line 374: This is what we want to measure for compile time, yes?

xuzhao9: In the framework, we use this code (https://github.com/pytorch/benchmark/blob/main/torchbenchmark/util/model.py#L356) to get compilation time and number of graph breaks.

userbenchmark/optim/regression_detector.py
```diff
@@ -2,6 +2,7 @@
 from ..utils import TorchBenchABTestResult, TorchBenchABTestMetric
 
 DEFAULT_REGRESSION_DELTA_THRESHOLD = 0.3
+COMPILE_TIME_REGRESSION_DELTA_THRESHOLD = 2.0
```
janeyx99, line 5: compile time is a lot more fickle cuz we run it once only (compilation is slowww)

userbenchmark/optim/run.py
```diff
@@ -405,33 +406,48 @@ def run_model(modelName, device, Optim, defaults, maybe_pt2_):
...
+            times = []
+            if device == "cuda":
+                torch.cuda.reset_peak_memory_stats()
+                torch.cuda.empty_cache()
+            for _ in range(5):
+                t0 = time.perf_counter()
+                pt2_optimizer_step(optim)
+                t1 = time.perf_counter()
+                times.append(t1 - t0)
+            compile_r = (f'compile_time, {sub_label}, {description}', times[0] - sum(times[2:5]) / 3)
```
xuzhao9, line 427: In torchbench framework, we measure the compile time by calling this API: https://github.com/pytorch/benchmark/blob/main/torchbenchmark/util/model.py#L358

xuzhao9, line 427: Do we also need to clear Dynamo cache to minimize the compilation time variance?

xuzhao9 (approved):
LGTM overall, adding minor comments

userbenchmark/optim/run.py
janeyx99: No, this is assuming that the later compile times are reading the cache

userbenchmark/optim/run.py
```diff
@@ -405,33 +406,48 @@ def run_model(modelName, device, Optim, defaults, maybe_pt2_):
...
         print(f'{datetime.datetime.now()}     python -m userbenchmark.optim.run -m {modelName} -d {device}' +
               f' -o {Optim.__name__} --df "{defaults_to_str(defaults)}" -f {maybe_pt2_}')
+
+        compile_r = None
+        sub_label = f'{modelName}, {optim.__class__.__name__}, {device}'
+        description = pt2_description + defaults_to_str(defaults)
+
+        # Get compile time by running 5 times and subtracting
+        #   first entry - avg(entries 3 through 5)
+        # skipping the second entry due to high variance of first cache hit
```
mlazos, line 416: Huh, I'm not sure why this is. The second run has higher variance than later runs?

mlazos: We also might want to do more runs, I think we do five warmup runs before taking a measurement common.py. 3 runs are needed to initialize cudagraphs. So that could possibly be it.

mlazos (approved):
lgtm, but please add the 5 warmup runs before taking a runtime measurement. I will add cudagraphs support to this soon.

userbenchmark/optim/run.py
janeyx99: The 5 warmup runs are exactly what PT2 bms do today + the compile time is based on those 5 warmup runs. I'll do 10 runs though...would you agree that the first run (which is significantly longer in all cases i've seen) should be what's taken as the first operand?

aka...would you agree compile time is close to:
```
first_run - avg(the last n runs)
```
If I run it 10 times, I will make n = 5.

janeyx99: Like today, we basically already do 5 warmup runs before calculating runtime. This calculation is purely for compiletime

mlazos: cool sounds good, yeah this appears to be right. Thought avg(last n runs) was used for the runtime estiamte.

Issue #1838, opened 2023-08-18
Test that grad is disabled in eval tests by default


xuzhao9:
With a couple of exceptions, we should check that gradient calculation is disabled by default in eval tests.

PR #1867, opened 2023-09-01
Add DTensor LLaMA inference model: simple_gpt

xmfan:
Adds simple_gpt + DTensor implemented in https://github.com/pytorch-labs/simple_gpt/pull/7 to torchbench

Tested via `python benchmarks/dynamo/torchbench.py -d cuda --output-directory=benchmark_logs --output=performance.csv --inference --performance --timing --print-memory --multiprocess --nothing --only simple_gpt`. Note: --nothing is used here to disable compile, since DTensor + compile isn't yet supported in main

```
dev,name,batch_size,speedup,abs_latency,compilation_latency,compression_ratio,eager_peak_mem,dynamo_peak_mem,calls_captured,unique_graphs,graph_breaks,unique_graph_breaks
cuda,simple_gpt,1,0.966153,196.819773,-0.059319,1.000000,4.576880,4.576880,0,0,0,0
cuda,simple_gpt,1,0.967389,196.608152,-0.058833,1.000000,4.577404,4.577404,0,0,0,0
cuda,simple_gpt,1,0.973152,196.093583,-0.059316,1.000000,4.593133,4.593133,0,0,0,0
cuda,simple_gpt,1,0.973087,196.124046,-0.075580,1.000000,4.611483,4.611483,0,0,0,0
cuda,simple_gpt,1,0.967908,193.998484,-0.040192,1.000000,4.593133,4.593133,0,0,0,0
cuda,simple_gpt,1,0.968949,193.798088,-0.028878,1.000000,4.593133,4.593133,0,0,0,0
```

2 changes were required to the model:
- decorate torch.no_grad() on the caches, previously this was done outside the model, the entire eval call was wrapped in a torch.no_grad() context. After using torchbench, I notice even with only inference mode, we don't disable gradient calculations
- rank/world size, added support from torchbench side in https://github.com/pytorch/pytorch/pull/108438 and updated model to fetch from the provided extra_args

### DIFF
--- a/torchbenchmark/models/simple_gpt/__init__.py
+++ b/torchbenchmark/models/simple_gpt/__init__.py
@@ -1,0 +1,104 @@
+import os
+
+import torch
+import lightning as L
+from torch.distributed._tensor import DeviceMesh
+from torch.distributed.tensor.parallel import parallelize_module
+from torch.distributed.tensor.parallel.style import ColwiseParallel, RowwiseParallel
+from torchbenchmark.tasks import NLP
+
+from ...util.model import BenchmarkModel
+from .model import LLaMA
+from .utils import LOCAL_RANK, LOCAL_WORLD_SIZE
+
+
+class Model(BenchmarkModel):
+    task = NLP.GENERATION
+    DEFAULT_EVAL_BSIZE = 1
+
+    def __init__(self, test, device, batch_size=None, extra_args=[]):
+        super().__init__(
+            test=test,
+            device=device,
+            batch_size=batch_size,
+            extra_args=extra_args,
+        )
+
+        error = self.validate_environment()
+        if error:
+            # per ADDING_MODELS.md, convention is to fail silently in __init__ and raise in eval
+            return
+
+        fabric = L.Fabric(devices=[LOCAL_RANK], precision="bf16-true")
+        with fabric.init_module(empty_init=True):
+            self.model = LLaMA.from_name("7B")
+
+        # Tensor parallelism using DTensor
+        mesh = DeviceMesh("cuda", list(range(LOCAL_WORLD_SIZE)))
+        for block in self.model.transformer.h:
+            # prepare attention weights to be parallelized
+            block.attn.prepare_qkv_for_dtensor_tp()
+
+            parallelize_module(
+                module=block,
+                device_mesh=mesh,
+                parallelize_plan={
+                    "attn.c_attn_q": ColwiseParallel(),
+                    "attn.c_attn_k": ColwiseParallel(),
+                    "attn.c_attn_v": ColwiseParallel(),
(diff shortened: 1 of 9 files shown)


torchbenchmark/models/simple_gpt/__init__.py
```diff
@@ -0,0 +1,99 @@
...
+    def __init__(self, test, device, batch_size=None, extra_args=[]):
+        super().__init__(
+            test=test,
+            device=device,
+            batch_size=batch_size,
+            extra_args=extra_args,
+        )
+
+        fabric = L.Fabric(devices=[self._rank], precision="bf16-true")
+        with fabric.init_module(empty_init=True):
```
H-Huang, line 27: Just curious, is the only use of `Lightning` this init_module? I'm not sure what this does, but can we remove it by initializing with meta device? Then the implementation doesn't rely on third-party libraries can stay as close to native pytorch as possible.

H-Huang (approved):
Looks good

torchbenchmark/models/simple_gpt/__init__.py
xmfan: Yeah good point, let me update this with the latest init code that removed the lightning dep

Chillee: I've actually updated the initial code to avoid the lightning dependency (and it's actually much faster too!)

https://github.com/pytorch-labs/simple_gpt/blob/main/generate.py#L162

xuzhao9:
> decorate torch.no_grad() on the caches, previously this was done outside the model, the entire eval call was wrapped in a torch.no_grad() context. After using torchbench, I notice even with only inference mode, we don't disable gradient calculations

This is because some models in torchbench (e.g. maml) does need gradient calculations in inference mode. So we leave the choice of whether enabling the gradient context to the eval test code.

There is an open issue about testing that the `no_grad` is enabled in the eval test: https://github.com/pytorch/benchmark/issues/1838

torchbenchmark/models/simple_gpt/__init__.py
xmfan: For nightly runs, do we usually load the actual weights? The load time should be quick, but the weights file is 10+GB just for LLaMA-7B. Otherwise, I could also update the default weights initialization to use random instead of just torch.zeros

msaroufim: Use the real weights otherwise you might fail accuracy checks in unpredictable ways

xmfan: Is this still needed if we only run this model through dynamorunner (model requires distributed, multiple GPUs)? The unit tests from this repo are skipped i.e. python test.py -k 'test_simple_gpt_'

msaroufim (approved):

PR #1877, opened 2023-09-06
Disable torchtext and related models in v2 workflow

xuzhao9:
Fix a couple of compatibility issues when running the V2 workflow.

Test workflow: https://github.com/pytorch/benchmark/actions/runs/6117597436


aaronenyeshi (approved):
LGTM!

Issue #1850, opened 2023-08-25
why I can not choose the compiler??


LiuYang328:
![image](https://github.com/pytorch/benchmark/assets/58350195/1140aafe-b461-41f0-adef-534a098642c5)

if i use this command, it will select noting
BUT

![image](https://github.com/pytorch/benchmark/assets/58350195/7b4a1816-3efd-4af5-905e-03de81a41779)

if I delete the compiler in "test_train" , it can work , can you explain why?

xuzhao9:
`test_bench.py` doesn't support choosing compiler backends. You need to use run.py:

```
python run.py BERT_pytorch -d cuda -t train --torchdynamo inductor
```

Checkout https://github.com/pytorch/benchmark/issues/1825 for details

LiuYang328:
> `test_bench.py` doesn't support choosing compiler backends. You need to use run.py:
> 
> ```
> python run.py BERT_pytorch -d cuda -t train --torchdynamo inductor
> ```
> 
> Checkout #1825 for details

so that means the readme is wrong?

xuzhao9:
> > `test_bench.py` doesn't support choosing compiler backends. You need to use run.py:
> > ```
> > python run.py BERT_pytorch -d cuda -t train --torchdynamo inductor
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Checkout #1825 for details
> 
> so that means the readme is wrong?

Sorry can I have the reference from the readme? `test_bench.py` only supports eager mode for now. We should fix readme if it claims the opposite.

LiuYang328:
> > > `test_bench.py` doesn't support choosing compiler backends. You need to use run.py:
> > > ```
> > > python run.py BERT_pytorch -d cuda -t train --torchdynamo inductor
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Checkout #1825 for details
> > 
> > 
> > so that means the readme is wrong?
> 
> Sorry can I have the reference from the readme? `test_bench.py` only supports eager mode for now. We should fix readme if it claims the opposite.

Here:
<img width="1076" alt="image" src="https://github.com/pytorch/benchmark/assets/58350195/17994750-33ba-4570-b308-f5b2134e528a">
if I type the "Bert_pytorch-cuda-eager", it can not choose any test

[...]

PR #1878, opened 2023-09-06
Fix the test name in readme

xuzhao9:
`test_bench.py` only supports eager mode and we don't have plan to make it support pt2 mode.

Fixes https://github.com/pytorch/benchmark/issues/1850


aaronenyeshi (approved):
LGTM!

PR #1881, opened 2023-09-07
SD: image encoder bench

msaroufim:
Problem was in our benchmarks we were only benchmarking the text encoder that's not actually the bottleneck so instead now benchmarking in addition the unet

The dynamo runners assume the inputs must be nn modules we might be able to relax this to a HF pipeline object but in the meantime this should be fine


msaroufim:
Interesting error

https://github.com/pytorch/benchmark/actions/runs/6114290635/job/16595557842?pr=1881#step:7:1379

xuzhao9:
> Interesting error
> 
> https://github.com/pytorch/benchmark/actions/runs/6114290635/job/16595557842?pr=1881#step:7:1379

This should be fixed by https://github.com/pytorch/benchmark/pull/1883, rerunning the workflows.

msaroufim:
Yup indeed, that worked thanks @xuzhao9 could I get a stamp to merge?

cpuhrsch (approved):

PR #1908, opened 2023-09-19
Add warmup and benchamrk iterations for cpu userbenchmark

chuanqi129:
To stabilize the benchmark results of cpu userbenchmark. 

Also works for #1293


xuzhao9 (approved):
LGTM!

PR #1914, opened 2023-09-20
Add interface to support multiple batches

xuzhao9:
The interface `self.gen_input_iter()` is an input generator that yields the next batch of available input.

We support dynamic shapes on 3 categories of models:
- huggingface (hf_*): we randomize the bucket_len of the input data
- timm_*: we randomize the batch_size of the input data
- torchvision: we randomize the batch_size of the input data

This PR also introduces a new type of train test, `train_dynamic`, to run the train test with a list of batches. By default, we will run 10 batches of inputs returned from the generator.


torchbenchmark/util/framework/detectron2/model_factory.py
```diff
@@ -163,22 +161,22 @@ def enable_fp16(self):
         self.example_inputs = prefetch(self.example_inputs, self.device, self.dargs.precision)
 
     def train(self):
+        batch_id = 0
```
ezyang, line 164: it feels a little suboptimal to have lost the ability to realize that the benchmark runner only knows how to run with a single input lol

ezyang (approved):
can you check that this works with the PT2 benchmark runner? Thanks!

torchbenchmark/util/framework/detectron2/model_factory.py
ezyang: ok

xuzhao9:
> can you check that this works with the PT2 benchmark runner? Thanks!

@ezyang I tried a couple of pt2 benchmark commands:

```
$ python torchbench.py --only hf_Bert -dcuda --training --inductor --performance --output-directory $HOME
loading model: 0it [00:27, ?it/s]
cuda train hf_Bert                            1.679x
```

```
$ python torchbench.py --only resnet50 -dcuda --training --inductor --performance --output-directory $HOME
loading model: 0it [00:21, ?it/s]
cuda train resnet50                    1.196x
```

They all seem to work. If it breaks any test or if you want me to test more commands, please let me know.

PR #1921, opened 2023-09-22
hf and multimodal clip

msaroufim:
multimodal clip is in canary because of torchtext dependency so this adds the hf version

Notably this PR also makes it possible to support dict based inputs to `get_module()` which is very common in HF code


eellison (approved):
🚀

xuzhao9 (approved):
LGTM!

msaroufim:
No space left on device seems like another upstream issue

msaroufim:
Ok I think the problem is `example_inputs` is a dict instead of a list so will need to unpack it correctly in the torchbench tests

PR #1923, opened 2023-09-22
Instructions to use torchbench as a library

msaroufim:
This is how the dynamo runner use torchbench, I found this to be extremely useful for quick local sanity checks so figured others might as well

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,7 +141,7 @@ model_name = "torchbenchmark.models.stable_diffusion_text_encoder" # replace thi
 module = importlib.import_module(model_name)
 
 benchmark_cls = getattr(module, "Model", None)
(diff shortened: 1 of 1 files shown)


xuzhao9 (approved):
Looks good to me! Ideally, we hope we could make it as simple as 

```
# from shell
pip install git+https://www.github.com:pytorch/benchmark.git
# In python file
import torchbench
...
```

msaroufim:
> Looks good to me! Ideally, we hope we could make it as simple as
> 
> ```
> # from shell
> pip install git+https://www.github.com:pytorch/benchmark.git
> # In python file
> import torchbench
> ...
> ```

I've done this for a few libraries already, let me work on that next

PR #1930, opened 2023-09-25
Rename aot_runtime to aoti_runtime

desertfire:
Summary:
X-link: https://github.com/pytorch/pytorch/pull/110007

Make the naming more explicit

Differential Revision: D49593528

### DIFF
--- a/userbenchmark/dynamo/common.py
+++ b/userbenchmark/dynamo/common.py
@@ -1134,7 +1134,7 @@ class AOTInductorModelCache:
 
             # Use a utility function for easier benchmarking
             source = """
-            #include <torch/csrc/inductor/aot_runtime/model.h>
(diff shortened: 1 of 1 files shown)


xuzhao9 (approved):

desertfire:
@xuzhao9 , I saw this file is not really in sync with the one in pytorch. Should we setup some kind of auto sync?

xuzhao9:
@desertfire That would be very helpful! We could setup an auto-sync in the directory `userbenchmark/dynamo/dynamobenchmark`.

One thing to note is that dynamo benchmark requires a couple of files from pytorch source repo that is not available in pytorch nightly release (https://github.com/pytorch/benchmark/tree/main/userbenchmark/dynamo/_dynamo). We need to sync those files as well, because torchbench only assumes pytorch nightly release, not pytorch source code.

PR #1932, opened 2023-09-25
[optim][ub] disable PT2 benchmarks until we investigate the noisy reports

janeyx99:
There have been many reports lately and they're not very actionable. It is time to actually investigate them instead of creating more issues every night. Until we investigate, the stats are noisy and not useful anyway.


janeyx99:
@xuzhao9  I don't think it's necessarily the PT2 code...it feels more infrastructure related, like maybe it's reading a cache and the file system is hard to reach.

PR #1933, opened 2023-09-25
torchbench is now a library

msaroufim:
```python
import torchbenchmark.models.densenet121
model, example_inputs = torchbenchmark.models.densenet121.Model(test="eval", device="cuda", batch_size=1).get_module()
model(*example_inputs)
```

So making the above example work with `import torchbenchmark` alone is a bit trickier there's a lot of relative imports that need to be fixed in pretty much every single file

I also added a simple `test_imports.py` file to make sure this doesn't break anything, I can setup a standalone github action or plug into an existing one (would rather do the latter to avoid another job that needs to run `install.py` again

This also solves @voznesenskym favorite design pattern from https://github.com/pytorch/pytorch/blob/main/benchmarks/dynamo/torchbench.py

```python
    for torchbench_dir in (
        "./torchbenchmark",
        "../torchbenchmark",
        "../torchbench",
        "../benchmark",
        "../../torchbenchmark",
        "../../torchbench",
        "../../benchmark",
    ):
```


xuzhao9:
Looking good! Though I am not sure what we should do about `install.py`. Just running `pip install git+https://www.github.com:pytorch/benchmark.git` does not mean we can run all the models. We still need to run `install.py` to install all the model dependencies, download data to the directory, etc.

msaroufim:
I think you should keep `install.py` as is since you do expect users to bring their own torch versions and there's a large number of dependencies that will be included in the torchbench package

So user should do 

```bash
# Install torch/audio/text from source or nightlies
python install.py
pip install .
```

xuzhao9 (approved):

setup.py
```diff
@@ -0,0 +1,16 @@
+from setuptools import setup, find_packages
+
+setup(
+    name='torchbench',
+    version='0.1',
+    description='Benchmarking library for PyTorch',
+    author='PyTorch Team',
+    url='https://github.com/pytorch/benchmark',  # replace with the actual URL of your project
```
xuzhao9, line 8: Shouldn't we remove the inline comments?

setup.py
```diff
@@ -0,0 +1,16 @@
...
+    packages=find_packages(include=['torchbenchmark*', 'userbenchmark*']),
+    classifiers=[
+        'Intended Audience :: Developers',
+        'Topic :: Software Development :: Build Tools',
+        'License :: OSI Approved :: BSD 3 License',  # assuming MIT License, modify accordingly
```
xuzhao9, line 13: Same here, remove the inline comment.

xuzhao9:
> I think you should keep `install.py` as is since you do expect users to bring their own torch versions and there's a large number of dependencies that will be included in the torchbench package
> 
> So user should do
> 
> ```shell
> # Install torch/audio/text from source or nightlies
> python install.py
> pip install .
> ```

Sounds reasonable, it is better to clarify in the `README.md` that they need to run both `python install.py`  and `pip install .`

PR #1935, opened 2023-09-25
Fix PyTorch CI HUD dashboard missing perf numbers: hf_Whisper

xmfan:
A few models were passing accuracy check, but surprisingly failing the perf run, resulting in dashboard entries like:
image

Reproing the hud's commands locally,
```
# pass
python benchmarks/dynamo/torchbench.py --accuracy --no-translation-validation --training --amp --backend inductor --disable-cudagraphs --device cuda --total-partitions 4 --partition-id 1 --output hf_Whisper_accuracy.csv --only hf_Whisper

# fail (on https://github.com/pytorch/benchmark/blob/4ea3bba3b8010f5d4a629bb8f530a92570f34518/torchbenchmark/util/model.py#L195C48-L195C48)
python benchmarks/dynamo/torchbench.py --performance --cold-start-latency --training --amp --backend inductor --disable-cudagraphs --device cuda --total-partitions 4 --partition-id 1 --output hf_Whisper_perf.csv --only hf_Whisper
```

The error suggests that hf_Whisper does not provide a batch size for the training mode perf run.

Summarizing discussion with @xuzhao9:
> I think we could:
> 1. set a default train batch size for hf_Whisper, if you still want to test forward/backward pass without a defined train test
> 2. in model.py, make sure self.batch_size is not None (before accuracy check overrides batch size to 4)

I implement 1, we set default batch sizes in the parent class of all benchmark models, with ability to be overwritten by individual models.

### DIFF
--- a/torchbenchmark/util/model.py
+++ b/torchbenchmark/util/model.py
@@ -201,6 +201,9 @@ class BenchmarkModel(metaclass=PostInitProcessor):
             default_batch_size = self.DEFAULT_TRAIN_BSIZE if self.test == "train" else self.DEFAULT_EVAL_BSIZE
             self.batch_size = default_batch_size
 
+        if not self.batch_size:
+            raise NotImplementedError(f"Model's {'DEFAULT_TRAIN_BSIZE' if self.test == 'train' else 'DEFAULT_EVAL_BSIZE'} is not implemented.")
+
         # Check if specified batch size is supported by the model
         if hasattr(self, "ALLOW_CUSTOMIZE_BSIZE") and (not getattr(self, "ALLOW_CUSTOMIZE_BSIZE")):
             if self.test == "train" and (not self.batch_size == self.DEFAULT_TRAIN_BSIZE):
@@ -209,10 +212,6 @@ class BenchmarkModel(metaclass=PostInitProcessor):
                 raise NotImplementedError(f"Model doesn't support customizing batch size, but {self.test} test is providing a batch size other than DEFAULT_EVAL_BSIZE")
         elif self.dargs.accuracy:
             self.batch_size = 4 if self.batch_size > 4 else self.batch_size
-
-        if not self.batch_size:
-            raise NotImplementedError(f"Model's {'DEFAULT_TRAIN_BSIZE' if self.test == 'train' else 'DEFAULT_EVAL_BSIZE'} is not implemented.")
-
 
     def _load_metadata(self):
         relative_path = self.__class__.__module__.split(".")
--- a/torchbenchmark/util/model.py
+++ b/torchbenchmark/util/model.py
@@ -171,38 +171,48 @@ class BenchmarkModel(metaclass=PostInitProcessor):
         assert hasattr(self, 'DEFAULT_NUM_BATCH'), f"We expect all models with dynamic shapes specify field `DEFAULT_NUM_BATCHES`."
         return self.DEFAULT_NUM_BATCH
 
-    def _determine_batch_size(self, batch_size=None):
+    def _get_batch_size_from_metadata(self) -> Optional[str]:
+        if self.device != "cuda":
+            current_device_name = str(self.device)
+        else:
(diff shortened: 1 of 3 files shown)


torchbenchmark/util/model.py
```diff
@@ -60,8 +60,10 @@ def enable_profiling_executor():
         torch._C._get_graph_executor_optimize(graph_executor)
 
 class BenchmarkModel(metaclass=PostInitProcessor):
-    DEFAULT_TRAIN_BSIZE: Optional[int] = None
-    DEFAULT_EVAL_BSIZE: Optional[int] = None
+    # May be overwritten by individual models
+    DEFAULT_TRAIN_BSIZE = 4
```
xuzhao9, line 64: Sorry but I am kinda concerned about doing this since many models do not support customizing batch sizes and setting it to 4 will mislead users.

I am thinking we only set `DEFAULT_TRAIN_BSIZE` for `hf_Whisper`

xuzhao9 (changes requested):

xuzhao9 (approved):
The code is clearer now. Thanks for making this improvement!

torchbenchmark/util/model.py
```diff
@@ -171,36 +171,45 @@ def _determine_dynamic_num_batches(self, user_specified_num_batches: Optional[in
...
+            raise NotImplementedError(f"Model's {'DEFAULT_TRAIN_BSIZE' if self.test == 'train' else 'DEFAULT_EVAL_BSIZE'} is not implemented.")
+
         # Check if specified batch size is supported by the model
         if hasattr(self, "ALLOW_CUSTOMIZE_BSIZE") and (not getattr(self, "ALLOW_CUSTOMIZE_BSIZE")):
             if self.test == "train" and (not self.batch_size == self.DEFAULT_TRAIN_BSIZE):
-                raise NotImplementedError("Model doesn't support customizing batch size.")
+                raise NotImplementedError(f"Model doesn't support customizing batch size, but {self.test} test is providing a batch size other than DEFAULT_TRAIN_BSIZE")
             elif self.test == "eval" and (not self.batch_size == self.DEFAULT_EVAL_BSIZE):
-                raise NotImplementedError("Model doesn't support customizing batch size.")
+                raise NotImplementedError(f"Model doesn't support customizing batch size, but {self.test} test is providing a batch size other than DEFAULT_EVAL_BSIZE")
```
eellison, line 212: I don't understand why a model would have `ALLOW_CUSTOMIZE_BSIZE` but we would end up in this branch. For context, I'm looking into why we are not running stable_diffusion_unet in inference

eellison: Why don't we just use the batch size of the model instead of failing ?

xuzhao9: If `ALLOW_CUSTOMIZE_BSIZE = False`, the model will only accept the default batch size, not the batch size specified by the user.

We could silently use the default batch size instead of failing, but my concern this will cause misunderstanding on the user side (for example, they might think the model is running in batch size 100, but `ALLOW_CUSTOMIZE_BSIZE = False` and the default batch size is 1, so it will run silently with batch size 1)

eellison: If batch_size is passed in as None, it seems okay to use the default specified on the model, instead of specified in `self.metadata["devices"][current_device_name][device_batch_size_key]`

eellison: Also, if we're worried about that case, we should also fix this upstream handling of it: https://github.com/pytorch/pytorch/blob/main/benchmarks/dynamo/torchbench.py#L369-L374

xuzhao9: > If batch_size is passed in as None, it seems okay to use the default specified on the model, instead of specified in `self.metadata["devices"][current_device_name][device_batch_size_key]`

Right, this is a bug. If `ALLOW_CUSTOMIZE_BSIZE = False` and `batch_size` passed in as None, we should use the default specified on the model instead of the device-specified batch size.

PR #1957, opened 2023-10-03
[backends] Fix Torch-TRT backend and userbenchmark

gs-olive:
- Improve error checking in benchmark
- Improve argument handling and processing in backend

### DIFF
--- a/torchbenchmark/util/backends/trt.py
+++ b/torchbenchmark/util/backends/trt.py
@@ -29,7 +29,7 @@ def parse_torch_trt_args(backend_args: List[str]):
     arg_parser.add_argument(
         "--ir",
         type=str,
-        help="Which internal representation to use: {'ts', 'dynamo_compile', 'fx_ts_compat', ...}",
+        help="Which internal representation to use: {'torch_compile', 'dynamo', 'ts', ...}",
     )
     args, unknown = arg_parser.parse_known_args(backend_args)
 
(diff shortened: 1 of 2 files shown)


xuzhao9 (approved):
LGTM, I am also looking into the Background_Matting model on the `set_module()` issue.

xuzhao9:
Test workflow run: https://github.com/pytorch/benchmark/actions/runs/6409029498

gs-olive:
@xuzhao9 - Thanks for running this! It looks like it worked! I am wondering where I could find TorchBench results times for Inductor or PyTorch Eager for comparison?

gs-olive:
Thanks for sharing - this is very helpful. I am wondering if you know, for performance speedup/absolute execution time in the Inductor HUD, as well as the reported times in the `eager` csv, like this: `"test_eval[vision_maskrcnn-cuda-eager]_cmem": 0.9033203125,`, what batch size or precisions are being used, or where would I find the code that controls this? I am trying to better understand how to compare the metrics and I want to make sure it is a fair comparison

xuzhao9:
@gs-olive The code controlled to generate the Inductor HUD is located at https://github.com/pytorch/pytorch/blob/main/benchmarks/dynamo/torchbench.py and https://github.com/pytorch/pytorch/blob/71beca4899dbe1e9b53f8451b8e33fb816465dcc/.ci/pytorch/test.sh#L494

PR #1959, opened 2023-10-04
Add test to the set_module() interface

xuzhao9:
All models should have a workable `set_module()` interface to work with the backends.


aaronenyeshi (approved):
LGTM!

msaroufim:
cc @xmfan might make more sense to get rid of the stable diffusion unet/text encoder division now and just have a single example

PR #1960, opened 2023-10-04
Cleanup the code in the `dynamo` userbenchmark

xuzhao9:
Skip importing the modules that are only available in the pytorch source code, not pytorch nightly release.

Test plan:
```
$ python run_benchmark.py dynamo --only alexnet --training --performance --inductor
loading model: 0it [00:05, ?it/s]
cuda train alexnet
running benchmark: 100%|█████████████████| 30/30 [00:00<00:00, 41.46it/s]
1.129x
```


desertfire:
I am not familiar with this part of code. What are we trying to achieve here? If you just need a stamp, I can do that.

xuzhao9:
> I am not familiar with this part of code. What are we trying to achieve here? If you just need a stamp, I can do that.

I am trying to make dynamobench runnable on PyTorch nightly release. Currently it imports a few packages that is only available in the pytorch source repo, but not available in the nightly release.

desertfire (approved):

PR #1961, opened 2023-10-04
Use full stable_diffusion model instead of just text_encoder and unet separately

xmfan:


xuzhao9:
Please also update the model list in the optim userbenchmark: https://github.com/pytorch/benchmark/blob/main/userbenchmark/optim/run.py#L222

torchbenchmark/models/stable_diffusion/install.py
HMellor, line 1: Errors/warnings in this file don't seem to be propagated to the user when `HUGGING_FACE_HUB_TOKEN` is not set. The top level `install.py` returns `OK` and then the tests fail with `module "diffusers" not found`.

Is this something that could be looked into while these models are being reorganised?

PR #1966, opened 2023-10-05
Upgrade pin numpy version

xuzhao9:
Upgrade the pin to numpy version for bisection.

Test plan:
Workflow: https://github.com/pytorch/benchmark/actions/runs/6425442814


aaronenyeshi (approved):
LGTM!

PR #1974, opened 2023-10-09
[userbenchmark] Broader error catching in Torch-TRT `userbenchmark`

gs-olive:
- Failures on a [recent sample run](https://github.com/pytorch/benchmark/actions/runs/6436382237) indicate that errors raised by the subprocess are not always Exceptions, but are still sometimes recoverable
- Add `except` clause to catch all such errors, short of keyboard interrupts, so the compilation can complete despite such errors


xuzhao9 (approved):
LGTM

gs-olive:
@xuzhao9 - thanks! If possible, could you start another benchmark run with this new change?

gs-olive:
Hi @xuzhao9 - I am wondering how we might be able to set up a weekly CI run of this userbenchmark? I noticed some of the other benchmarks have a `.yaml`. Would something like `ci.yaml`, with the following work?
```yaml
platform:   "gcp_a100"
schedule:   "weekly"
```

xuzhao9:
@gs-olive Yes it will work - but we need to fix the existing CI errors first: https://github.com/pytorch/benchmark/actions/runs/6518636322/job/17704274032

gs-olive:
I see - thanks for sharing this, it is likely because I am saving the error message as a string in the JSON dictionary when compilation or model building fails. Should I not save an entry at all in such situations instead?

xuzhao9:
> I see - thanks for sharing this, it is likely because I am saving the error message as a string in the JSON dictionary when compilation or model building fails. Should I not save an entry at all in such situations instead?

Yes, the metric value field in the JSON dictionary accepts float only. We suggest using a separate file or multiple separate files to save the complete error message.

gs-olive:
Thanks for the information! I have added the necessary changes to fix the string-values in the dictionary, here: #1998

PR #1976, opened 2023-10-10
Fix & update `metadata.yaml` for hf_T5` variants and `timm_nfnet`

HMellor:
Changes:
- `timm_nfnet/metadata.yaml` had a typo which has been corrected (`evice` -> `device`)
- `hf_T5`/`hf_T5_base`/`hf_T5_large` train changed so that it is only skipped on machines used by upstream CI


xuzhao9 (approved):
Thanks for fixing this!

PR #2008, opened 2023-10-23
yolov3: reduce batch size due to cudagraphs OOM

xmfan:
yolov3 w/ cudagraphs (known to use more memory) is failing perf test due to OOM (https://hud.pytorch.org/benchmark/torchbench/inductor_with_cudagraphs?startTime=Mon,%2016%20Oct%202023%2020:19:47%20GMT&stopTime=Mon,%2023%20Oct%202023%2020:19:47%20GMT&granularity=hour&mode=training&dtype=amp&lBranch=main&lCommit=0b424ee0b7bfe09e0a438a63e8336e95eea85901&rBranch=main&rCommit=29048be41ca3aa8974795d93b9ea9fd6dee415fc)

I'm reducing the batch size from 16 to 8 to keep the same batch size for all yolov3 HUD benchmarks

### DIFF
--- a/torchbenchmark/models/yolov3/__init__.py
+++ b/torchbenchmark/models/yolov3/__init__.py
@@ -30,7 +30,7 @@ class Model(BenchmarkModel):
     task = COMPUTER_VISION.SEGMENTATION
     # Original train batch size: 16
     # Source: https://github.com/ultralytics/yolov3/blob/master/train.py#L447
-    DEFAULT_TRAIN_BSIZE = 16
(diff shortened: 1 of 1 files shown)


xuzhao9 (commented):
We got to be careful here because the original code repo uses 16 as the default batch size (https://github.com/ultralytics/yolov3/blob/master/train.py#L449), and we don't know how bs=8 will affect the E2E training accuracy.

I am leaning to keep the default batch size to be consistent with upstream, and only customize it in the dynamo runner or other userbenchmarks that test CUDAGraph.

xmfan:
Moved the implementation to pytorch/benchmark/dynamo/torchbench.py: https://github.com/pytorch/pytorch/pull/111959

PR #2015, opened 2023-10-26
Default to supported model batch size instead of metadata if batch size not specified

eellison:
* __->__ #2015

This was breaking stable_diffusion_unet.

Differential Revision: [D50747937](https://our.internmc.facebook.com/intern/diff/D50747937)

### DIFF
--- a/torchbenchmark/util/model.py
+++ b/torchbenchmark/util/model.py
@@ -190,7 +190,7 @@ class BenchmarkModel(metaclass=PostInitProcessor):
     def _determine_batch_size(self, user_specified_batch_size=None):
         self.batch_size = user_specified_batch_size
 
-        if not self.batch_size and getattr(self, "ALLOW_CUSTOMIZE_BSIZE", False):
+        if not self.batch_size and getattr(self, "ALLOW_CUSTOMIZE_BSIZE", True):
             device_specified_batch_size = self._get_batch_size_from_metadata()
             self.batch_size = device_specified_batch_size
 
--- a/torchbenchmark/util/model.py
+++ b/torchbenchmark/util/model.py
@@ -188,9 +188,6 @@ class BenchmarkModel(metaclass=PostInitProcessor):
             return batch_size
 
     def _determine_batch_size(self, user_specified_batch_size=None):
-        # batch size priority for eval tests: not ALLOW_CUSTOMIZE_BSIZE > user specified > device specified > default
-        # batch size priority for train tests: not ALLOW_CUSTOMIZE_BSIZE > user specified > default
-
         self.batch_size = user_specified_batch_size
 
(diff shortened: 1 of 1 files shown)


torchbenchmark/util/model.py
```diff
@@ -193,7 +193,7 @@ def _determine_batch_size(self, user_specified_batch_size=None):
```
xmfan, line 193: could we also update the comment here:
```
# batch size priority for eval tests: not ALLOW_CUSTOMIZE_BSIZE > user specified > default > device specified
```

this also gets me wondering, when is it intended to use device specified batch size @xuzhao9? I think all models specify their defaults

xmfan (approved):

msaroufim (approved):

torchbenchmark/util/model.py
eellison: I find this a little difficult to interpret `not ALLOW_CUSTOMIZE_BSIZE > user specified > default > device specified` already. What does it mean if `ALLOW_CUSTOMIZE_BSIZE == True` ?

eellison: Maybe the code is sufficient ?

xmfan: I interpreted the "not ALLOW_CUSTOMIZE_BSIZE" as a condition check, it does seem that adding expanding it would end up being pseudo code

xuzhao9 (approved):

torchbenchmark/util/model.py
```diff
@@ -188,12 +188,9 @@ def _get_batch_size_from_metadata(self) -> Optional[str]:
             return batch_size
 
     def _determine_batch_size(self, user_specified_batch_size=None):
-        # batch size priority for eval tests: not ALLOW_CUSTOMIZE_BSIZE > user specified > device specified > default
-        # batch size priority for train tests: not ALLOW_CUSTOMIZE_BSIZE > user specified > default
-
         self.batch_size = user_specified_batch_size
 
-        if not self.batch_size:
+        if not self.batch_size and getattr(self, "ALLOW_CUSTOMIZE_BSIZE", False):
```
xuzhao9, line 193: I thought this should be `getattr(self, "ALLOW_CUSTOMIZE_BSIZE", True)`: if `ALLOW_CUSTOMIZE_BSIZE` is not defined, by default we will believe the model batch size is customizable. In this case, we should use the per-device batch size by default.

xuzhao9:
@eellison Can you help import this PR to fbcode so that we can land it? Sorry but pytorch/benchmark does not use GH first for now.

eellison:
@eellison has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D50747937).

eellison:
imported !

PR #2022, opened 2023-11-02
detectron2: require coco2017-minimal data

davidberard98:
It seems like detectron2 requires coco2017-minimal data. It gets installed with vision-maskrcnn, but if you want to install requirements only for one model we need the installer scripts for the detectron models to download coco2017-minimal.

### DIFF
--- a/torchbenchmark/util/framework/detectron2/__init__.py
+++ b/torchbenchmark/util/framework/detectron2/__init__.py
@@ -54,7 +54,6 @@ def remove_tools_directory():
         pass
 
 def install_detectron2(model_name, model_dir):
-    s3_utils.checkout_s3_data("INPUT_TARBALLS", "coco128.tar.gz", decompress=True)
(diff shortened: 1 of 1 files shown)


torchbenchmark/util/framework/detectron2/__init__.py
```diff
@@ -55,6 +55,7 @@ def remove_tools_directory():
 
 def install_detectron2(model_name, model_dir):
     s3_utils.checkout_s3_data("INPUT_TARBALLS", "coco128.tar.gz", decompress=True)
```
xuzhao9, line 57: I think we could remove this? I remember detectron2 models only requires coco2017-minimal, not coco128 - I don't remember the details though.

xuzhao9 (approved):
LGTM with minor comment

torchbenchmark/util/framework/detectron2/__init__.py
davidberard98: I grepped for coco128 and couldn't find anything other than instances of `checkout_s3_data`, so I think you're right - will do that.

PR #2032, opened 2023-11-10
Fix stable_diffusion_xl fp16

BowenBao:
Rename `enable_fp16_half` to `enable_fp16` to correctly override base class method


BowenBao:
@msaroufim PTAL

xuzhao9 (approved):

PR #2038, opened 2023-11-14
Register grid buffers for Super_SloMo.

ysiraichi:
This PR registers both `gridX` and `gridY` as buffers for the `backWarp` module, which belongs to the `Super_SloMo` benchmark.

Without this PR, those buffers wouldn't be moved to a device on `model.to(device)`. This causes problems, specifically with XLA.

Test Plan:

```
PJRT_DEVICE=CPU python benchmarks/dynamo/torchbench.py --performance --trace-on-xla --backend openxla --inference --only Super_SloMo
```


xuzhao9:
Looks good to me - can we have an example command to reproduce in the PR text?

xuzhao9 (approved):

ysiraichi:
Using PyTorch torchbench.py script: `PJRT_DEVICE=CPU python benchmarks/dynamo/torchbench.py --performance --trace-on-xla --backend openxla --inference --only Super_SloMo`

Issue #1893, opened 2023-09-12
Add TorchBench to HUD and Run Periodically


atalman:

xuzhao9:
It would be great if we could add a simple dashboard that covers all userbenchmarks through fetching data from S3.

xuzhao9:
TODO:
1. Add pytorch version date to userbenchmarks so that we can track the commits by their dates
2. Update the dashboard and Rockset query to track the metrics by date

xuzhao9:
Fixed by https://github.com/pytorch/test-infra/pull/4743

PR #2039, opened 2023-11-14
Log both torch.version.git_version and torch.__version__

xuzhao9:
In the HUD dashboard, we will need to log both git version and `torch.__version__`. This is helpful to identify the date when the pytorch is built.

Address TODO item 1 in https://github.com/pytorch/benchmark/issues/1893


aaronenyeshi (approved):
LGTM

PR #2050, opened 2023-11-20
Improve bisection routine by creating GitHub issue only when the bisection is successful

xuzhao9:
This is to resolve the noisy issues of the V3 nightly workflow.
When a regression is detected, we will first kick off a bisection workflow, and only create the GitHub issue when the bisection workflow fails or finds the root cause commit. If the bisection succeeds, but it does not find any result, we will not create an issue.

Test plan:
https://github.com/pytorch/benchmark/actions/runs/7506472200


aaronenyeshi (approved):
LGTM!

PR #2061, opened 2023-11-27
Pin rotary-embedding-torch to 0.3.3

desertfire:
Summary: A different rotary-embedding-torch version may cause a graph break count difference in PT2 tests, see https://github.com/pytorch/pytorch/pull/114598 for details

### DIFF
--- a/torchbenchmark/models/DALLE2_pytorch/requirements.txt
+++ b/torchbenchmark/models/DALLE2_pytorch/requirements.txt
@@ -1,3 +1,4 @@
 git+https://github.com/lucidrains/DALLE2-pytorch@00e07b7d61e21447d55e6d06d5c928cf8b67601d
 beartype==0.15.0
+rotary-embedding-torch==0.3.3
 tensorboard


malfet (approved):

malfet:
LGTM, but I wonder if imo we should have a policy of updating those dependencies say every 6 month or so (testing against the same 0.3.3 in 2025 would be somewhat moot, wouldn't it?)

xuzhao9:
> LGTM, but I wonder if imo we should have a policy of updating those dependencies say every 6 month or so (testing against the same 0.3.3 in 2025 would be somewhat moot, wouldn't it?)
 
Is there an upstream PR in rotary-embedding-torch to fix the graph break issue? I am thinking to add a link to an upstream PR in the `requirements.txt`. When the upstream PR is closed, it means we can safely remove the version pin.

desertfire:
I don't have a plan to fix upstream rotary-embedding-torch. Have some kind of scheduled pin upgrade sounds like the right thing to do here, but it can take quite some effort given how many dependent libs we may end up with for the whole torchbench. It probably makes sense to focus on hardening torch.compile itself and eventually we will see less and less surprises like this.

PR #2067, opened 2023-11-29
Make models named argument

xuzhao9:
To run all models with a single command line, we need to make `models` a named argument.
For example, to test all models with the torchscript backend on CUDA, we can run the following command:

```
python run_benchmark.py test_bench -d cuda -t train,eval --backend torchscript
```


aaronenyeshi (approved):
LGTM!

PR #2073, opened 2023-12-01
[BE] Remove test proxy suggestion

xuzhao9:
... so that we can clone and install the packages in an environment where direct https/http access is not available.


aaronenyeshi (approved):
LGTM!

PR #2116, opened 2024-01-16
Enable torchao quantization in framework and group_bench

xuzhao9:
Summary:
Support torchao quantization code in the framework.

Add a new config `torch_ao.yaml` in the group_bench userbenchmark.

Differential Revision: D52802534

### DIFF
--- a/userbenchmark/group_bench/configs/torch_ao.yaml
+++ b/userbenchmark/group_bench/configs/torch_ao.yaml
@@ -1,0 +1,13 @@
+model: "*"
+test: eval
+device: cuda
+extra_args: --precision bf16 --torchdynamo inductor --inductor-compile-mode max-autotune
+metrics:
+  - latencies
+test_group:
+  test_batch_size_default:
+    subgroup:
+      - extra_args:
+      - extra_args: --quantization int8dynamic
+      - extra_args: --quantization int8weightonly
+      - extra_args: --quantization int4weightonly
--- a/torchbenchmark/util/backends/torchdynamo.py
+++ b/torchbenchmark/util/backends/torchdynamo.py
@@ -79,6 +79,12 @@ def parse_torchdynamo_args(dynamo_args: List[str]) -> argparse.Namespace:
         help="Enable max autotune gemm"
     )
     parser.add_argument(
+        "--inductor-compile-mode",
+        default=None,
+        choices=['max-autotune'],
+        help="torch.compile mode argument for inductor runs.",
+    )
+    parser.add_argument(
         "--torchinductor_enable_split_cat_fx_pass",
         action='store_true',
         help="enable split_cat_fx_pass in Inductor"
@@ -104,6 +110,11 @@ def parse_torchdynamo_args(dynamo_args: List[str]) -> argparse.Namespace:
         default="false",
         help="Enable triton code dump by setting torch._inductor.config.debug",
     )
+    parser.add_argument(
+        "--quantization",
+        choices=["int8dynamic", "int8weightonly", "int4weightonly"],
+        help="Apply quantization to the model before running it",
+    )
     args, extra_args = parser.parse_known_args(dynamo_args)
     return args, extra_args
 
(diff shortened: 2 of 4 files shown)


HDCharles:
correct me if i'm wrong, the dynamo benchmarks are comparing the compiled vs non compiled models and this change would be quantized+compiled vs quantized, right? We'd want to compare quantized+compiled vs compiled to get something usable.

xuzhao9:
> correct me if i'm wrong, the dynamo benchmarks are comparing the compiled vs non compiled models and this change would be quantized+compiled vs quantized, right? We'd want to compare quantized+compiled vs compiled to get something usable.

@HDCharles This change would be comparing quantized+compiled vs. compiled.

userbenchmark/group_bench/configs/torch_ao.yaml
```diff
@@ -0,0 +1,13 @@
+model: "*"
+test: eval
+device: cuda
+extra_args: --precision bf16 --torchdynamo inductor --inductor-compile-mode max-autotune
```
xuzhao9, line 4: @HDCharles In this config, we define the baseline extra_args being `--precision bf16 --torchdynamo inductor --inductor-compile-mode max-autotune`, so it will apply this to every test_group/subgroup defined below.

userbenchmark/group_bench/configs/torch_ao.yaml
```diff
@@ -0,0 +1,13 @@
...
+metrics:
+  - latencies
+test_group:
+  test_batch_size_default:
+    subgroup:
+      - extra_args:
+      - extra_args: --quantization int8dynamic
```
xuzhao9, line 11: As shown in D52802534 test plan, here are the test results:

```
Running TorchBenchModelConfig(name='resnet50', test='eval', device='cuda', batch_size=None, extra_args=['--precision', 'bf16', '--torchdynamo', 'inductor', '--inductor-compile-mode', 'max-autotune', '--quantization', 'int8dynamic'], extra_env=None, output_dir=None) ... [done]
Running TorchBenchModelConfig(name='resnet50', test='eval', device='cuda', batch_size=None, extra_args=['--precision', 'bf16', '--torchdynamo', 'inductor', '--inductor-compile-mode', 'max-autotune', '--quantization', 'int8weightonly'], extra_env=None, output_dir=None) ... [done]
Running TorchBenchModelConfig(name='resnet50', test='eval', device='cuda', batch_size=None, extra_args=['--precision', 'bf16', '--torchdynamo', 'inductor', '--inductor-compile-mode', 'max-autotune', '--quantization', 'int4weightonly'], extra_env=None, output_dir=None) ... [done]
```

They are all running with compiler enabled.

userbenchmark/group_bench/configs/torch_ao.yaml
HDCharles: got it, is it possible to see the result of a run so i can see what is catalogued?

HDCharles:
so looking at the run in the test plan, it looks really good.

I see its collecting latencies, we'd also like to collect is peak memory usage and compare everything to the compiled baseline.


Also theoretically run the test for bs=1 for the weight only quantization types. Though that's less important.

HDCharles (approved):
looks good though would like to add metric for peak cuda memory usage 

and if it can't do it already, compare to baseline.

userbenchmark/group_bench/configs/torch_ao.yaml
```diff
@@ -0,0 +1,13 @@
+model: "*"
+test: eval
+device: cuda
+extra_args: --precision bf16 --torchdynamo inductor --inductor-compile-mode max-autotune
+metrics:
+  - latencies
```
xuzhao9, line 6: To add CPU/GPU peak memory, add `cpu_peak_mem` and `gpu_peak_mem` here.

xuzhao9:
@HDCharles To add CPU and GPU memory, simply add `cpu_peak_mem` and `gpu_peak_mem` to the `metrics` section in the YAML file.

This PR is only a proof-of-concept of what it can do. We can leave further development to follow-up PRs.

PR #2123, opened 2024-01-20
Fix the A10G CI due to permission issues

xuzhao9:
After the version upgrade, the runner UID changed from 1000 to 1001, so we need to setup the directory permission correctly to fix it on A10G CI.


aaronenyeshi (approved):
LGTM!

xuzhao9:
@cota 
We changed from `python setup.py develop` to `pip install --user -e .` because setup.py is being deprecated: https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html
I can create a PR that removes `--user` and see if it passes the CI (I think it will). If it does, we can remove this.

Issue #2064, opened 2023-11-27
Failed CI on A100


xuzhao9:
The test test_llama_v2_7b_16h_example_cuda failed between 20231115 and 20231116.

Failed workflow: https://github.com/pytorch/benchmark/actions/runs/7006721966/job/19059198530

Detailed error and command to reproduce:

```
$ python run.py llama_v2_7b_16h -d cuda --accuracy
fp64 golden ref were not generated for llama_v2_7b_16h. Setting accuracy check to cosine
CUDA error: CUBLAS_STATUS_NOT_INITIALIZED when calling `cublasCreate(handle)`
CUDA error: CUBLAS_STATUS_NOT_INITIALIZED when calling `cublasCreate(handle)`
Traceback (most recent call last):
  File "/data/users/xzhao9/git/benchmark/torchbenchmark/util/env_check.py", line 510, in check_accuracy
    correct_result = run_n_iterations(
  File "/data/users/xzhao9/git/benchmark/torchbenchmark/util/env_check.py", line 395, in run_n_iterations
    _model_iter_fn(mod, inputs, contexts, optimizer, collect_outputs=False)
  File "/data/users/xzhao9/git/benchmark/torchbenchmark/util/env_check.py", line 393, in _model_iter_fn
    return forward_pass(mod, inputs, contexts, collect_outputs)
  File "/data/users/xzhao9/git/benchmark/torchbenchmark/util/env_check.py", line 370, in forward_pass
    return mod(*inputs)
  File "/home/xzhao9/.conda/envs/py38/lib/python3.8/site-packages/torch/nn/modules/module.py", line 1510, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
  File "/home/xzhao9/.conda/envs/py38/lib/python3.8/site-packages/torch/nn/modules/module.py", line 1519, in _call_impl
    return forward_call(*args, **kwargs)
  File "/home/xzhao9/.conda/envs/py38/lib/python3.8/site-packages/transformers/models/llama/modeling_llama.py", line 820, in forward
    outputs = self.model(
  File "/home/xzhao9/.conda/envs/py38/lib/python3.8/site-packages/torch/nn/modules/module.py", line 1510, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
  File "/home/xzhao9/.conda/envs/py38/lib/python3.8/site-packages/torch/nn/modules/module.py", line 1519, in _call_impl
    return forward_call(*args, **kwargs)
[...]

PR #2124, opened 2024-01-22
Do not skip `llama_v2_7b_16h` on NVIDIA A100 40G.

ysiraichi:
This PR partially reverts #2095, since #2064 seems not to be an issue anymore.


xuzhao9:
This will require another fix: https://github.com/pytorch/benchmark/pull/2123

ysiraichi:
Does it? I ran the command in #2064 with an A100, and it seems to pass. Is there anything I'm missing?

xuzhao9:
> Does it? I ran the command in #2064 with an A100, and it seems to pass. Is there anything I'm missing?

Yes, there is a bug in our A10G workflow specific to our CI environment, and https://github.com/pytorch/benchmark/pull/2123 fixes it.

PR #2130, opened 2024-01-23
Remove `--user` as per user requested.

xuzhao9:
Addresses user comment on https://github.com/pytorch/benchmark/pull/2123


cota:
LGTM. This should fix my issue -- I can double-check on my side while we wait for an Approval.

cota:
> I can double-check on my side while we wait for an Approval.

Confirmed. This fixes the issue for me.

aaronenyeshi (approved):

PR #2148, opened 2024-02-02
Add bisection interactive mode for debugging.

xuzhao9:
In https://github.com/pytorch/benchmark/actions/runs/7745905863, the bisection returns empty result where it should find the root cause commit of hf_longformer bugfix.

1. Fix a bug where bisection returns empty result: the regression detector will change the commit metrics, so we need to pass in a copy of the commit digest.
2. Add an interactive mode for easier debugging.

Test Plan:

To use the interactive mode, run (pytorch repo sources are still needed):
```
python bisection.py --work-dir ~/work/bisection/test --torch-repos-path ~/tmp/torch_srcs --config ~/work/bisection/test/regression-2024-02-01.yaml --torchbench-repo-path $PWD --output ~/work/bisection/test/output.json --interactive
```

Run the nightly bisection workflow on 2024-02-01:

https://github.com/pytorch/benchmark/actions/runs/7793771518

The workflow finds the right commit: https://github.com/pytorch/benchmark/issues/2151


aaronenyeshi (approved):
This is a great addition!

PR #2167, opened 2024-02-15
Remove the date constraint in the scheduled runs

xuzhao9:
We met an issue where the K8s cluster does not update the image for the runners.
Reported to upstream: https://github.com/actions/actions-runner-controller/discussions/3292

Before we find a workaround, we have to remove the constraint on the running date.

Test Plan:
https://github.com/pytorch/benchmark/actions/runs/7911493482


aaronenyeshi (approved):
LGTM

PR #2173, opened 2024-02-20
Update README.md

piyu200594:
Updating the command to run python filr from python filename.py to python3 filename.py


aaronenyeshi (approved):
LGTM

PR #2175, opened 2024-02-20
Further delay the jobs by 1 hour

xuzhao9:
We have found that the K8s cluster will renew all container images at 12PM EST, and at 1PM EST all runners will have the latest container image. It is not clear how to change this behavior.

To run the latest container image in our nightly jobs, we have to further delay our nightly job to start at 1PM EST


aaronenyeshi (approved):

aaronenyeshi:
LGTM!

PR #2176, opened 2024-02-21
Add moondream and llava from huggingface transformers

xuzhao9:
For llava model:

* The CPU accuracy test fail because of CI runner CPU memory OOM
* The GPU accuracy test fail because of GPU CUDA OOM


aaronenyeshi (approved):
LGTM!

PR #2182, opened 2024-03-04
Add sam_fast torchbench

jamesjwu:
* __->__ #2182

Took the same benchmark used by torchbench/sam, just using the https://github.com/pytorch-labs/segment-anything-fast/tree/main/segment_anything_fast registry instead which invokes dynamo and inductor.

Differential Revision: [D54589585](https://our.internmc.facebook.com/intern/diff/D54589585)


torchbenchmark/models/sam_fast/__init__.py
```diff
@@ -0,0 +1,71 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# This software may be used and distributed according to the terms of the GNU General Public License version 3.
+
+from ...util.model import BenchmarkModel
+from segment_anything_fast.build_sam import sam_model_fast_registry
+from segment_anything_fast.predictor import SamPredictor
```
jamesjwu, line 6: This is really the only difference between this and torchbenchmark/models/sam/__init__.py

msaroufim:
Cool let's see what CI says - keep in mind after this will also need to update the test-infra dashboard https://github.com/pytorch/test-infra/pull/4453 and also the torchbench pin in pytorch https://github.com/pytorch/pytorch/pull/106009

jamesjwu:
Hmm it looks like trying to run sam_fast with cpu fails, which maybe makes sense because we're always using the inductor backend and it's intended for triton kernel? Let me see if there's a way to only have a cuda benchmark

msaroufim:
Check out metadata.yaml - here's an example of a cpu skip https://github.com/pytorch/benchmark/blob/main/torchbenchmark/models/stable_diffusion_unet/metadata.yaml

torchbenchmark/models/sam_fast/__init__.py
```diff
@@ -0,0 +1,73 @@
...
+        self.image = cv2.imread(image_path)
+        self.image = cv2.cvtColor(self.image, cv2.COLOR_BGR2RGB)
+        self.sample_image = torch.randn((3, 256, 256)).to(device)
+
+    def get_module(self):
+        example_input = [
+            {
+                "image": self.sample_image,
+                "original_size": (256, 256),
+                "point_coords": torch.tensor([[[1,1]]], device=self.device),
```
jamesjwu, line 42: Had to add some dummy point coords because segment_anything_fast actually requires at least one of point_coords, boxes, or masks. Otherwise, this line will fail since sparse_embeddings is undefined otherwise:
https://github.com/pytorch-labs/segment-anything-fast/blob/main/segment_anything_fast/modeling/prompt_encoder.py#L170

This seems like a bug since normal sam doesn't have this requirement. But working around for now.

msaroufim (approved):
Assuming CI is green stamping to unblock, you'll need to import this internally to land it

msaroufim (approved):
Assuming CI is green stamping to unblock, you'll need to import this internally to land it

msaroufim:
I don't think you introduced this error https://github.com/pytorch/benchmark/actions/runs/8146765362/job/22265984577?pr=2182#step:6:4914 - cc @xuzhao9 seems like there's an issue with MAML on master

xuzhao9:
This might be we recently upgraded cuDNN version in the pytorch build, I am looking into it.

jamesjwu:
Hmm so TestBenchmark.test_sam_fast_example_cuda is also timing out on that PR for some reason, not sure if it's being confounded by the other issues; it doesn't time out locally for me though. Let me try rebasing. Is the best option if it continues to time out to increase the timeout?

xuzhao9:
> Hmm so TestBenchmark.test_sam_fast_example_cuda is also timing out on that PR for some reason, not sure if it's being confounded by the other issues; it doesn't time out locally for me though. Let me try rebasing. Is the best option if it continues to time out to increase the timeout?

Currently we set timeout to 5 mins, we would not increase the test timeout in this case, but we could skip the test by adding "not_implemented: - example" to `metadata.yaml`.

The CUDNN issue should be fixed, rerunning the CI to see if the timeout issue persists.

xuzhao9 (approved):

bhack:
Can you check why the CI is failing on `sam_fast` at https://github.com/pytorch/benchmark/pull/2204 ?

bhack:
```python
 ======================================================================
FAIL: test_sam_fast_eval_cuda (__main__.TestBenchmark.test_sam_fast_eval_cuda)
  File "/benchmark/test.py", line 114, in eval_fn
    with task.watch_cuda_memory(
  File "/home/runner/miniconda3/envs/torchbench/lib/python3.11/contextlib.py", line 144, in __exit__
    next(self.gen)
  File "/benchmark/torchbenchmark/__init__.py", line 627, in watch_cuda_memory
    assert_equal(
AssertionError: 0 != 1024

----------------------------------------------
```

xuzhao9:
The error should be unrelated to the PR. It says there is memory leak in this model (CUDA memory not recycled after we deleted the model instance.)

PR #2187, opened 2024-03-07
workaround to make torch dynamo context manager see no_grad setting 

snadampal:
this is required to enable torch inductor fx pass to trigger graph optimizations.

For example, for this test
`python run_benchmark.py cpu --model hf_Bert --test eval --torchdynamo inductor`

### DIFF
--- a/torchbenchmark/util/framework/huggingface/model_factory.py
+++ b/torchbenchmark/util/framework/huggingface/model_factory.py
@@ -101,6 +101,9 @@ class HuggingFaceModel(BenchmarkModel):
     def eval(self) -> Tuple[torch.Tensor]:
         with torch.no_grad():
             with self.amp_context():
+                # TODO: mysterious workaround: accessing torch._inductor.config.cpp.weight_prepack
+                # here is making the torch dynamo context manager see no_grad() setting correctly
+                print("weight_prepack", torch._inductor.config.cpp.weight_prepack)
                 out = self.model(**self.example_inputs)
         # logits: prediction scores of language modeling head
         # https://github.com/huggingface/transformers/blob/v4.16.2/src/transformers/modeling_outputs.py#L455


williamwen42 (approved):

xuzhao9:
Some special models will require grad enabled in eval test, e.g., maml. Is there a way to re-enable this for those special models?

snadampal:
thanks for the review, @xuzhao9 . at torch.inductor level i saw `autograd` coming as enabled for HuggingFaceModels eval tests.
However, the `HuggingFaceModel` class already has `with torch.no_grad():`
something else might be missing then.

```
class HuggingFaceModel(BenchmarkModel):
..
..
..
    def eval(self) -> Tuple[torch.Tensor]:
        with torch.no_grad():
            with self.amp_context():
                out = self.model(**self.example_inputs)
        # logits: prediction scores of language modeling head
        # https://github.com/huggingface/transformers/blob/v4.16.2/src/transformers/modeling_outputs.py#L455
        # transformations such as fx2trt will cast the original output type to dict
        if isinstance(out, tuple):
            return out
        elif hasattr(out, 'logits'):
            return (out.logits, )
        else:
            return (out["logits"], )

```

snadampal:
Hi @xuzhao9, I have updated the PR to disable autograd selectively for HuggingFace and few BERT models. Based on my debugging, it looks like grad disable during the eval is not sufficient, because the graph compilation is happening before.

xuzhao9:
@snadampal  Sorry can you better explain why graph compilation happens before running `eval()`? We are decorating the `eval()` function with `torch.compile()`, and I think the graph compilation would happen when running `eval()`.

snadampal:
Hi @xuzhao9 , I had the same understanding that on the first call to model inference is where the compilation gets triggered, but even with decorating eval with torch.no_grad(), 
```
    def eval(self) -> Tuple[torch.Tensor]:
        with torch.no_grad():
            with self.amp_context():
                out = self.model(**self.example_inputs)
```
I see that torch dynamo context is coming with auto grad enabled. for example, in the below check, I still see torch.is_grad_enabled() is enabled.
https://github.com/pytorch/pytorch/blob/main/torch/_inductor/compile_fx.py#L1279

snadampal:
I found that it is required to call the base model eval with no grad to make the inductor to see the grad disabled.
@xuzhao9 , can you please review?

xuzhao9:
@snadampal Sorry I still do not get why it requires the base class to call with `no_grad()` to make inductor see the grad disabled, since we are applying dynamo to the `eval()` function: https://github.com/pytorch/benchmark/blob/main/torchbenchmark/util/backends/torchdynamo.py#L209. 
This will break the models that do require grad in `eval()`.

snadampal:
Hi @xuzhao9 , I'm suspecting some race condition with setting these grad variables, for example, if I just print the `torch._inductor.config.cpp.weight_prepack` in eval, dynamo context manager is picking the no_grad config. the behavior is the same on aarc64 and x86 machines, so no architecture specific race conditions here.

```
    def eval(self) -> Tuple[torch.Tensor]:
        with torch.no_grad():
            with self.amp_context():
                print("torch._inductor.config.cpp.weight_prepack", torch._inductor.config.cpp.weight_prepack)                         
                out = self.model(**self.example_inputs)
```
Does it give any clue?  i'm not very familiar with dynamo internals.
thanks for the `pick_grad()`, I will use this as a workaround if this can't be root-caused

xuzhao9:
@snadampal I still hope that we can find a minimal reproduction of the failure and report it upstream. Can you please try with that?

snadampal:
the CI failures don't seem to be related to this PR

xuzhao9:
This should be replaced by https://github.com/pytorch/benchmark/pull/2256

snadampal:
closing this as the other [PR](https://github.com/pytorch/benchmark/pull/2256) got merged.

xuzhao9:
Closed by https://github.com/pytorch/benchmark/pull/2256

PR #2189, opened 2024-03-08
Allow setting dtype in sam_fast

jamesjwu:
* __->__ #2189

### DIFF
--- a/torchbenchmark/models/sam_fast/__init__.py
+++ b/torchbenchmark/models/sam_fast/__init__.py
@@ -15,7 +15,7 @@ class Model(BenchmarkModel):
     task = COMPUTER_VISION.SEGMENTATION
     DEFAULT_EVAL_BSIZE = 32
 
-    def __init__(self, test, device, batch_size=1, extra_args=[]):
+    def __init__(self, test, device, batch_size=1, dtype=torch.bfloat16, extra_args=[]):
         super().__init__(
             test=test, device=device, batch_size=batch_size, extra_args=extra_args
         )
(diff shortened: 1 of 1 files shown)


xuzhao9:
This parameter is not needed. On the framework level, we already support using `--precision <dtype>` to cast any example input types to the user specified dtype: https://github.com/pytorch/benchmark/blob/main/torchbenchmark/util/model.py#L355.

For example, to run sam_fast with bf16:

```
python run.py sam_fast -d cuda --precision bf16
```

jamesjwu:
Hmm so the context here is that I'm trying to run torchbench in the main repo like this: 
```
benchmarks/dynamo/torchbench.py --inference --performance --backend=inductor --explain --only sam_fast
```
Unfortunately, when trying to validate the "eager mode" execution of this model, it runs into the dynamo call in the image encoder, which fails because, while trying to inline, dynamo tries to pass a FakeTensor with dtype float32 to the convolution which is set to bfloat16. Because this is in dynamo using a FakeTensor, it looks like it's bypassing any of the stuff that casts arguments to the correct type. 

Stack trace:
https://gist.github.com/jamesjwu/c6a92b00ea15807c13ca68a75ea9fc6b


We torch.compile here after we've already set the model to use bf16 (presumably using bf16 for its bias): https://github.com/pytorch-labs/segment-anything-fast/blob/main/segment_anything_fast/build_sam.py#L78. 

Dynamo doesn't actually run until the image encoder gets passed the input image. And when dynamo runs, it passes a FakeTensor to self.proj in an attempt to inline here: https://github.com/pytorch-labs/segment-anything-fast/blob/main/segment_anything_fast/modeling/image_encoder.py#L251C18-L252C1

Thing is, not sure if it's a dynamo bug or something else, but the FakeTensor's dtype when compiling that inline is always float32, even if the example input is bfloat16, with any of the methods we've mentioned above. Then, when it tries to inline that function by calling the original nn.Module (conv2d) on the inputs, it understandably crashes and dynamo reraises the error.

I don't really understand why the FakeTensor for this input isn't using bfloat16, nor why running the unit test in the benchmark repo succeeds just fine doing basically the same thing, unless those use amp or something? 



So I was attempting to add this dtype here in hopes that I could pass in float32 as the prepared dtype instead of bfloat16 in torchbench.py. Unfortunately, adding the precision as an extra argument to the model didn't work, since the FakeTensor dynamo is passing doesn't seem to touch that codepath. Any tips?

xuzhao9:
Torchbench's assumption is that before compile, both model and example_inputs are already bf16. If in this case dynamo is still passing in FakeTensor with fp32, I believe this is a dynamo bug

jamesjwu:
I won't need to do it this way, but the original idea was that I would actually pass in dtype=float32 here and use that. But I agree it's a dynamo bug needing further evalutation. 

Either way, this PR won't be necessary

PR #2198, opened 2024-03-15
Fix amp_bf16 train with staged_train_test

yanbing-j:
This PR is to fix the issue of amp_bf16 train with staged_train_test in CPU. Need set `forward_contexts` correctly with `torch.cpu.amp.autocast(dtype=torch.bfloat16)`, otherwise, in staged_train_test, model cannot run into bf16 successfully.

### DIFF
--- a/torchbenchmark/util/extra_args.py
+++ b/torchbenchmark/util/extra_args.py
@@ -116,6 +116,7 @@ def apply_decoration_args(model: 'torchbenchmark.util.model.BenchmarkModel', dar
             import torch
             model.add_context(lambda: torch.cuda.amp.autocast(dtype=torch.float16), stage=TEST_STAGE.FORWARD)
     elif dargs.precision == "amp_bf16":
+        assert model.device == "cpu", "amp_bf16 is only supported on cpu device."
         if model.test == "eval":
             import torch
             model.add_context(lambda: torch.cpu.amp.autocast(dtype=torch.bfloat16))
@@ -124,8 +125,10 @@ def apply_decoration_args(model: 'torchbenchmark.util.model.BenchmarkModel', dar
                 import torch
                 model.add_context(lambda: torch.cpu.amp.autocast(dtype=torch.bfloat16), stage=TEST_STAGE.FORWARD)
             else:
-                import torch
-                model.add_context(lambda: torch.cpu.amp.autocast(dtype=torch.bfloat16))
+                if hasattr(model, 'enable_amp'):
+                    model.enable_amp()
+                else:
+                    assert False, f"model has no enable_amp()"
     elif not dargs.precision == "fp32":
         assert False, f"Get an invalid precision option: {dargs.precision}. Please report a bug."
 
--- a/torchbenchmark/util/extra_args.py
+++ b/torchbenchmark/util/extra_args.py
@@ -32,7 +32,10 @@ def check_precision(model: 'torchbenchmark.util.model.BenchmarkModel', precision
         if model.test == 'train' and model.device == 'cuda':
             return hasattr(model, 'enable_amp') or is_staged_train_test(model)
     if precision == "amp_bf16":
-        return model.device == 'cpu'
+        if model.test == 'eval' and model.device == 'cpu':
+            return True
+        if model.test == 'train' and model.device == 'cpu':
+            return hasattr(model, 'enable_amp') or is_staged_train_test(model)
     assert precision == "fp32", f"Expected precision to be one of {AVAILABLE_PRECISIONS}, but get {precision}"
     return True
 
@@ -113,8 +116,16 @@ def apply_decoration_args(model: 'torchbenchmark.util.model.BenchmarkModel', dar
             import torch
             model.add_context(lambda: torch.cuda.amp.autocast(dtype=torch.float16), stage=TEST_STAGE.FORWARD)
     elif dargs.precision == "amp_bf16":
-        import torch
-        model.add_context(lambda: torch.cpu.amp.autocast(dtype=torch.bfloat16))
+        if model.test == "eval":
+            import torch
+            model.add_context(lambda: torch.cpu.amp.autocast(dtype=torch.bfloat16))
+        elif model.test == "train":
+            if is_staged_train_test(model):
+                import torch
+                model.add_context(lambda: torch.cpu.amp.autocast(dtype=torch.bfloat16), stage=TEST_STAGE.FORWARD)
+            else:
+                import torch
+                model.add_context(lambda: torch.cpu.amp.autocast(dtype=torch.bfloat16))
     elif not dargs.precision == "fp32":
         assert False, f"Get an invalid precision option: {dargs.precision}. Please report a bug."
 
--- a/torchbenchmark/util/framework/vision/model_factory.py
+++ b/torchbenchmark/util/framework/vision/model_factory.py
@@ -71,10 +71,14 @@ class TorchVisionModel(BenchmarkModel):
         return self.model, self.example_inputs
 
     def forward(self):
-        self.example_outputs = (torch.rand_like(self.model(*self.example_inputs)), )
+        with torch.no_grad():
+            self.example_outputs = (torch.rand_like(self.model(*self.example_inputs)), )
         for data, target in zip(self.example_inputs, self.example_outputs):
-            pred = self.model(data)
-            return self.loss_fn(pred, target)
+            # Alexnet returns non-grad tensors in forward pass
+            # Force to call requires_grad_(True) here
+            pred = self.model(data).requires_grad_(True)
+            u = self.loss_fn(pred, target)
+            return u
 
     def backward(self, loss):
         loss.backward()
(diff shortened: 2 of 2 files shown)


yanbing-j:
Hi @xuzhao9 @chuanqi129 , could you please help review this PR? Thanks!

chuanqi129 (approved):

torchbenchmark/util/extra_args.py
```diff
@@ -113,8 +116,16 @@ def apply_decoration_args(model: 'torchbenchmark.util.model.BenchmarkModel', dar
             import torch
             model.add_context(lambda: torch.cuda.amp.autocast(dtype=torch.float16), stage=TEST_STAGE.FORWARD)
     elif dargs.precision == "amp_bf16":
-        import torch
-        model.add_context(lambda: torch.cpu.amp.autocast(dtype=torch.bfloat16))
+        if model.test == "eval":
+            import torch
+            model.add_context(lambda: torch.cpu.amp.autocast(dtype=torch.bfloat16))
```
xuzhao9, line 121: This context only applies when the device is `cpu`, is that correct?

torchbenchmark/util/framework/vision/model_factory.py
```diff
@@ -71,8 +71,7 @@ def get_module(self):
         return self.model, self.example_inputs
 
     def forward(self):
-        with torch.no_grad():
-            self.example_outputs = (torch.rand_like(self.model(*self.example_inputs)), )
+        self.example_outputs = (torch.rand_like(self.model(*self.example_inputs)), )
```
xuzhao9, line 74: I am wondering why we are removing the `torch.no_grad()` context here? They are added by https://github.com/pytorch/benchmark/pull/1558

xuzhao9 (changes requested):

torchbenchmark/util/extra_args.py
yanbing-j: Yes, the device check is in https://github.com/pytorch/benchmark/pull/2198/files#diff-6ccf656b90a64ee9ee4e55aec794320710e717b65271baeae74e69940524bb6aR35.

torchbenchmark/util/framework/vision/model_factory.py
yanbing-j: vision forward will be called in `_invoke_staged_train_test`, since vision models train is staged train https://github.com/pytorch/benchmark/blob/main/torchbenchmark/util/model.py#L299. It will call `backward` then. However, `torch.no_grad` will make the input no grad, so `backward` will run into failure.

Here I have a concern as well, why there is no failure of `amp_fp16` in cuda device when running vision training test.

torchbenchmark/util/extra_args.py
```diff
@@ -113,8 +116,16 @@ def apply_decoration_args(model: 'torchbenchmark.util.model.BenchmarkModel', dar
             import torch
             model.add_context(lambda: torch.cuda.amp.autocast(dtype=torch.float16), stage=TEST_STAGE.FORWARD)
     elif dargs.precision == "amp_bf16":
-        import torch
-        model.add_context(lambda: torch.cpu.amp.autocast(dtype=torch.bfloat16))
+        if model.test == "eval":
```
xuzhao9, line 119: Can we add an assert before line 119? 
```suggestion
        assert model.device == "cpu", "amp_bf16 is only supported on cpu device."
        if model.test == "eval":
```

torchbenchmark/util/extra_args.py
xuzhao9: Can we add an assert here? (see above comment)

torchbenchmark/util/extra_args.py
```diff
@@ -113,8 +116,16 @@ def apply_decoration_args(model: 'torchbenchmark.util.model.BenchmarkModel', dar
...
+            import torch
+            model.add_context(lambda: torch.cpu.amp.autocast(dtype=torch.bfloat16))
+        elif model.test == "train":
+            if is_staged_train_test(model):
+                import torch
+                model.add_context(lambda: torch.cpu.amp.autocast(dtype=torch.bfloat16), stage=TEST_STAGE.FORWARD)
+            else:
+                import torch
+                model.add_context(lambda: torch.cpu.amp.autocast(dtype=torch.bfloat16))
```
xuzhao9, line 128: Here we are applying the `amp.autocast()` to the entire train iteration. Is this what we want?

I thought we should call `enable_amp()` if it exists, and throw an exception if it is not.

torchbenchmark/util/framework/vision/model_factory.py
xuzhao9: I am thinking here we are generating `example_output`, not `example_input`. The `example_input` tensor is still with grad. Is my understanding correct?

torchbenchmark/util/extra_args.py
yanbing-j: Done.

torchbenchmark/util/extra_args.py
yanbing-j: Done.

torchbenchmark/util/extra_args.py
yanbing-j: Done.

torchbenchmark/util/framework/vision/model_factory.py
yanbing-j: Hi @xuzhao9 , thanks for the comments.
```
    def forward(self):
        with torch.no_grad():
            self.example_outputs = (torch.rand_like(self.model(*self.example_inputs)), )
        for data, target in zip(self.example_inputs, self.example_outputs):
            pred = self.model(data)
            return self.loss_fn(pred, target)
```

This is a confused issue. With torch.no_grad(), when running `python run.py alexnet -d cpu -t train --precision amp_bf16`, it will encounter the failure as below:
```
  File "run.py", line 623, in <module>
    main()  # pragma: no cover
  File "run.py", line 593, in main
    run_one_step(
  File "run.py", line 173, in run_one_step
    func()
  File "/workspace/benchmark/torchbenchmark/util/model.py", line 309, in invoke
    return self._invoke_staged_train_test(num_batch=self.num_batch)
  File "/workspace/benchmark/torchbenchmark/util/model.py", line 301, in _invoke_staged_train_test
    self.backward(losses)
  File "/workspace/benchmark/torchbenchmark/util/framework/vision/model_factory.py", line 81, in backward
    loss.backward()
  File "/workspace/pytorch/torch/_tensor.py", line 525, in backward
    torch.autograd.backward(
  File "/workspace/pytorch/torch/autograd/__init__.py", line 267, in backward
    _engine_run_backward(
  File "/workspace/pytorch/torch/autograd/graph.py", line 744, in _engine_run_backward
    return Variable._execution_engine.run_backward(  # Calls into the C++ engine to run the backward pass
RuntimeError: element 0 of tensors does not require grad and does not have a grad_fn
```
I print out `pred.requires_grad` and `target.requires_grad`, they are all `False`. Then in backward, it will fail. It's weird that in a simple UT of the forward logic, `pred.requires_grad` is always True. But in torchbench, `pred.requires_grad` will be False, it has to remove `with torch.no_grad()`.
```
import torch
model = torch.nn.Linear(2, 2)
input = torch.randn(2, 2)
inputs = [input]

with torch.no_grad():
    outputs = (torch.rand_like(model(*inputs)), )
for data, target in zip(inputs, outputs):
    pred = model(data)

print(data.requires_grad)
print(pred.requires_grad)
print(target.requires_grad)
```
I'm not sure how to fix this in torchbench. Could you please help on this?

xuzhao9: Can I have the command to reproduce this failure? I tested in https://github.com/pytorch/benchmark/pull/2206, and it actually runs without throwing the backward/gradient exception.

yanbing-j: @xuzhao9 `python run.py alexnet -d cpu -t train --precision amp_bf16` can reproduce the failure.

xuzhao9: @yanbing-j Yeah strangely this only happens on alexnet, not `resnet18` or other torchvision models. I am looking into it, perhaps it is a torchvision bug.

xuzhao9: @yanbing-j I am thinking if this is only `alexnet` issue, we could just keep `with torch.no_grad()` and deal with this model later.

yanbing-j: @xuzhao9 Thanks for the investigation. Could you please give some suggestions of dealing with `alexnet` later?

xuzhao9: @yanbing-j I think I have found the issue. For alexnet, the `pred = self.model(data)` call returns a tensor without grad even if `data` requires grad. Changing it to `pred = self.model(data).requires_grad_(True)` will fix this issue. Can you help try it?

xuzhao9: Here is an example of my change that passes the test: https://github.com/pytorch/benchmark/pull/2206/files

xuzhao9 (approved):
LGTM!

torchbenchmark/util/framework/vision/model_factory.py
xuzhao9: Thanks! The PR looks good to me now

Issue #2049, opened 2023-11-20
I encountered an issue with the `torchbenchmark` package. When installed, it is only importable within the project directory and not from other directories. This behavior seems unexpected.


clemente0731:
ModuleNotFoundError                       Traceback (most recent call last)
Cell In[2], line 1
----> 1 import torchbenchmark

File <string>:11, in _custom_import(module_name, globals, locals, fromlist, level)

File ~/miniconda/envs/python38_torch201_cuda/lib/python3.8/site-packages/torchbenchmark/__init__.py:18
     14 from urllib import request
     16 import torch
---> 18 from components._impl.tasks import base as base_task
     19 from components._impl.workers import subprocess_worker
     20 from . import models

File <string>:11, in _custom_import(module_name, globals, locals, fromlist, level)

ModuleNotFoundError: No module named 'components'

Try to import torchbenchmark from a directory outside the project.
python
Copy code
import torchbenchmark
Expected Behavior:

I expect to be able to import the torchbenchmark package from any directory after it has been installed.

Actual Behavior:

The import works only within the project directory and fails from other directories.

Environment:

Python version: [e.g., 3.8]
Operating System: [e.g., Windows, Linux]
Other relevant information: [e.g., virtual environment, package versions]
Additional Information:

Include any additional information that might be relevant to understanding or resolving the issue.

Note:

Feel free to attach any screenshots or additional details that can help in diagnosing the problem.

css
Copy code

Make sure to replace placeholders like `[e.g., 3.8]` and `[e.g., Windows, Linux]` with the spe

xuzhao9:
Right, we only tested to import `torchbenchmark.models.*` packages, not `torchbenchmark` it self.

We can work on adding this support.

ColinPeppler:
Similar issue, I'm running `~/local/pytorch/benchmarks/dynamo/run_all.sh` but seeing the same `ModuleNotFoundError`.

I haven't found a workaround for this yet.

xuzhao9:
https://github.com/pytorch/benchmark/pull/2201 should fix this problem, can you help try it?

PR #2201, opened 2024-03-16
Move `components` to `torchbenchmark/_components`

xuzhao9:
When running as a package and user changed to a different directory, we need to also package the `components` directory.

Fixes https://github.com/pytorch/benchmark/issues/2049


aaronenyeshi (approved):
LGTM

PR #2204, opened 2024-03-21
Unet: disable deterministic on training

bhack:
It seems that the backward of some upsample cuda kernels is not deterministic:
https://github.com/pytorch/pytorch/pull/121324#issuecomment-2011091426

/cc @ezyang

### DIFF
--- a/torchbenchmark/models/pytorch_unet/__init__.py
+++ b/torchbenchmark/models/pytorch_unet/__init__.py
@@ -6,7 +6,6 @@ import torch.nn.functional as F
 from torch import optim
 from typing import Tuple
 
-torch.backends.cudnn.deterministic = True
 torch.backends.cudnn.benchmark = False
 
 from .pytorch_unet.unet import UNet
@@ -89,6 +88,7 @@ class Model(BenchmarkModel):
             self.model = torch.jit.script(self.model)
 
     def eval(self) -> Tuple[torch.Tensor]:
+        torch.backends.cudnn.deterministic = True
         self.model.eval()
         with torch.no_grad():
             with torch.cuda.amp.autocast(enabled=self.args.amp):
--- a/torchbenchmark/models/pytorch_unet/metadata.yaml
+++ b/torchbenchmark/models/pytorch_unet/metadata.yaml
@@ -5,4 +5,4 @@ eval_benchmark: false
 eval_deterministic: true
 eval_nograd: true
 train_benchmark: false
-train_deterministic: true
+train_deterministic: false


bhack:
It seems it doesn't like train deterministic `fale` and:
https://github.com/pytorch/benchmark/blob/2196021e9bc0b72a547121bbf298ae854a85a21a/torchbenchmark/models/pytorch_unet/__init__.py#L9

Is there a way to enable it only at inference?

ezyang:
This is the wrong place to put it, I think; want to put it in the benchmark runner in pytorch/pytorch

bhack:
Also, the CI failures are not related to this PR.

ezyang:
There's a separate benchmark runner in pytorch/pytorch, and you should be able to set this on unet only there. It will be easier to deploy too, since making the change here means you also have to then bump the torchbench hash on pytorch/pytorch lol

bhack:
The problem is that it seems it is not the same in the pytorch runner. 

It seems we don't have the meta/machinery to disable the determinism only on training/backward.

Also the PR was reverted cause it was failing here/HUD and not on pytorch/pytorch. 

Is this interpretation correct?

xuzhao9:
> The problem is that it seems it is not the same in the pytorch runner.
> 
> It seems we don't have the meta/machinery to disable the determinism only on training/backward.
> 
> Also the PR was reverted cause it was failing here/HUD and not on pytorch/pytorch.
> 
> Is this interpretation correct?

The failing CI in this PR is because of a separate pytorch/pytorch issue: https://github.com/pytorch/pytorch/issues/122575. There is a CUDA memleak bug caused by another pytorch PR.

bhack:
> The failing CI in this PR is because of a separate pytorch/pytorch issue: https://github.com/pytorch/pytorch/issues/122575. There is a CUDA memleak bug caused by another pytorch PR.

Thanks but it was unrelated to the mentioned comment as it was more related to why we need to use this PR here instead on `pytorch/pytorch`.

lezcano:
can you give more context? In what way is the eager implementation non-deterministic?

bhack:
We had some eager twice reproducibility issue emerged at:
https://github.com/pytorch/pytorch/pull/121324#issuecomment-2004585356

xuzhao9 (approved):
LGTM

torchbenchmark/models/pytorch_unet/__init__.py
```diff
@@ -89,6 +88,7 @@ def jit_callback(self):
             self.model = torch.jit.script(self.model)
 
     def eval(self) -> Tuple[torch.Tensor]:
+        torch.backends.cudnn.deterministic = True
```
xuzhao9, line 91: This function is not being used by the downstream `benchmarks/dynamo/torchbench.py`. So I am wondering how this change will fix the CI issue in pytorch/pytorch? @bhack

xuzhao9:
I am fine with changing it here, so long as the `eval()` and `train()` code is consistent with the `metadata.yaml`.

However, I suggest:

1. In the upstream code (https://github.com/milesial/Pytorch-UNet/blob/master/train.py), there is no specific change on `torch.backends.cudnn.deterministic`, so it seems we should avoid changing the default value at all, to keep it consistent with upstream.

2. Please note that the CI in `pytorch/pytorch` does not call `eval()`. Instead, it only calls `get_module()` and uses its own benchmark function in `benchmarks/dynamo/torchbench.py`. I am wondering how changing the `eval()` function will fix the CI in `pytorch/pytorch`.

bhack:
I don't understand point 1.
Are we still manipulating it in train?

About 2. I will reply inline

torchbenchmark/models/pytorch_unet/__init__.py
bhack: The problem is that more in general pytorch repo benchmark  is not considering deterministic meta filed in this repository so it is hard to understand what to do there.

xuzhao9:
My explanation on point 1:
The goal of this repo is to have a quick test of popular PyTorch model code on single GPU device. Therefore, we would like to keep consistent behavior with upstream model code. The upstream model code of pytorch_unet is https://github.com/milesial/Pytorch-UNet. In this repo, the train code simply does not touch the config option of `torch.backends.cudnn.deterministic`, i.e., keeps using the default value. Therefore, it seems to me that we should not tweak with this option in our code either. @bhack

bhack:
> In this repo, the train code simply does not touch the config option of torch.backends.cudnn.deterministic, i.e., keeps using the default value. Therefore, it seems to me that we should not tweak with this option in our code either

I still don't understand.. Are we tweaking in train? It seems to me not.. as it was in the init ad we moved it only in the evail.

torchbenchmark/models/pytorch_unet/__init__.py
```diff
@@ -6,7 +6,6 @@
 from torch import optim
 from typing import Tuple
 
-torch.backends.cudnn.deterministic = True
 torch.backends.cudnn.benchmark = False
```
xuzhao9, line 9: I think we should also remove this line, to keep a consistent behavior with upstream.

torchbenchmark/models/pytorch_unet/__init__.py
xuzhao9, line 91: What about we remove line 91 ` torch.backends.cudnn.deterministic = True` AND line 9-10 `torch.backends.cudnn.benchmark = False` and `torch.backends.cudnn.deterministic = True` altogether?

torchbenchmark/models/pytorch_unet/metadata.yaml
```diff
@@ -5,4 +5,4 @@ eval_benchmark: false
 eval_deterministic: true
 eval_nograd: true
 train_benchmark: false
-train_deterministic: true
+train_deterministic: false
```
bhack, line 8: Why all the removal you are proposing what it will be the scope of this metadata?
Also who is controlling `eager_two_runs_differ`? See https://github.com/pytorch/pytorch/pull/121324#issuecomment-2004585356

xuzhao9: @bhack The metadata would be default value of pytorch, which is `False` for both `train` and `eval`.

xuzhao9: `torchbench.py` is controlling `eager_two_runs_differ`: https://github.com/pytorch/pytorch/blob/76a87e33a0fdd0639842218ce26943e7b4f3838b/benchmarks/dynamo/common.py#L2478

bhack: Is it on this repo too?
https://github.com/search?q=repo%3Apytorch%2Fbenchmark%20eager_two&type=code

xuzhao9: > Is it on this repo too? [github.com/search?q=repo%3Apytorch%2Fbenchmark%20eager_two&type=code](https://github.com/search?q=repo%3Apytorch%2Fbenchmark%20eager_two&type=code)

The version in `pytorch/pytorch` is the ground truth, and the pytorch/benchmark version will be auto-sync-ed from `pytorch/pytorch`.

bhack: But who is going to define if two unet eager runs need to be the same or not?

bhack: It seems that it could be deterministic when compiled if I have interpreted this correctly:
https://github.com/pytorch/pytorch/pull/121769#issuecomment-2030738866

What do you think?

xuzhao9: > But who is going to define if two unet eager runs need to be the same or not?

Sorry about the confusion. Yes, it is defined by `pytorch/benchmark`: https://github.com/pytorch/benchmark/blob/main/test.py#L72.  If two eager runs differ, the `pytorch/benchmark` CI will fail.

bhack: So as it seems to be deterministic only compiled, at least in training, what do you want to do here and in the pytorch repo?

xuzhao9: In our previous tests it is deterministic in the eager mode. If there is a PR changes this behavior and makes it only deterministic when compiled and non-deterministic in eager mode, it is up to the PyTorch team to decide if they should accept it. If the PyTorch Dev Infra team accepts it, we can skip the eager-mode deterministic test for this model.

PR #2228, opened 2024-04-11
add condition in enable_amp to fix eval not work under amp

wincent8:
when we running alexnet eval with amp, it will failed due to AttributeError: 'Model' object has no attribute 'forward_contexts'
when i check the code, i found that it will go into the is_staged_train_test and even i just want to run eval. 
so i modify the condition and it will pass then.

### DIFF
--- a/torchbenchmark/util/model.py
+++ b/torchbenchmark/util/model.py
@@ -490,7 +490,7 @@ class BenchmarkModel(metaclass=PostInitProcessor):
                 self.add_context(self.amp_context, TEST_STAGE.FORWARD)
             else:
                 warnings.warn(
-                        "Usually models do not want to enable amp only in forward path, so expected "
+                        "Usually models only want to enable AMP in forward path, so expected "
                         "model to have staged train support."
                         )
 
--- a/torchbenchmark/util/model.py
+++ b/torchbenchmark/util/model.py
@@ -483,7 +483,7 @@ class BenchmarkModel(metaclass=PostInitProcessor):
             self.amp_context = lambda: torch.cpu.amp.autocast()
         elif self.device == "cuda":
             self.amp_context = lambda: torch.cuda.amp.autocast()
-        if is_staged_train_test(self):
+        if self.test == "train" and is_staged_train_test(self):
             self.forward_contexts.append(self.amp_context)
 
     @property
--- a/torchbenchmark/util/model.py
+++ b/torchbenchmark/util/model.py
@@ -483,8 +483,18 @@ class BenchmarkModel(metaclass=PostInitProcessor):
             self.amp_context = lambda: torch.cpu.amp.autocast()
         elif self.device == "cuda":
             self.amp_context = lambda: torch.cuda.amp.autocast()
-        if self.test == "train" and is_staged_train_test(self):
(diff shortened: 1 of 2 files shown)


xuzhao9 (approved):
Thanks for fixing this!

torchbenchmark/util/model.py
```diff
@@ -483,7 +483,7 @@ def enable_amp(self):
             self.amp_context = lambda: torch.cpu.amp.autocast()
         elif self.device == "cuda":
             self.amp_context = lambda: torch.cuda.amp.autocast()
-        if is_staged_train_test(self):
+        if self.test == "train" and is_staged_train_test(self):
             self.forward_contexts.append(self.amp_context)
```
xuzhao9, line 487: I think we should handle `self.test == "eval"` here as well?
For "eval" test, we should just add it to `self.run_contexts`.

wincent8: yes, you are right. 
while i modify it, i have a question, do we need to assert error if self.test is train and is_staged_train_test is false, i see that in the amp_fp16 path, it asserts error. please refer to below code.
https://github.com/pytorch/benchmark/blob/main/torchbenchmark/util/extra_args.py#L189C9-L198

xuzhao9: Either assert error or print a warning in this case.
The reason is because usually models do not want to enable amp in the backward pass, only forward pass.

I am leaning towards print a warning message in both `--amp` and `--amp_fp16`, what do you think?

wincent8: you are right, I added the warning, please take a look.

torchbenchmark/util/model.py
```diff
@@ -483,8 +483,18 @@ def enable_amp(self):
...
+            self.add_context(self.amp_context)
+        elif self.test == "train":
+            if is_staged_train_test(self):
+                self.add_context(self.amp_context, TEST_STAGE.FORWARD)
+            else:
+                warnings.warn(
+                        "Usually models only want to enable AMP in forward path, so expected "
+                        "model to have staged train support."
+                        )
+
```
xuzhao9, line 496: In this case, we can keep this warning and add `self.amp_context` to `TEST_STAGE.ALL`. Can you help address this comment?

wincent8: try to added it, and also updated in extra_args.py in amp_fp16 path.

xuzhao9: I don't think the `amp_fp16` path added in `extra_args.py` is correct, because it is supposed to add `torch.cuda.amp.autocast(dtype=torch.float16)` context, but here we are adding the `torch.cuda.amp.autocast()` context. What do you think?

wincent8: you are right. fixed. then should we also modify the amp_bf16 path when the model not support staged train?

aaronenyeshi (approved):

torchbenchmark/util/model.py
xuzhao9: We can keep that for now. Later we may want to refactor it to clean up the code.

xuzhao9 (approved):

PR #2232, opened 2024-04-16
Prune cuda installation for docker image

FindHao:
The [`install_cuda.sh`](https://github.com/pytorch/builder/blob/main/common/install_cuda.sh#L126) in pytorch/builder provides functions `prune_121` and `prune_118` to remove unnecessary parts in the installed CUDA. 

In my local testing, the image size generated by this docker file is reduced from 23.5GB to 19.7GB.


xuzhao9 (approved):
LGTM, thanks!

PR #2235, opened 2024-04-20
Windows fixes

aaron-boxer:
This PR contains a few small fixes to allow setup to run on Windows without erroring out.

### DIFF
--- a/torchbenchmark/__init__.py
+++ b/torchbenchmark/__init__.py
@@ -59,7 +59,7 @@ def _install_deps(model_path: str, verbose: bool = True) -> Tuple[bool, Any]:
         [sys.executable, install_file],
     ]
     run_env = os.environ.copy()
-    run_env["PYTHONPATH"] = this_dir.parent
+    run_env["PYTHONPATH"] = Path(this_dir.parent).as_posix()
     run_kwargs = {
         "cwd": model_path,
         "check": True,
--- a/torchbenchmark/__init__.py
+++ b/torchbenchmark/__init__.py
@@ -67,7 +67,7 @@ def _install_deps(model_path: str, verbose: bool = True) -> Tuple[bool, Any]:
     }
 
     output_buffer = None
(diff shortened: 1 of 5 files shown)


msaroufim:
Assuming CI passes this should be fine to merge, just triggered it now

DiweiSun:
Regarding to the os.remove part, I suppose the temp file cannot be deleted since the file descriptor is not closed yet. Can we try with the following:

```
  +  fd, stdout_fpath = tempfile.mkstemp()

    try:
        output_buffer = io.FileIO(stdout_fpath, mode="w")
    except Exception as e:
        return (False, e, io.FileIO(stdout_fpath, mode="r").read().decode())
    finally:
        output_buffer.close()
        del output_buffer
        os.close(fd)
        os.remove(stdout_fpath)
```

aaron-boxer:
> Regarding to the os.remove part in **init**.py, I suppose the temp file cannot be deleted since the file descriptor is not closed yet. Can we try with the following:
> 
> ```
>    fd, stdout_fpath = tempfile.mkstemp()
> 
>     try:
>         output_buffer = io.FileIO(stdout_fpath, mode="w")
>     except Exception as e:
>         return (False, e, io.FileIO(stdout_fpath, mode="r").read().decode())
>     finally:
>         output_buffer.close()
>         del output_buffer
>         os.close(fd)
>         os.remove(stdout_fpath)
> ```

Thanks, that worked - this is a much better fix.

torchbenchmark/__init__.py
```diff
@@ -5,6 +5,7 @@
 import io
 import os
 import pathlib
+import platform
```
xuzhao9, line 8: Is this package being used in the code? Can we remove redundant dependency?

aaron-boxer: Good catch - I have removed it.

xuzhao9 (approved):
LGTM, thanks for fixing it on Windows!

PR #2243, opened 2024-04-25
Add tritonbench OSS CI

xuzhao9:


bertmaher:
Cool!  Are results going to be visible somewhere after this diff?  Or is more to come?

bertmaher (approved):

xuzhao9:
> Cool! Are results going to be visible somewhere after this diff? Or is more to come?

There will be follow-up PRs to make the results visible, stay tuned!

sijiac (approved):

PR #2254, opened 2024-05-01
Install detectron2 without build isolation

### LINKED ISSUE #2239: Unable to install with the latest `pip`
This is a dependency issue. See https://github.com/facebookresearch/detectron2/issues/4921. Either `pip<=22` or explicit `--no-build-isolation` in scripts in this repo (after explicitly installing some build-time dependencies) is required.

roastduck:
Fix #2239.

### DIFF
--- a/torchbenchmark/util/framework/detectron2/__init__.py
+++ b/torchbenchmark/util/framework/detectron2/__init__.py
@@ -43,7 +43,10 @@ def install_model_weights(model_name, model_dir):
 def pip_install_requirements():
     requirements_file = os.path.join(CURRENT_DIR, "requirements.txt")
     subprocess.check_call(
-        [sys.executable, "-m", "pip", "install", "-q", "-r", requirements_file]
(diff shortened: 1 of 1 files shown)


torchbenchmark/util/framework/detectron2/__init__.py
```diff
@@ -43,7 +43,10 @@ def install_model_weights(model_name, model_dir):
 def pip_install_requirements():
     requirements_file = os.path.join(CURRENT_DIR, "requirements.txt")
     subprocess.check_call(
-        [sys.executable, "-m", "pip", "install", "-q", "-r", requirements_file]
+        [sys.executable, "-m", "pip", "install", "-q", "wheel", "numpy", "cython"] # Build-time requirements
```
xuzhao9, line 46: In order to be compatible with the upstream pytorch build-time numpy version, we have to pin the numpy version here: https://github.com/pytorch/benchmark/blob/main/requirements.txt#L20.

Can we assume that numpy has been installed in the environment and remove numpy here?

roastduck: Now it's removed.

PR #2272, opened 2024-05-29
move detectron2 pin to the most recent tip

tenpercent:
With the existing pin, cpp extension compilation on ROCm fails.

# Testing
`$ pip install --user --verbose -r torchbenchmark/util/framework/detectron2/requirements.txt`

(ensure the extension is successfully built)


xuzhao9 (approved):
LGTM, but we need to wait for the CI signal

PR #2275, opened 2024-05-31
Add sys.executable to ncu command

davidberard98:
Previously, the ncu option would exit with code 255 for me:
```
python run_benchmark.py triton --op int4_gemm --metrics ncu_trace
```

For me, adding the python executable path into the command fixed this.

### DIFF
--- a/torchbenchmark/util/triton_op.py
+++ b/torchbenchmark/util/triton_op.py
@@ -817,7 +817,7 @@ class BenchmarkOperator(metaclass=PostInitProcessor):
         import subprocess
         from pathlib import Path
 
-        op_task_args = copy.deepcopy(sys.argv)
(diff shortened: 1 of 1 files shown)


torchbenchmark/util/triton_op.py
```diff
@@ -817,7 +817,7 @@ def ncu_trace(self, input_id: int, fn_name: str, replay: bool=False) -> str:
         import subprocess
         from pathlib import Path
 
-        op_task_args = copy.deepcopy(sys.argv)
+        op_task_args = [sys.executable] + copy.deepcopy(sys.argv)
```
xuzhao9, line 820: This might not work for fbcode command. Can you please test this using `buck run`?

Issue #2290, opened 2024-06-10
CI Failure bisection


xuzhao9:
Success: 20240515 (e5cd7bea0283f82d32097626f86475fd9c127354)
Fail: 20240606 (529bedad3e6cbd98387b9e46ce0c0e290ed12ab5)

Userbenchmark name: test_bench

Userbenchmark args: -m detectron2_maskrcnn_r_50_fpn --accuracy

python run.py detectron2_maskrcnn_r_50_fpn -d cuda -t eval --accuracy

xuzhao9:
Test workflow: https://github.com/pytorch/benchmark/actions/runs/9454579616

PR #2306, opened 2024-06-17
Fix ci by skipping the accuracy test

xuzhao9:
Between 20240515 and 20240606, detectron2 models have some issues running with the accuracy test.
We are disabling these tests to fix the CI, will look into how to enable the test.

See issue: https://github.com/pytorch/benchmark/issues/2290


aaronenyeshi (approved):
LGTM!

PR #2307, opened 2024-06-17
Pin yolo dep

eellison:
This is the cause of https://github.com/pytorch/pytorch/issues/128845. It is upgrading to numpy 2.0 which has BC incompat changes.


seemethere (approved):

xuzhao9:
@seemethere I am wondering if we should also change numpy version here? https://github.com/pytorch/builder/blob/3b032fc9e0b7a17b74eb421da52b264089e0a15e/wheel/build_wheel.sh#L144

xuzhao9:
Replaced by https://github.com/pytorch/benchmark/pull/2311

PR #2309, opened 2024-06-17
Remove torchvision and torchaudio dependency when testing userbenchmark.

xuzhao9:
Most userbenchmarks handle dependencies themselves, and torchvision/torchaudio are not required.


aaronenyeshi (approved):
LGTM!

PR #2316, opened 2024-06-19
Hardcode the numpy version

xuzhao9:
When there is a big numpy version bump (1.21.2 -> 2.0.0), `pip install` will somehow automatically upgrade numpy version to 2.0.0 even when the old version (1.21.2) has been installed.

Therefore, we have to hardcode numpy version both globally and for the models whose install will cause numpy to unexpectedly upgrade. Since downstream CI supports Python 3.8 as the lowest supported Python version, we have to pin numpy version to the lowest version used in the downstream CI.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -17,7 +17,8 @@ psutil
 pyyaml
 # We need to pin numpy version to the same as the torch testing environment
 # which still supports python 3.8
-numpy==1.21.2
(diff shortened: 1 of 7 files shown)


xuzhao9:
We probably can add a constraint to the numpy version by dynamically adding a `numpy_constraint.txt` file during the installation.

aaronenyeshi (approved):

PR #2318, opened 2024-06-19
Add package contraints to torchbench

xuzhao9:
This PR introduces package constraints because we want to guarantee that certain packages are the same version before and after the installation. This is important for the CI in pytorch/pytorch because overriding critical package versions will break the CI due to ABI incompatibility.

If we find a certain model breaks the constraints in the future, we could ask the model installer to apply the `constraint.txt` file with their requirements.txt file.

PyTorch CI Test: https://github.com/pytorch/pytorch/pull/129114


utils/python_utils.py
```diff
@@ -14,14 +17,50 @@
...
 
 
+def pip_install_requirements(requirements_txt="requirements.txt",
+                            continue_on_fail=False,
+                            no_build_isolation=False,
+                            extra_args: Optional[List[str]]=None):
+    import sys
+    constraints_file = REPO_DIR.joinpath("build", "constraints.txt")
+    if not constraints_file.exists():
+        warnings.warn("constraints.txt could not be found, please rerun install.py.")
```
huydhn, line 35: My read here is that `build/constraints.txt` is an optional constraints file that TorchBench users could provide.  But the warning here seems to says that the file is a requirement?

xuzhao9: Yes, it is an optional constraint. We could update the warning message to sth like:

The `build/constrants.txt` file is not found. Please consider rerunning the `install.py` script to generate it.
It is recommended to have the `build/constrants.txt` file to prevent unexpected version change of `numpy` or `torch`.

utils/python_utils.py
```diff
@@ -14,14 +17,50 @@
...
 }
+REPO_DIR = Path(__file__).parent.parent
 
 
 def create_conda_env(pyver: str, name: str):
     command = ["conda", "create", "-n", name, "-y", f"python={pyver}"]
     subprocess.check_call(command)
 
 
+def pip_install_requirements(requirements_txt="requirements.txt",
```
huydhn, line 28: I'm not sure if it's easy to add a verbose option here and when it's set to true, print out the output of the pip install command below.  When I used `install.py` locally, it installs stuffs silently even with `-v`, which makes it a bit slow to figure out which models install what

utils/python_utils.py
huydhn: Ah ok, I see the `generate_pkg_constraints` func above now

huydhn (approved):
This is a nice fix.  TIL, there is a pip constraint file

utils/python_utils.py
xuzhao9: Adding a verbose option sounds reasonable, but it will be non-trival to turn it on across all models, since each model has its own `install.py`.

PR #2326, opened 2024-06-21
Fix the soft_actor_critic model

xuzhao9:
Unfortunately, https://github.com/pytorch/benchmark/pull/2318 has a bug that breaks the `soft_actor_critic` model.


aaronenyeshi (approved):
LGTM

Issue #2135, opened 2024-01-26
Use CPU-only instance to run the CPU tests


xuzhao9:
Currently we are running all PR tests on A100 or A10G machines, this is a waste.
Run CPU tests on a CPU only machine, this should also fix the flaky CPU tests failed because of timeout.

xuzhao9:
Fixed by https://github.com/pytorch/benchmark/pull/2327

PR #2327, opened 2024-06-21
Improve the ci by separating CPU and CUDA tests

xuzhao9:
Currently we are running both CPU and CUDA tests in the A100/A10G runners. This is a waste of resource.
In this PR, we separate CPU and CUDA tests with different runners. The CUDA tests still run on the GCP A100 runner, while the CPU tests runs on a CPU-only runner of `linux.24xlarge`.

Reduces the CI time from ~120m to ~90m.

Fixes https://github.com/pytorch/benchmark/issues/2135


huydhn (approved):
LGTM!

PR #2351, opened 2024-07-02
Add workflow to cleanup Docker images

xuzhao9:
Update the workflow to:
1. Update the GCP base docker every month
2. Add a cleanup workflow to only keep the most recent 200 docker images.

Each docker images is ~28GB, so 200 images will take around 5.6 TB storage in total.


aaronenyeshi (approved):
LGTM

PR #2365, opened 2024-07-05
Run performance test non-alternately

xuzhao9:
Summary:
By default, performance tests (speedup experiments) will run the baseline and test backend alternately.

However, this does not work for the torchao backend, which will change the model in-place, therefore the baseline run will also run with torchao backend since the model has already been quantized.

Add a new experiment "latency_experiment" to run performance tests non-alternately (first run baseline for a few iterations, then run the test backend).

note: PR also has updates for torchao apis to make sure this works with newest version of torchao

Test command on A100:

```
Test Plan:

python run_benchmark.py torchao --only AlbertForMaskedLM --quantization noquant --performance --inference --bfloat16 --inductor-compile-mode max-autotune
python run_benchmark.py torchao --only BartForCausalLM --quantization noquant --performance --inference --bfloat16 --inductor-compile-mode max-autotune
python run_benchmark.py torchao --only timm_efficientnet --quantization noquant --performance --inference --bfloat16 --inductor-compile-mode max-autotune

(should all be ~1.0
0.997x
1.006x
0.994x
```

Differential Revision: D59332736

### DIFF
--- a/userbenchmark/dynamo/dynamobench/common.py
+++ b/userbenchmark/dynamo/dynamobench/common.py
@@ -4389,7 +4389,13 @@ def run(runner, args, original_dir=None):
                 fullgraph=args.nopython,
                 mode=args.inductor_compile_mode,
             )
-            runner.model_iter_fn = baseline_ctx(runner.model_iter_fn)
+            model_iter_fn = baseline_ctx(runner.model_iter_fn)
+            # needed to avoid error that causes inconsistent timing due to:
+            # Unable to hit fast path of CUDAGraphs because of pending, uninvoked backwards
+            def model_iter_fn_and_mark_step(*args, **kwargs):
+                torch.compiler.cudagraph_mark_step_begin()
+                model_iter_fn(*args, **kwargs)
+            runner.model_iter_fn = model_iter_fn_and_mark_step
             optimize_ctx = torchao_optimize_ctx(args.quantization)
         else:
             optimize_ctx = torch._dynamo.optimize(args.backend, nopython=args.nopython)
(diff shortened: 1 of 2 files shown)


HDCharles:
I updated the PR to conform to our newest APIs in torchao, but i'm seeing some weird stuff when i test this i.e.

python run_benchmark.py torchao --only AlbertForMaskedLM --quantization autoquant --performance --inference --bfloat16 --inductor-compile-mode max-autotune

yields 

11.846x

but on the old versions I got:

Running eval method from AlbertForMaskedLM on cuda in dynamo inductor mode with input batch size 4 and precision bf16.
GPU Time per batch:   34.484 milliseconds
CPU Wall Time per batch:  34.512 milliseconds
CPU Wall Time:        34.512 milliseconds
Time to first batch:        41578.0914 ms
GPU 3 Peak Memory:              3.4944 GB
CPU Peak Memory:                5.1348 GB
PT2 Compilation time:      32.592 seconds
autoquant
noquant
/home/cdhernandez/local/miniconda3/envs/pytorch/lib/python3.10/site-packages/pandas/core/arrays/masked.py:60: UserWarning: Pandas requires version '1.3.6' or newer of 'bottleneck' (version '1.3.5' currently installed).
  from pandas.core import (
Running eval method from AlbertForMaskedLM on cuda in dynamo inductor mode with input batch size 4 and precision bf16.
GPU Time per batch:   46.879 milliseconds
CPU Wall Time per batch:  46.914 milliseconds
CPU Wall Time:        46.914 milliseconds
Time to first batch:        12480.9602 ms
GPU 3 Peak Memory:              3.5765 GB
CPU Peak Memory:                5.0508 GB
PT2 Compilation time:      17.503 seconds
noquant 
1.359x

I tried to take a look at whats going on and it looks like maybe the non quantized model isn't being compiled?

xuzhao9:
@HDCharles Can you please import the PR as a Diff so that we can land it?

PR #2366, opened 2024-07-05
Fix v2 nightly and bisection workflows

xuzhao9:
Fix the broken v2 nightly workflow.

Test Plan:
https://github.com/pytorch/benchmark/actions/runs/10118537615/job/27985526154


aaronenyeshi (approved):
LGTM

PR #2367, opened 2024-07-05
Fallback to delete-package-versions v4 to use node16

xuzhao9:
actions/clean-package-version v5 requires node20 which pytorch infra is not ready yet.
fallback to v4 to use node 16.

Test plan: 
https://github.com/pytorch/benchmark/actions/runs/9811750556

We only keep the most recent 200 images after running the workflow: 
https://github.com/pytorch/benchmark/pkgs/container/torchbench


aaronenyeshi (approved):
LGTM!

PR #2371, opened 2024-07-09
[tritonbench] increase dynamo cache size to avoid inadvertent fallbacks

bertmaher:


bertmaher:
Duh, this is not needed, I only thought it was because I had removed the dynamo.reset while debugging something else

PR #2379, opened 2024-07-17
TMA support for FA

manman-ren:


torchbenchmark/operators/flash_attention/operator.py
```diff
@@ -109,6 +110,7 @@ class Operator(BenchmarkOperator):
 
     def __init__(self, tb_args: argparse.Namespace, extra_args: Optional[List[str]] = None):
         super().__init__(tb_args, extra_args)
+        self.use_cuda_graphs = False
```
manman-ren, line 113: Currently TMA doesn't work with cuda_graph, I wonder if there is a way of disabling it only for one variant triton_tutorial_flash_v2_tma. CC @xuzhao9

embg: Possibly dumb question since I haven't read the code: is there any way we can set up the TMA descriptors outside the benchmarked function? That would make cudagraphs possible. cc @bertmaher who did something similar recently.

torchbenchmark/util/kernels/triton_fused_attention.py
```diff
@@ -182,6 +183,164 @@ def _attn_fwd(Q, K, V, sm_scale, M, Out,  #
...
 
 
+@triton.jit
+def _attn_fwd_inner_tma(acc, l_i, m_i, q,  #
+                    K_desc_ptr, V_desc_ptr, Q, qvk_offset, stride_kn, stride_vn, stride_vk,  #
+                    start_m, qk_scale,  #
+                    BLOCK_M: tl.constexpr, HEAD_DIM: tl.constexpr, BLOCK_N: tl.constexpr,  #
+                    STAGE: tl.constexpr, offs_m: tl.constexpr, offs_n: tl.constexpr,  #
+                    N_CTX: tl.constexpr, fp8_v: tl.constexpr):
+    # range of values handled by this stage
```
embg, line 193: Nit: can you leave a comment so folks know that the required TMA fence is in `_attn_fwd_tma()`?

torchbenchmark/util/kernels/triton_fused_attention.py
```diff
@@ -182,6 +183,164 @@ def _attn_fwd(Q, K, V, sm_scale, M, Out,  #
...
+    # load scales
+    qk_scale = sm_scale
+    qk_scale *= 1.44269504  # 1/log(2)
+    # load q: it will stay in SRAM throughout
+    q = tl.load(Q_block_ptr)
+    # stage 1: off-band
+    # For causal = True, STAGE = 3 and _attn_fwd_inner gets 1 as its STAGE
+    # For causal = False, STAGE = 1, and _attn_fwd_inner gets 3 as its STAGE
+    if STAGE & 1:
+        acc, l_i, m_i = _attn_fwd_inner_tma(acc, l_i, m_i, q, desc_k, desc_v, Q, qvk_offset, stride_kn, stride_vn,
```
embg, line 320: I didn't know you can call a Triton kernel from another kernel. Super cool!

torchbenchmark/util/kernels/triton_fused_attention.py
```diff
@@ -508,4 +667,121 @@ def backward(ctx, do):
...
+        desc_k = torch.tensor(desc_k, device=v.device)
+        desc_v = torch.tensor(desc_v, device=v.device)
+        grid = lambda args: (triton.cdiv(q.shape[2], args["BLOCK_M"]), q.shape[0] * q.shape[1], 1)
+        '''
+        desc_k = torch.empty((TMA_SIZE), device="cuda", dtype=torch.int8)
+        desc_v = torch.empty((TMA_SIZE), device="cuda", dtype=torch.int8)
+        def grid_tma(META):
+            nonlocal desc_k
+            nonlocal desc_v
+            k_buf = torch.empty_like(desc_k, device="cpu")
```
embg, line 730: Let's enable async memcpy! Just add `pinned_memory=True` for all `torch.empty_like(..., device="cpu")` calls.

torchbenchmark/util/kernels/triton_fused_attention.py
```diff
@@ -508,4 +667,121 @@ def backward(ctx, do):
...
+                triton.runtime.driver.active.utils.fill_2d_tma_descriptor(
+                    v.data_ptr(),
+                    BATCH * H * N_CTX,
+                    HEAD_DIM_Q,
+                    META['BLOCK_N'],
+                    HEAD_DIM_Q,
+                    v.element_size(),
+                    v_buf.numpy(),
+                )
+            desc_k.copy_(k_buf)
```
embg, line 761: After adding `pinned_memory=True` above, you can add `non_blocking=True` here to enable async memcpy.

torchbenchmark/operators/flash_attention/operator.py
xuzhao9: @manman-ren For fair comparison, I would actually prefer to keep the same function across all kernels. If `triton_tutorial_flash_v2_tma` is not compatible with `do_bench_cudagraph()`, let's just use `do_bench()` across all kernels.

bertmaher: Yep, I did some tricks in fp8_gemm to allocate the TMA descriptors outside of the benchmarked region, let me put up a suggestion for how to do that here

bertmaher: Ok, I think you want something like this: https://gist.github.com/bertmaher/971cbae202d314913594a97e45058553

So when you're setting up the benchmark, you first call `alloc_tma_desc` to create the descriptors for k, v, then you use those in the benchmark function rather than reallocating

Is it beneficial to use TMA for q and o as well as just k v?

manman-ren: Thanks Bert! I don't see perf win for TMA with this implementation on a few cases I checked. I will try to cover more cases and report the results here.

I am starting with k and v since they are handled inside the loop. I am thinking about landing the initial implementation and land additional improvements as separate patches. What do you think? @bertmaher @embg

manman-ren:
Addressed the review comments. @bertmaher @embg Does the patch look okay to you?

embg (approved):

PR #2384, opened 2024-07-18
Add simple fused Triton kernel benchmark for jagged_softmax

jananisriram:
Summary:
Add Triton kernel benchmark implementing a simple fused `softmax` for the `jagged_softmax` operator. This Triton kernel performs a `softmax` operation along the ragged dimension of a nested tensor of logical dimensions `(B, *, M)`, where `*` is the ragged dimension.

The kernel implements `softmax` in four phases.
1. Calculate the maximum values along the ragged dimension of input using a buffer that the loop updates. Given each `(BLOCK_SIZE_RAGGED, BLOCK_SIZE_M)` block of input, find the maximum values along the ragged dimension and update `buffer_max` if necessary.
2. Looping through the same blocks of input, add stable, exponentiated blocks of input to a new buffer. Subtract from each input block the maximum value along the ragged dimension, then exponentiate this stable input and add to a buffer.
3. Calculate the `sum` of the exponentiated buffer, which will be the denominator of the `softmax` function.
4. Divide each input block, the numerator of the `softmax` function, by the `sum` of the exponentiated buffer.

As with previous jagged operators, the kernel is benchmarked against a padded PyTorch implementation and verifies accuracy against a baseline PyTorch `unbind` implementation.

This implementation uses the `buffer_then_sum` method; it adds all exponentiated input to a buffer, then takes the sum at once. This method has been proven to be faster, as seen in previous jagged operators (like `jagged_sum` and `jagged_mean`) and seems to prove that storing to and loading from many registers is faster than taking multiple `sum`s in Triton.

Other approaches I tried:
- To minimize the number of repetitive operations, specifically in the calculation of the exponentiated stable input, I tried to cache (`tl.store`) the result of this expression and load (`tl.load`) it back in the second phase (iterative loop). However, the latency almost doubled with this implementation, implying that a singular `tl.store` operation is more expensive than a singular `tl.exp` operation.
- I also tried using `tl.div_rn` to divide the exponentiated input by the buffer sum, which almost tripled or quadrupled the latency.

Notes:
- The kernel currently has three `for` loops, which may be less efficient. We probably cannot avoid this if we still want to use the buffer implementation to collect the exponentiated `sum`. As mentioned above, I did try caching, which was slower than just recalculating the required values.
- This implementation works for nested tensors where variable-length tensors have `seqlen = 0`
- A previous implementation of this kernel took the maximum value of the input from the `operator`, which did not accurately measure the time taken for the kernel to calculate the maximum value itself, nor did it accurately measure the maximum value itself. This new version includes that implementation detail (the [`softmax` source code](https://www.internalfb.com/code/fbsource/[6ebd07f65500ddfb0a2599e96dc33d91c3d88bf0]/fbcode/caffe2/torch/_refs/__init__.py?lines=3890) helped with understanding this!).

Reviewed By: davidberard98

Differential Revision: D59299726

### DIFF
--- a/torchbenchmark/operators/jagged_softmax/__init__.py
+++ b/torchbenchmark/operators/jagged_softmax/__init__.py
@@ -1,0 +1,1 @@
+from .operator import Operator
--- a/torchbenchmark/operators/jagged_softmax/kernels.py
+++ b/torchbenchmark/operators/jagged_softmax/kernels.py
@@ -1,0 +1,120 @@
+import itertools
+
+import triton
+import triton.language as tl
+
+
+BLOCK_SIZES_RAGGED = [2**n for n in range(3, 12, 4)]
+BLOCK_SIZES_M = [2**n for n in range(3, 7, 3)]
+NUM_WARPS = [4, 8]
+NUM_STAGES = [2, 4]
+
+
+@triton.autotune(
+    configs=[
+        triton.Config(
+            {
+                "BLOCK_SIZE_RAGGED": b_r,
+                "BLOCK_SIZE_M": b_m,
+            },
+            num_warps=w,
+            num_stages=s,
+        )
+        for b_r, b_m, w, s in itertools.product(
+            BLOCK_SIZES_RAGGED,  # block sizes on non-reduction dimension
+            BLOCK_SIZES_M,  # block sizes on reduction dimension
+            NUM_WARPS,  # number of warps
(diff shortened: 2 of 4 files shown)


kit1980:
This is already in main as https://github.com/pytorch/benchmark/commit/c75d1a51d35cd4d7ba3518c991b23f500bf59b9f

PR #2386, opened 2024-07-18
Add variable-length loop Triton kernel benchmark for jagged_softmax

jananisriram:
Summary:
Add Triton kernel benchmark implementing a variable-length loop `softmax` for the `jagged_softmax` operator. This Triton kernel performs a `softmax` operation along the ragged dimension of a nested tensor of logical dimensions `(B, *, M)`, where `*` is the ragged dimension.

The kernel implements `softmax` in two phases.
1. As with previous implementations of the variable-length loop kernel (e.g. D59175612), the first phase loops through the maximum sequence length `MAX_SEQLEN` and loads in the input, filling in any blocks outside of the ragged bounds with negative infinity (as `e^-inf = 0`). It then subtracts the maximum value in the input tensor from the input block and takes the exponent, then stores the exponentiated block into the buffer. After the first phase, the kernel takes the `sum` of the buffer along the ragged dimension, resulting in a `sum` over all exponentiated input.
2. The second phase iterates over `MAX_SEQLEN` again to load the input, then divides the input by the `sum` of the buffer and stores this result in the output. For division, I used `tl.fdiv`, which performs just a bit faster than the regular division operator `/`.

As with previous jagged operators, the kernel is benchmarked against a padded PyTorch implementation and the simple fused Triton kernel, and it verifies accuracy against a baseline PyTorch `unbind` implementation.

This implementation uses the `buffer_then_sum` method; it adds all exponentiated input to a buffer, then takes the `sum` at once. This method has been proven to be faster, as seen in previous jagged operators (like `jagged_sum` and `jagged_mean`) and seems to prove that storing to and loading from many registers is faster than taking multiple sums in Triton.

For more information regarding other approaches I tried and some notes on the implementation, see D59299726.

Reviewed By: davidberard98

Differential Revision: D59309772

### DIFF
--- a/torchbenchmark/operators/jagged_softmax/__init__.py
+++ b/torchbenchmark/operators/jagged_softmax/__init__.py
@@ -1,0 +1,1 @@
+from .operator import Operator
--- a/torchbenchmark/operators/jagged_softmax/kernels.py
+++ b/torchbenchmark/operators/jagged_softmax/kernels.py
@@ -1,0 +1,120 @@
+import itertools
+
+import triton
+import triton.language as tl
+
+
+BLOCK_SIZES_RAGGED = [2**n for n in range(3, 12, 4)]
+BLOCK_SIZES_M = [2**n for n in range(3, 7, 3)]
+NUM_WARPS = [4, 8]
+NUM_STAGES = [2, 4]
+
+
+@triton.autotune(
+    configs=[
+        triton.Config(
+            {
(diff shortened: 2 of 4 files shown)


kit1980:
This already landed as https://github.com/pytorch/benchmark/commit/11ebabb2a15ef1d4ba2eb8ec60c713f703710fa2
Closing this PR.

PR #2389, opened 2024-07-19
Add packaging to docker

xuzhao9:
flash_attn requires packaging to build, so we need to include it in the docker image.

Test plan:

https://github.com/pytorch/benchmark/actions/runs/10024753897


manman-ren (approved):
Looks good to me! Thanks!

PR #2391, opened 2024-07-23
[tritonbench] Fix dimension order for FlashAttention-3

bertmaher:
Summary: We were passing tensors as B, H, S, D, which is expected by the Triton kernel, but FA3 expects B, S, H, D.

Also, the FA3 paper sizes inputs such that B*S=16k and H*Dhead=2048.  Let's replicate that here.  Also, print the complete tensor sizes in the output table, not just the sequence length.

Test Plan:
```
python run_benchmark.py triton --op flash_attention --only triton_tutorial_flash_v2,flash_v3 --metrics tflops --d-head 128
  (Batch, Heads, SeqLen, Dhead    triton_tutorial_flash_v2-tflops    flash_v3-tflops
------------------------------  ---------------------------------  -----------------
            (32, 16, 512, 128)                            328.882            430.32
           (16, 16, 1024, 128)                            389.723            517.099
            (8, 16, 2048, 128)                            433.721            565.544
            (4, 16, 4096, 128)                            458.359            590.233
            (2, 16, 8192, 128)                            470.732            614.234
           (1, 16, 16384, 128)                            476.753            620.73
```

### DIFF
--- a/torchbenchmark/operators/flash_attention/operator.py
+++ b/torchbenchmark/operators/flash_attention/operator.py
@@ -189,13 +189,17 @@ class Operator(BenchmarkOperator):
         )
         return fn
 
-    @register_benchmark(enabled=False)
+    @register_benchmark()
     def flash_v3(
         self,
         q: torch.Tensor,
         k: torch.Tensor,
         v: torch.Tensor,
     ) -> Callable:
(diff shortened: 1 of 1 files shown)


manman-ren (approved):
Thanks!

xuzhao9 (approved):

PR #2392, opened 2024-07-23
Run only a subset for optim benchmarks, attempt to fix

janeyx99:
A first attempt to fix https://github.com/pytorch/benchmark/actions/workflows/userbenchmark-regression-detector.yml by only running a subset of models

Impact of this change:
This workflow is green: https://github.com/pytorch/benchmark/actions/runs/10066723358/job/27828703722

It also takes so much shorter. At one point, the benchmark was 7 hours! Now it's 7 minutes.

## Test Plan

Test run: https://github.com/pytorch/benchmark/actions/runs/10065401507

Final test run: https://github.com/pytorch/benchmark/actions/runs/10066723358/job/27828703722

### DIFF
--- a/regression_detector.py
+++ b/regression_detector.py
@@ -412,7 +412,7 @@ if __name__ == "__main__":
     if not start_date:
         raise RuntimeError(
             f"No start date in previous JSONS found to compare towards the end date {end_date}. User specified start date: {args.start_date}. "
-            + f"Available JSON dates: {available_metrics_jsons.keys()}. No regression info has been generated."
+            + "No regression info has been generated."
         )
 
     print(
--- a/userbenchmark/optim/run.py
+++ b/userbenchmark/optim/run.py
@@ -60,7 +60,7 @@ SUBSET_OF_MODEL_NAMES: List[str] = [
     "hf_GPT2_large",
     "hf_T5_large",
     "resnet50",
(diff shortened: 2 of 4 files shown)


userbenchmark/optim/run.py
```diff
@@ -579,8 +592,6 @@ def run_model(modelName, device, Optim, defaults, maybe_pt2_):
             len(params),
             params[0].device,
         )
-        if Optim.__name__ == "SGD":
-            defaults["lr"] = 1e-2
```
janeyx99, line 583: no longer necessary as we have a default LR for SGD now

regression_detector.py
```diff
@@ -412,7 +412,7 @@ def get_metrics_by_date(
     if not start_date:
         raise RuntimeError(
             f"No start date in previous JSONS found to compare towards the end date {end_date}. User specified start date: {args.start_date}. "
-            + f"Available JSON dates: {available_metrics_jsons.keys()}. No regression info has been generated."
```
janeyx99, line 415: available_metrics_jsons doesn't have keys anymore as it's a List[str] and not a dict. It's unclear whether there were ever keys. Just remove the printout for now as it's not needed for the error message.

xuzhao9 (approved):

PR #2398, opened 2024-07-30
[tritonbench] Print traceback when exception occurs during benchmarking

bertmaher:
Without the traceback it's impossible to tell what went wrong.

### DIFF
--- a/torchbenchmark/util/triton_op.py
+++ b/torchbenchmark/util/triton_op.py
@@ -557,7 +557,7 @@ class BenchmarkOperator(metaclass=PostInitProcessor):
                 del self.example_inputs
                 gc.collect()
         except (KeyboardInterrupt, Exception):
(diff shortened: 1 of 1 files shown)


int3 (approved):
lgtm

int3:
Actually, doesn't re-raising the exception propagate the traceback as well?

bertmaher:
It does, but something further up the call stack swallows the exception.

bertmaher:
Okay yeah, here's the bit: https://github.com/pytorch/benchmark/blob/974d0c43851ecbd589610a8a9ac69e6864f6ea02/userbenchmark/triton/run.py#L151-L171.  There's a `finally` block ending with a `return` and that causes the exception to be swallowed.  We could move this `logger.warning` up to that level, but kind of feels like it could just stay here.

I'm also not sure how I feel about catching KeyboardInterrupt.  Usually that just leads to frustration, but 🤷

PR #2414, opened 2024-08-09
Fix CUPTI and Debugger libraries in the docker image

xuzhao9:
CUPTI and Debugger libraries are moved in the new cuda installer, we need to move the header files and libraries to the original place.

Test plan: 
https://github.com/pytorch/benchmark/actions/runs/10326154546
docker:
https://github.com/pytorch/benchmark/actions/runs/10326881008


aaronenyeshi (approved):
LGTM

PR #2419, opened 2024-08-14
[docker] Revert moving the cupti headers and library files

xuzhao9:
It turns out that the failed Kineto workflow is not related to Torchbench docker. We do not need to manually move the cupti header files, so let's keep it simple.


aaronenyeshi (approved):
LGTM! Sorry about the confusion

xuzhao9:
> LGTM! Sorry about the confusion

No worries!

PR #2420, opened 2024-08-15
logging hardware specific data in userbench

adamomainz:
Summary:
currently we are saving the device type in this table ie cuda vs rocm vs cpu etc. We want to further update to allow for hardware specific data ie a100 vs h100. Here we are specifying this for cuda based devices only.

If using cuda we will dynamically find the gpu model via nvidia-smi and log that model.

in this case for H100 we will see "Nvidia H100" and for a100 we will see "NVidia A100"

this will allow us to run the same benchmark on both types of gpu and find the difference in the results quickly via scuba at any given time

Differential Revision: D61229059

### DIFF
--- a/userbenchmark/triton/run.py
+++ b/userbenchmark/triton/run.py
@@ -162,8 +162,8 @@ def _run(args: argparse.Namespace, extra_args: List[str]) -> BenchmarkOperatorRe
             else:
                 print(metrics)
         if not hasattr(torch_version, "git_version") and args.log_scuba:
-            from userbenchmark.triton.fb import log_benchmark
-            log_benchmark(metrics)
+            from pytorch.benchmark.fb.run_utils import log_benchmark
+            log_benchmark(metrics, args.op)
         if args.plot:
             try:
                 opbench.plot()


userbenchmark/triton/run.py
```diff
@@ -162,8 +162,8 @@ def _run(args: argparse.Namespace, extra_args: List[str]) -> BenchmarkOperatorRe
             else:
                 print(metrics)
         if not hasattr(torch_version, "git_version") and args.log_scuba:
-            from userbenchmark.triton.fb import log_benchmark
-            log_benchmark(metrics)
+            from pytorch.benchmark.fb.run_utils import log_benchmark
```
nmacchioni, line 165: why is the import changing?

nmacchioni (approved):
LGTM!

userbenchmark/triton/run.py
nmacchioni: Oh nvm I see

adamomainz: Xu added another scuba logging util into this run_utils and im consolidating the files. see [D61229059](https://www.internalfb.com/diff/D61229059) internally

danzimm (approved):
I figured it out!

PR #2425, opened 2024-08-20
fixed a bug in triton fused attention when using fp16

sfzhu93:
Previously, it gave an error when using fp16:

AssertionError: First input (bf16) and second input (fp16) must have the same dtype!


xuzhao9:
Will it break the use case when the precision is bf16?

sfzhu93:
> Will it break the use case when the precision is bf16?

Yes. Working on a more general fix.

sfzhu93:
will do it in another PR.

PR #2429, opened 2024-08-21
Log chromium events to scuba

jamesjwu:
Summary:
X-link: https://github.com/pytorch/pytorch/pull/134118

For some reason exporting D61392607 is just not working, so I'm resubmitting this.

This diff implements a bunch of views for internal scuba viewing.

TODOS that I might punt to another diff:
- Saving cache stats via counter is definitely sus here, but there's not really a good way to track "fx graph cache hit for this compile phase" right now. Will think about this more.
- We should definitely log frame id, compile id, etc
- We should definitely be logging configs. That way, we can A/B test based on whether a config is turned on.
- idk what I'm doing with compile_uuid yet, but it's useful when you want to look at samples for a single run. I think if we had mast job info this field is not needed, but it's nice to be able to drill down to a single run and get its chrome trace view or icicle view, so idk

Reviewed By: oulgen

Differential Revision: D61603243


ezyang (approved):

jamesjwu:
Note: this is the same as https://github.com/pytorch/benchmark/pull/2429, but our internal exporter wasn't working properly so I had to resubmit

PR #2482, opened 2024-10-01
Improve release benchmark suites with a lower value of epoch

juliagmt-google:
https://github.com/pytorch/benchmark/issues/2468


atalman (approved):
lgtm

PR #2501, opened 2024-10-10
Change default gpu metric backend

FindHao:
The current GPU memory metric backend includes dcgm and nvml. They are reported from hardware and should be accurate. This PR adds the native torch way to collect GPU memory usage. It uses `torch.cuda.max_memory_allocated()`. The benefit is that it has lower overhead and is accurate on a shared GPU server when there are mutliple GPU processes from other users. It is because we don't implement the process filter for the other two backends.  

Use `--metrics-gpu-backend torch` to set the backend.


FindHao:
Sorry, mixed some other commits. will update it later.

FindHao:
> Sorry, mixed with some other commits. will update it later.

resolved

xuzhao9:
How about we make `torch` backend as default to make it consistent with PT2 benchmark runner?

FindHao:
> How about we make `torch` backend as default to make it consistent with PT2 benchmark runner?

done in https://github.com/pytorch/benchmark/pull/2501/commits/4398557a0e87df4fde310366cb52ba14977aad1e

userbenchmark/triton/run.py
```diff
@@ -92,6 +92,18 @@ def get_parser(args=None):
         default=None,
         help="Metrics to collect, split with comma. E.g., --metrics latency,tflops,speedup.",
     )
+    parser.add_argument(
+        "--metrics-gpu-backend",
+        choices=["default", "nvml"],
+        default="default",
```
xuzhao9, line 98: Let's change the default mode name to `torch` for readability.

run.py
```diff
@@ -477,18 +477,17 @@ def main() -> None:
     )
     parser.add_argument(
         "--metrics-gpu-backend",
-        choices=["dcgm", "default"],
+        choices=["dcgm", "default", "nvml"],
```
xuzhao9, line 480: Same here, let's change the default backend name to `torch`

FindHao: renamed in https://github.com/pytorch/benchmark/pull/2501/commits/8b48eeaecac07f8649241d5d8287b207cccd3c55

userbenchmark/triton/run.py
FindHao: renamed in https://github.com/pytorch/benchmark/pull/2501/commits/8b48eeaecac07f8649241d5d8287b207cccd3c55

PR #2503, opened 2024-10-10
Add --op-collection option

FindHao:
This PR add `--op-collection` to tritonbench. It can run multiple ops in defined operator collections. The default collection includes all ops not included in other collections. 

Operator collections are defined in `torchbenchmark/operators_collection/`. For each collection, you should define a `get_operators` function to return operators included in this collection.


torchbenchmark/operators/__init__.py
```diff
@@ -40,8 +59,16 @@ def _list_opbench_paths() -> List[str]:
     return opbench
 
 
-def list_operators() -> List[str]:
+def list_operators(op_collection: OP_COLLECTION = OP_COLLECTION.DEFAULT) -> List[str]:
```
FindHao, line 62: @xuzhao9 maybe we should use str list as arguments?

userbenchmark/triton/run.py
```diff
@@ -139,8 +146,12 @@ def get_parser(args=None):
     args, extra_args = parser.parse_known_args(args)
     if args.op and args.ci:
         parser.error("cannot specify operator when in CI mode")
-    elif not args.op and not args.ci:
-        parser.error("must specify operator when not in CI mode")
+    elif not args.op and not args.ci and not args.op_collection:
```
FindHao, line 149: since we set a default value for --op-collection, this will never be triggerred.

xuzhao9:
I am thinking we should have a clearer place to manage operator collection. Can we create a new directory `torchbenchmark/operator_collection` and place each operator collection there?

FindHao:
> use folders to represent op collections

updated in https://github.com/pytorch/benchmark/pull/2503/commits/5e1e7b164d7605a4de712fba371841a9d7d3a271

PR #2514, opened 2024-10-17
Fixes to prep for weights_only default flip

mikaylagawarecki:
Some fixes for https://github.com/pytorch/pytorch/pull/137602

### DIFF
--- a/torchbenchmark/models/opacus_cifar10/__init__.py
+++ b/torchbenchmark/models/opacus_cifar10/__init__.py
@@ -1,3 +1,4 @@
+import os
 from typing import Tuple
 
 import torch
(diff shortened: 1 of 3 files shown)


xuzhao9:
cc @kit1980, can you help review this?

torchbenchmark/models/opacus_cifar10/__init__.py
```diff
@@ -29,7 +31,13 @@ def __init__(self, test, device, batch_size=None, extra_args=[]):
         )
 
         self.model = models.resnet18(num_classes=10)
+        prev_wo_envvar = os.environ.get("TORCH_FORCE_NO_WEIGHTS_ONLY_LOAD", None)
```
mikaylagawarecki, line 34: We don't have access to the `torch.load` call within opacus here, updating opacus also won't help because torchbench doesn't depend on a nightly of opacus but rather stable release

PR #2524, opened 2024-10-22
Account for older numpy versions in #2514

mikaylagawarecki:
Tested in https://github.com/pytorch/pytorch/pull/138826


mikaylagawarecki:
`linux_test_cpu` failure is unrelated

kit1980 (approved):

PR #2547, opened 2024-11-25
Enable auto_quant for CPU userbenchmark

zxd1997066:


torchbenchmark/util/backends/torchdynamo.py
```diff
@@ -183,25 +183,36 @@ def apply_torchdynamo_args(
 
         if args.quantization:
             import torchao
-            from torchao.quantization import (
-                change_linear_weights_to_int4_woqtensors,
-                change_linear_weights_to_int8_dqtensors,
-                change_linear_weights_to_int8_woqtensors,
-            )
+            if model.device == "cuda":
```
chuanqi129, line 186: Let's keep original behavior in here, suggest to change as 
```
if model.device == "cpu":
    xxx
else:
    original path
```

torchbenchmark/util/backends/torchdynamo.py
```diff
@@ -183,25 +183,36 @@ def apply_torchdynamo_args(
...
+                elif args.quantization == "int8weightonly":
+                    torch._inductor.config.use_mixed_mm = True
+                    change_linear_weights_to_int8_woqtensors(module)
+                elif args.quantization == "int4weightonly":
+                    change_linear_weights_to_int4_woqtensors(module)
+            elif model.device == "cpu" and model.test == "eval":
+                if args.quantization == "auto_quant":
+                    module, example_inputs = model.get_module()
+                    with torch.no_grad():
+                        module=torchao.autoquant(torch.compile(module, mode='max-autotune'))
```
chuanqi129, line 210: why we need compile the module before autoquant? and why we need to hard code the compile mode as `max-autotune`? Can we follow similar style with original part?

torchbenchmark/util/backends/torchdynamo.py
```diff
@@ -183,25 +183,36 @@ def apply_torchdynamo_args(
...
+                if args.quantization == "int8dynamic":
+                    torch._inductor.config.force_fuse_int_mm_with_mul = True
+                    change_linear_weights_to_int8_dqtensors(module)
+                elif args.quantization == "int8weightonly":
+                    torch._inductor.config.use_mixed_mm = True
+                    change_linear_weights_to_int8_woqtensors(module)
+                elif args.quantization == "int4weightonly":
+                    change_linear_weights_to_int4_woqtensors(module)
+            elif model.device == "cpu" and model.test == "eval":
+                if args.quantization == "auto_quant":
```
chuanqi129, line 207: add else statement to warining that the other quant modes are still not support for cpu device

PR #2550, opened 2024-11-29
[Userbenchmark] Allow to continue on error when installing `test_bench`

shink:


shink:
Hey @FindHao does this change make sense? Could you please help review? Thanks so much!

shink:
@FindHao Thanks for your approval. Please let me know if there's anything that needs to be done.

FindHao:
> @FindHao Thanks for your approval. Please let me know if there's anything that needs to be done.

Nothing else need to be done on your side. I've exported to internal review flow. Should be mergered tomorrow.

shink:
@FindHao ah thanks so much for your quick reply!

PR #2561, opened 2024-12-17
Resurrect ao benchmark on AWS A100 runner

huydhn:
I'm bringing back some example models first, one for each set (TIMM, HF, TorchBench), to have some data to unblock our TorchAO ClickHouse migration.  More can be added later if we decide to keep this workflow.

### Testing

https://github.com/pytorch/benchmark/actions/runs/12388956432/job/34581035274

The results are now available on `oss_ci_benchmark_v3` table `select * from oss_ci_benchmark_v3 where workflow_id = 12388956432`

### DIFF
--- a/.github/workflows/_linux-benchmark-cuda.yml
+++ b/.github/workflows/_linux-benchmark-cuda.yml
@@ -76,6 +76,6 @@ jobs:
         uses: pytorch/test-infra/.github/actions/upload-benchmark-results@main
         with:
           benchmark-results-dir: ${{ env.OUTPUT_DIR }}
-          dry-run: true  # DEBUG: TO BE REMOVED
+          dry-run: false
           schema-version: v3
           github-token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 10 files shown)


xuzhao9:
I thought the torchao team has decided to build benchmark using individual models like https://github.com/pytorch/pytorch/tree/main/benchmarks/gpt_fast. @jerryzh168 

Do we want to keep maintaining this?

jerryzh168:
> I thought the torchao team has decided to build benchmark using individual models like [pytorch/pytorch@`main`/benchmarks/gpt_fast](https://github.com/pytorch/pytorch/tree/main/benchmarks/gpt_fast). @jerryzh168
> 
> Do we want to keep maintaining this?

we are looking at multiple benchmark suite:

1). gpt_fast: has some issues right now actually
2). torchao/_models: right now running locally every day in devserver and uploads the results to s3
3). torch/benchmark torchbench model set: same as 2)

we want to show dashboard for 2) and 3), summarize geomean speedup and track the progress

huydhn:
I think my goal here is to bring this back for one example TorchBench model, then we will have some data to decide on the next step here 1) to bring this back or 2) to improve this in some way to achieve the (3) goal that @jerryzh168 mentions

xuzhao9 (approved):

PR #2563, opened 2025-01-07
NotImplementedError: Model's DEFAULT_TRAIN_BSIZE is not implemented.

ostrowskimarcin:
Some tests fail in train mode. I found missing fields in specific models and added DEFAULT_TRAIN_BSIZE with the same value as eval batch size.

### DIFF
--- a/torchbenchmark/models/cm3leon_generate/__init__.py
+++ b/torchbenchmark/models/cm3leon_generate/__init__.py
@@ -7,6 +7,7 @@ from .model import create_model, SequenceGenerator
 
 class Model(BenchmarkModel):
     task = NLP.LANGUAGE_MODELING
+    DEFAULT_TRAIN_BSIZE = 1
     DEFAULT_EVAL_BSIZE = 1
 
     def __init__(self, test, device, batch_size=None, extra_args=[]):


torchbenchmark/models/pyhpc_turbulent_kinetic_energy/__init__.py
```diff
@@ -129,6 +129,7 @@ class Model(BenchmarkModel):
     # Source: https://github.com/dionhaefner/pyhpc-benchmarks/blob/650ecc650e394df829944ffcf09e9d646ec69691/run.py#L25
     # Pick data-point when i = 20, size = 1048576
     DEFAULT_EVAL_BSIZE = 1048576
+    DEFAULT_TRAIN_BSIZE = 1048576
```
xuzhao9, line 132: We do not need to add train bsize because the `train()` function is not implemented or supported by this model.

mostrowskix:
@xuzhao9 You're right. Most of these classes have train method but they return an exception. I have to take it into account while testing. But still, there's one model that has the method implemented without DEFAULT_TRAIN_BSIZE field. I updated my PR. Take a look, please.

ostrowskimarcin:
@xuzhao9 Yes, you're right. Most of the classes have a train method but they return only an exception. I have to take it into account while testing. But still, there's one model that has an implemented train method without DEFAULT_TRAIN_BSIZE field. I updated my PR. Can you take a look, please?

ostrowskimarcin:
@xuzhao9

ostrowskimarcin:
@xuzhao9 could you assign anyone else if you're busy?

xuzhao9 (approved):

ostrowskimarcin:
@xuzhao9 thanks for approval. Can you merge?

PR #2622, opened 2025-07-02
Check for requirements before installing `conda` packages during `doctr_(det|reco)_predictor`

hameerabbasi:
In the `install.py` for two modules, `conda` is invoked to install packages. This PR adds a short check to only invoke `conda install` if the packages don't already exist in the conda environment. This speeds up the installation significantly in cases where the requirements already exist in the environment.


hameerabbasi:
@xuzhao9 From the commit history, it seems like you're the right person to review this PR.

PR #2626, opened 2025-07-10
Ignore local version identifiers in constraints.

senokay:
Torch is frequently installed with a local version identifier (e.g. `+cu126`), but that identifier causes an issue with pip because there is no public package with that identifier unless an explicit alternative index is used. The issue also happens to come with a quite cryptic ResolutionImpossible error (pypa/pip#13422) which completely obscures the actual issue.

Stripping the local version identifier should be safe, as we are only using constraints to ensure that per-model dependencies are aligned with existing installations.


xuzhao9:
This is a bit weird because we test on pytorch nightly, where the version is like `2.8.0.dev20250606+cu128`, so I thought pip can correctly handle version suffix like `+cu128`...

PR #2635, opened 2025-08-20
Revamp TorchBench nightly image build process

huydhn:
Notable changes:

1. Use NVIDIA image `nvidia/cuda:${CUDA_VERSION}-devel-ubuntu22.04` as the base so that we don't need to setup CUDA
2. Replace `conda` with `uv`
3. Cleanup unused code path
4. Unpin TIMM and transformers to get their latest stable version

### Testing

https://github.com/pytorch/benchmark/actions/runs/17114624920/job/48542992235


requirements.txt
```diff
@@ -9,10 +9,8 @@ pytest
 pytest-benchmark
 requests
 tabulate
-git+https://github.com/huggingface/pytorch-image-models.git@730b907
-# this version of transformers is required by linger-kernel
-# https://github.com/linkedin/Liger-Kernel/blob/main/pyproject.toml#L23
-transformers==4.44.2
+timm==1.0.19
```
Rohanjames1997, line 12: Why was the version of `timm` pinned?

huydhn: Oh, I plan to update them in a more controlled manner via Dependabot (PR is coming)

PR #2639, opened 2025-08-27
Maintain 2 versions of maml_omniglot for numpy 1x and 2x

huydhn:
I want to update TorchBench pinned commit on PyTorch CI and need to maintain 2 flavors of this file for numpy 1.x and 2.x.  The discrepancy here is because TorchBench CI uses latest numpy (2.x) while PyTorch CI pins [numpy to 1.x](https://github.com/pytorch/pytorch/blob/main/.ci/docker/requirements-ci.txt#L137) for Python =< 3.13.  I don't want to update numpy on PyTorch CI because it's far from easy.  So, it's easier to maintain 2 flavors here.

This will unblock https://github.com/pytorch/pytorch/pull/161461

### DIFF
--- a/torchbenchmark/data/index.yaml
+++ b/torchbenchmark/data/index.yaml
@@ -14,4 +14,5 @@ INPUT_TARBALLS:
   - sam_inputs.tar.gz
 MODEL_PKLS:
   - drq/obs.pkl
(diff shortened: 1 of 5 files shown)


huydhn:
https://github.com/pytorch/benchmark/pull/2640 is the right fix for this

PR #2640, opened 2025-08-28
Install the right numpy version matching PyTorch CI

huydhn:
As pointed out in https://github.com/pytorch/benchmark/pull/2639, the new Docker image missed this step and installed numpy 2.x, which is not the same version used in PyTorch CI https://github.com/pytorch/pytorch/blob/main/.ci/docker/requirements-ci.txt.  With this done, I could revert `batch-20250825.pt` back to the original `batch.pt`.

I also clean up `.github/scripts/run.sh` which is not used in CI anymore.  Utilities like `cuda_utils.py` would need to be refreshed too, but I will do that later in the context of other functionality to bring back like bisection.

### Testing

[numpy==1.26.2](https://github.com/pytorch/benchmark/actions/runs/17282489625/job/49053486634) is installed in the image now.


xuzhao9 (approved):

huydhn:
> We may want to keep batch-20250825.pt in the S3 bucket for future use, when pytorch ci (inductor benchmark tests) are upgraded to numpy 2

Yes, I keep the file on S3 in case

PR #2652, opened 2025-10-16
Set the right dependency name for GAT, GCN, and SAGE pyg-library

huydhn:
It's called https://pypi.org/project/pyg-library now, I'm not sure if we even need to keep these canary models, but this is an easy tweak


xuzhao9 (approved):
Thanks for the fix!

PR #2654, opened 2025-11-19
Add Flux benchmark

StrongerXi:
This adds a benchmark for the Flux image generation pipeline. Specifically, it only benchmarks the diffusion transformer (and omits the text encoder and vae, which don't take up much time for the e2e generation in Flux).

Needs https://github.com/pytorch/pytorch/pull/168176 to run in pytorch repo:
```
python ./benchmarks/dynamo/torchbench.py --accuracy --inference --backend=inductor --only flux
python ./benchmarks/dynamo/torchbench.py --performance --inference --backend=inductor --only flux
```

### DIFF
--- a/torchbenchmark/models/flux/__init__.py
+++ b/torchbenchmark/models/flux/__init__.py
@@ -1,0 +1,72 @@
+import torch
+from torchbenchmark.tasks import COMPUTER_VISION
+from torchbenchmark.util.framework.huggingface.model_factory import HuggingFaceAuthMixin
+from torchbenchmark.util.model import BenchmarkModel
+
+from .install import load_model_checkpoint
+
+
+class Model(BenchmarkModel, HuggingFaceAuthMixin):
+    task = COMPUTER_VISION.GENERATION
+
+    DEFAULT_TRAIN_BSIZE = 1
+    DEFAULT_EVAL_BSIZE = 1
+    ALLOW_CUSTOMIZE_BSIZE = False
+    # Skip deepcopy because it will oom on A100 40GB
+    DEEPCOPY = False
+    # Default eval precision on CUDA device is fp16
+    DEFAULT_EVAL_CUDA_PRECISION = "fp16"
+
+    def __init__(self, test, device, batch_size=None, extra_args=[]):
+        HuggingFaceAuthMixin.__init__(self)
+        super().__init__(
+            test=test, device=device, batch_size=batch_size, extra_args=extra_args
+        )
+        self.pipe = load_model_checkpoint()
+        self.example_inputs = {
+            "prompt": "A cat holding a sign that says hello world",
+            "height": 1024,
+            "width": 1024,
+            "guidance_scale": 3.5,
+            "num_inference_steps": 50,
+            "max_sequence_length": 512,
+            "generator": torch.Generator("cpu").manual_seed(0),
+        }
+        self.pipe.to(self.device)
+
+    def enable_fp16(self):
+        # This model uses fp16 by default
+        # Make this function no-op.
+        pass
+
+    def get_module(self):
+        # A common configuration:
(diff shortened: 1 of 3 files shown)


StrongerXi:
cc @sayakpaul we are trying to add some diffusers models to our nightly benchmark, this represents a first attempt using Flux. Would love to get your thoughts on
1. What models to add. I'm thinking focusing on category first (e.g., Flux for txt2img, Qwen Edit for txt-img2img, Wan for txt2video, ...), then add more popular models to improve coverage.
2. Whether there are incentives to consolidate this and the [diffusers benchmarks](https://github.com/huggingface/diffusers/blob/b71269675ec1b85193107a691dd35c308e46f0a5/benchmarks/README.md). At a high level I think the pytorch benchmark will enable early detections of pytorch changes that break compile + diffusers (help nightly users and avoid bug congestion during release), and the diffusers benchmark will do similar thing, but for diffusers changes and users. If the benchmark infra are sufficiently different maybe it's fine to have a bit of code dup.
3. What exactly to benchmark -- right now we are just benchmarking the denoising transformer. Are there incentives to benchmark the txt/img encoder and vae as well? Say if the number of steps are low for distilled model variants? There are also variables like image size, which actually affects torch.compile speedup ratio, as larger size make it more compute bound e2e.

Also cc @anijain2305 @BoyuanFeng.

torchbenchmark/models/flux/__init__.py
```diff
@@ -0,0 +1,72 @@
+import torch
+from torchbenchmark.tasks import COMPUTER_VISION
```
StrongerXi: These 3 files are mostly an adaptation of https://github.com/pytorch/benchmark/tree/f0b2a09591de5cf276bbfa7ce06a3d35f508da84/torchbenchmark/models/stable_diffusion_unet for Flux.

sayakpaul:
> What models to add. I'm thinking focusing on category first (e.g., Flux for txt2img, Qwen Edit for txt-img2img, Wan for txt2video, ...), then add more popular models to improve coverage.

Yeah, this list is perfect! Many models in the image and video gen space just come with a lot of promise and vanish quickly. Only a few stick around. The ones you mentioned are the most promising ones and have somewhat stood the "test of time". I would also add SDXL to this mix.

> Whether there are incentives to consolidate this and the diffusers benchmarks. At a high level I think the pytorch benchmark will enable early detections of pytorch changes that break compile + diffusers (help nightly users and avoid bug congestion during release), and the diffusers benchmark will do similar thing, but for diffusers changes and users. If the benchmark infra are sufficiently different maybe it's fine to have a bit of code dup.

I am fine with that, I think that's a great idea! We could always test against the nightlies (both PyTorch and Diffusers) to ensure no performance regression. [This](https://github.com/huggingface/diffusers/actions/runs/19382758708/workflow#L25) ([reference on AWS](https://aws.amazon.com/ec2/instance-types/g6e/)) is the infra on which our benchmark is run.

> What exactly to benchmark -- right now we are just benchmarking the denoising transformer. Are there incentives to benchmark the txt/img encoder and vae as well? Say if the number of steps are low for distilled model variants? There are also variables like image size, which actually affects torch.compile speedup ratio, as larger size make it more compute bound e2e.

I don't think there are any incentives to benchmark the other components, TBH, as any gain in the denoiser is just a compounding factor in most cases. I think benchmarking for different batch sizes and for 2/3 different resolutions is a good idea. Also, there are multiple use cases beyond text-to-{image,video,audio} such as image-guided editing, etc. But I think it's fine to just keep it at the text-to-X level as it's easy to extrapolate to the other use cases.

Just as a note, in our benchmark, we also check for quantization backends and memory features like layerwise casting. I understand that this deviates from a core PyTorch benchmark, but these features are quite popular in the community, hence we benchmark them. Nothing to fret about, just noting.

PR #2659, opened 2025-11-26
Allow using alternative pip installers in installation scripts

yuxuan-z19:
Currently, installation scripts always invoke pip via `sys.executable -m pip` in subprocesses.  
This PR introduces `get_pip_cmd()`, which:

- Uses the `PIP_MODULE` environment variable to override the installer if set.
    
    ```bash
    PIP_MODULE="uv pip" python install.py
    ```

- Defaults to `sys.executable -m pip` otherwise.

This improves flexibility and compatibility with environments that use alternative pip executables. (See [install.log](https://github.com/user-attachments/files/23767242/install.log) for an example with `uv`.)

### DIFF
--- a/utils/python_utils.py
+++ b/utils/python_utils.py
@@ -30,8 +30,6 @@ def create_conda_env(pyver: str, name: str):
 def get_pip_cmd():
     if env := os.getenv("PIP_MODULE"):
         return env.split()
-    if shutil.which("uv"):
-        return ["uv", "pip"]
     else:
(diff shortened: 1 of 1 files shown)


yuxuan-z19:
Removed the uv fallback, so the installer now only honors `PIP_MODULE` when provided and otherwise falls back to `sys.executable -m pip`, as discusses [here](https://github.com/meta-pytorch/tritonbench/pull/672#issuecomment-3590098168)

PR #2683, opened 2026-04-15
Improve benchmark installation speed

benjaminglass1:
This solves a problem noticed while installing benchmarks and watching htop.  A significant amount of time per installation is wasted spinning up new Python subprocesses to repeatedly import packages and check their versions. This wastes substantial time when the package import takes several second (as with `torch`). This PR cuts down benchmark install time on my machine from 35 minutes to 12.

I also fixed one or two minor Ruff lints in the files I touched.

### DIFF
--- a/torchbenchmark/util/env_check.py
+++ b/torchbenchmark/util/env_check.py
@@ -3,14 +3,17 @@ PyTorch benchmark env check utils.
 This file may be loaded without torch packages installed, e.g., in OnDemand CI.
 """
 
-import argparse
 import copy
 import logging
 import os
 import shutil
 from collections.abc import Mapping
 from contextlib import contextmanager, ExitStack
-from typing import Any, Dict, List, Optional
+from importlib.metadata import version
+from typing import Dict, List, Optional, TYPE_CHECKING
+
+if TYPE_CHECKING:
+    import torchbenchmark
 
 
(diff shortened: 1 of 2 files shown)


xmfan (approved):

benjaminglass1:
Test failures look the same as those on `main`.

benjaminglass1:
@xmfan Please take a second look, I had to rewrite this after realizing this function is _also_ used to validate that package versions haven't changed after installing benchmarks (which caching would invalidate). Currently getting some timing runs, but I still expect roughly the same speedup.

benjaminglass1:
On my machine, installing all benchmarks on `main` takes roughly 35 minutes. With this change, installing all benchmarks takes roughly 12 minutes. These timings are obviously super variable, but that hopefully gives a sense of how impactful this change is.

The main reason for this is that we checked the version of several dependencies twice per benchmark. We got the version by importing the package in a new subprocess and querying `__version__`. For `torch`, this takes 2-3 seconds (on my machine), so that's a minute wasted every 10-15 packages. The new version queries PEP 566 metadata from the package in less than a millisecond, effectively eliminating the delay.

huydhn (approved):

xuzhao9 (commented):
LGTM!

PR #2700, opened 2026-08-19
Disable sam_fast custom flash without Triton block pointers

warrendeng:
The optional `segment-anything-fast` `flash_4` kernel still uses Triton's legacy `tl.make_block_ptr` and `tl.advance` APIs, which were removed by triton-lang/triton#10833. This causes the full `sam_fast` TorchBench model to fail during eager validation with newer Triton revisions.

Detect those APIs before importing `segment_anything_fast` and use its documented `SEGMENT_ANYTHING_FAST_USE_FLASH_4=0` switch when they are unavailable. The benchmark continues exercising the full model through its existing SDPA fallback, while older Triton behavior remains unchanged.

Test plan:
- `python3.12 -m py_compile torchbenchmark/models/sam_fast/__init__.py`
- Verified with Triton 3.7 that the custom kernel remains enabled when both APIs exist
- Simulated missing `tl.advance` and verified the adapter disables the custom kernel before package import

### DIFF
--- a/torchbenchmark/models/sam_fast/__init__.py
+++ b/torchbenchmark/models/sam_fast/__init__.py
@@ -6,6 +6,14 @@ import os
 import cv2
 import numpy as np
 import torch
+import triton.language as tl
+
+# The optional flash_4 kernel requires Triton's legacy block-pointer frontend.
+# Disable it before importing segment_anything_fast so the model uses its SDPA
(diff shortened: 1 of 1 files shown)


warrendeng:
Superseded by #2701, which uses a branch in pytorch/benchmark so the repository CI credentials are available.

PR #2708, opened 2026-09-20
feat: support privateuse1 backends in benchmark harness

markc-614:
Generalize device handling so third-party accelerators registered via the  privateuse1 mechanism can run the benchmark, instead of hardcoding CUDA

  - Detect and append the registered privateuse1 backend to SUPPORT_DEVICE_LIST
  - Replace torch.cuda.synchronize() with backend-agnostic torch.get_device_module(device).synchronize() in timing/memory paths
  - Add register_deterministic_backend() and thread `device` through save/load_deterministic_dict() to apply backend-specific determinism flags
  - Gate gpu_peak_mem on max_memory_allocated availability instead of device == "cuda"
  - Generalize OOM detection from torch.cuda.OutOfMemoryError to torch.OutOfMemoryError
  - Use get_device_module for empty_cache / amp / device-name resolution


markc-614:
@xuzhao9 @benjaminglass1 @retonym Could you please review this PR when you get a chance? Thanks!

